PR #13, opened 2017-05-31
Serverstats: Better fitting FA-Icons

Ryvid:
Replaced: fa-user -> fa-user-plus ; "fa-trash" for "Online members" -> fa-user ; "fa-trash" for "Total members" -> fa-users


jonas747:
Amazing

PR #19, opened 2017-06-02
Corrected typos

Ryvid:


jonas747:
na its graw here,  its the reddit package i used a while ago although i use another one now.

PR #72, opened 2017-07-14
Create commands.md

timcampy118:
Wasn't sure if we needed screenshots or not. Even if we did, hopefully someone has something better than paint to mark the screenshots. 

Basic info about commands page including

- Global settings
- Channel overwrite setting
- description of each option they can change
- prefix changing
- redirecting moderation commands to moderation page to turn on


jonas747:
can you put this in `commands/assets/help-page.md` ?

PR #110, opened 2017-10-25
Update swearwords.go

OliverLusa:
Ass blocks me saying bypass etc


jonas747:
No it doesn't, the whole word has to match, automod does not do partial matches currently.

PR #164, opened 2018-02-15
Update moderation.html

buthed010203:


jonas747:
Don't think i want that there tbh

buthed010203:
True, Its quite long

PR #168, opened 2018-02-17
Throw update

buthed010203:
I like throwing things at other things


jonas747:
remove 

a ping sock
a generic object

buthed010203:
Changed to explosives and myself

PR #174, opened 2018-02-20
Update yagcommmand.go

buthed010203:


jonas747:
Please in the future describe what you did in the commit message, rather than just "update xxx file"

PR #204, opened 2018-03-12
another spelling/grammar thing

Evildeathpro:
soundboard


Evildeathpro:
oh btw these are all seperate incase you think some of these are meant to be there and some arent meant to be there

PR #233, opened 2018-04-20
Refactor stdcommands into command packages.

LoganK:
There is no new functionality here. This is merely a step toward Go ideals of
accept interfaces return structs.

The intent is to simplify dependencies, make it easier to test, and make
command additions almost as simple as coping a directory and modifying. I kept
the manual command list in stdcommands to preserve existing command ordering in
the help, but a future implementation may simply use a sorted, implicit
registry.


jonas747:
Hmmm not sure if i agree with each command having its own package, something as each command having their own file would have been enough imo

Also not sure about this new interface, what does this improve exactly?

LoganK:
There are a few primary advantages to separating out into their own directories:

- Like-resources are grouped together (e.g., tzoffsets and catfacts databases)
- Names can overlap. Copying a directory as a template for a new command no longer requires renaming all of the variables as a first step: you simply change the package name.
- Easier to avoid unintended dependencies between commands
- Test cases (if any were written) can be run against a single command (so quick to run)
- Faster compilation

As for the interface, this is common golang practice[1]. In most languages, I'd agree that it provides little value as the cost of this abstraction weighs on development, but Go's behavior-based interface system avoids all of that. As an idea of where this might be headed...

- Commands no longer depend on github.com/jonas747/yagpdb/stdcommands so it becomes possible to write dynamically loadable plug-ins.
- If commands are dynamically loaded, you could conceivably update or add commands without rebooting the server.
- With another round of refactoring, the commands could focus solely on the behavior and less on the YAGPDB specifics.
- Once they are decoupled from YAGPDB, it becomes natural to implement common patterns in a single place (`PersistChannelSetAdd(channel,` "music_categories", "dubstep")` as opposed to raw Redis commands to construct a key)
- The system then becomes easier to refactor and less prone to breakages when updating libraries
- It's conceivable at a certain level that commands can be shared by different engines (think IRC, Slack, or whatever comes after Discord)


So beyond being idiomatic and good practice, there's also a bit of long-term vision that you, as the architect, have to buy into. That's why I sent this small exploratory change to see what your thoughts are. Maybe you don't see it as a problem that every command has 3--5 "standard" dependencies across a few different repositories, and, if that's the case, I probably shouldn't proceed with these cleanup requests as we'd be at odds. I'm just lurking in the Discord channel at the moment, but it's a difficult concept to convey over text. I suspect, thought, based on the way the bot/ stuff is implemented, that you already appreciate a number of these points.

As I said, this is unique to go and it's dynamic interface mechanisms. In other static languages it's a nightmare of dependencies that I wouldn't wish on anyone. In dynamic languages you get the same lovely flexibility but things crash if you don't have lots of test cases. My experience with large Go systems is that this "static duck-typing" (as I call it) works well can be a real advantage.


As an aside, thanks for writing such a great structure and open-sourcing it. I was really shocked that there weren't more frameworks providing persistence and web-based configuration, but YAGPDB does pretty much everything except localization. Such an awesome job. :-)


[1] http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/

LoganK:
So ignoring all of my pie-in-the-sky spiel about abstraction... (Especially as I've read more of the code base and see we are very aligned on many of these things.)

I still think this change is a good one. The `go fix` command you link to implements the command as a distinct package in a distinct directory, very similar to this refactoring. (Although it exposes the struct directly.)

I'd like to get this off your plate, so we have three primary options:
1) Fix conflicts and accept
2) Keep as separate packages but remove the useless Cmd() functions
3) Simply close it out

I'd prefer number 2, but I'd be glad to /dev/null it for now.

jonas747:
I'm fine with 2 aswell

LoganK:
Updated as discussed.

Because the diffs are tough to read, note that 5271c4 and 3ce168 both modified the same files. I can also flatten into a single commit if it would help.

jonas747:
Hm nah im squashing it anyways, looks good

PR #235, opened 2018-04-23
Web validation: Fix tests, add support for slices

LoganK:
Working on a feature request where I thought I'd need slices, but, as that no longer appears to be the case, submitting a pull request with no intent to use. It has tests so isn't all bad. 😄 

I fixed the other unit tests along the way as they were simple enough.


jonas747:
I should probably change to a proper validation library at some point, but thanks

PR #242, opened 2018-05-03
Multi-message support for join/leave messages and custom commands

LoganK:
The intended use case is to diversify the messages to avoid fatigue. It's more fun if it's mixed up a little. ;-)

Please note that there is no UI to let the user know they have exceeded the message limits. The limit is meant to be generous so I opted for simplicity.


jonas747:
Ah very nice, I'm gonna go through it more in depth and possibly merge it in the weekend.

jonas747:
Looks good, nice work, i do wanna add to this later to do more than just random responses though!

Issue #239, opened 2018-04-28
Suggestion: Auto add reactions to messages in channels


Dubz:
It would be fantastic if there was a way to set a channel (or multiple channels) up for auto reactions.

For instance, if you set up an announcements for events and want to allow users to vote Yes, Maybe, or No as to whether or not they will be attending said event.
Another example being a suggestion/feedback, where users can vote thumbs up/down on a suggestion sent, rather than blowing it up with messages of "I agree" or "I want this too" sort of messages. Keeps the channel more free of general chatter and allows for focus on those points.
You could also add a voting scale for one, and react with 1-5.

Just a few examples of how this could be used, for both productivity and fun.

jonas747:
Adding this functionality to custom commands would be enough i believe, something like a {{addReaction "..."}} function, then you could just set it to run when the message contains something like startvote or whatever

buthed010203:
As Jonas said that would be plenty as you could create a command with a regex trigger of (?) that means any text will trigger the command, you could make it only usable in certain channels

buthed010203:
^ The reactions could be perfectly timed if the {{wait}} thing ever got added....

PR #243, opened 2018-05-07
addReactions template function for adding reactions

LoganK:
There is still a lot to be desired (see commit comments), but it addresses much of the functionality and provides a reasonable base to continue development.

It can be used in either of two ways (or mixed):

= Fixed =

{{addReactions "👍" "👎"}}

The emojis hard-coded into the template will be included.

= Dynamic =

{{addReactions .CmdArgs}}

The emojis provided as arguments to the custom command will be used, such as
"Best animal? poll: 🙈 🦁 🐞 🐥" (where "poll:" is the trigger phrase).

#239


jonas747:
Ah very nice, even some tests 👍 

i say just leave in strippedmsg for now.

As for the others, not sure what you mean by named emojis, its my understanding that discord converts them to the unicode equivalent, is that not the case?

As for custom emojis, they are in the format of <name:id> in the message, shouldn't be too hard to match them against a regex i believe, and you add them to messages using the name:id format.

As for emojis needing to be separated by space, i think that's fine, for now.

LoganK:
From my testing, the Discord "Create Reaction" API accepts only two forms for the emoji:

* The actual emoji (such as the copy/paste) above
* A custom emoji in the form ":name:id"

My comments are about these restrictions. There is currently no support for named emoji (e.g., ":skull_crossbones:") or assistance with the custom lookup (e.g., simply ":name:" without the ID).

Custom emojis are particularly bad because the ID isn't really accessible unless you watch the network traffic or use the Get Guild Emoji API (which I don't think is being cached anywhere in YAGPDB today). Further, I don't think an API exists for fetching global emoji (for Discord Nitro users).

That being said, ":name:id" does work in this pull request, but it's just complicated enough that I haven't bothered to document the behavior.

Speaking of...

I added one more commit so that addReactions is documented.

I think it's the right implementation, but I wanted to capture that there is still a lot of low-hanging fruit for improvement.  If people are using the feature, then it can be added YAGPDB later with no foreseen impact to the functionality as-is.

jonas747:
The format for custom emojis is `name:id`
but how they appear in messages is `<name:id>` (you can see this if you escape it with `\`)
a regex like `<:.*:[0-9]{10,25}>` wold find that.

but yeah i will merge this as is if you fix the merge conflict, the proper custom emoji support can be added on later.

PR #244, opened 2018-05-07
Add new template func "slice" for creating sub-slices

LoganK:
I have no immediate use for this functionality. I was using this to more effectively make use of .Args in command templates, but the new .CmdArgs serves my needs with far less effort. I offer it here only in case you think it might be useful at some point.

Example use:

{{slice .Args 1 | addReactions}}

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -294,7 +294,7 @@ func slice(item reflect.Value, indices ...reflect.Value) (reflect.Value, error) 
 		}
 
 		if args[0] < 0 || args[0] >= v.Len() {
-			return reflect.Value{}, errors.Errorf("beg index out of range: %d", args[0])
+			return reflect.Value{}, errors.Errorf("begin index out of range: %d", args[0])
 		} else if args[1] <= args[0] || args[1] >= v.Len() {
 			return reflect.Value{}, errors.Errorf("end index out of range: %d", args[1])
 		}
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -255,3 +255,52 @@ func tmplFormatTime(t time.Time, args ...string) string {
 
 	return t.Format(layout)
 }
+
+// slice returns the result of creating a new slice with the given arguments.
+// "slice x 1 2" is, in Go syntax, x[1:2], and "slice x 1" is equivalent to
+// x[1:].
+func slice(item reflect.Value, indices ...reflect.Value) (reflect.Value, error) {
+	v, _ := indirect(item)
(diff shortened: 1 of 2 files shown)


common/templates/general.go
```diff
@@ -255,3 +255,52 @@ func tmplFormatTime(t time.Time, args ...string) string {
...
+		case 1:
+			args = append(args, v.Len()+1-args[0])
+		case 2:
+			break
+		default:
+			return reflect.Value{}, errors.Errorf("unexpected slice arguments %d", len(args))
+		}
+
+		if args[0] < 0 || args[0] >= v.Len() {
+			return reflect.Value{}, errors.Errorf("beg index out of range: %d", args[0])
```
jonas747: i don't see a reason for shorting "beginning" to "beg", it will only create confusion.

common/templates/general.go
```diff
@@ -255,3 +255,52 @@ func tmplFormatTime(t time.Time, args ...string) string {
...
+// slice returns the result of creating a new slice with the given arguments.
+// "slice x 1 2" is, in Go syntax, x[1:2], and "slice x 1" is equivalent to
+// x[1:].
+func slice(item reflect.Value, indices ...reflect.Value) (reflect.Value, error) {
+	v, _ := indirect(item)
+	if !v.IsValid() {
+		return reflect.Value{}, errors.New("index of untyped nil")
+	}
+
+	var args []int
```
jonas747: Hmm feel like this is a bit sketchy, but atleast make it []int64 since that's used more and will stop unexpected problems from happening on 32bit systems.

LoganK: I really hope there isn't a scenario where someone needs a 64-bit slice. 😄 I chose int rather than int64 because reflect.Value.{Len,Slice}() use an int and so I a) end up with fewer type conversions and b) am not pretending to support ranges that aren't true.

I'm glad to switch if you still wish, but I don't see the specific architecture concerns.

I didn't intend this to be "sketchy".

common/templates/general.go
LoganK: Agreed and done.

(Chose "begin index" and kept "end index". I wanted to use the Slice arguments, but "i" and "j" are terrible labels. 😛)

LoganK:
Label changed pushed.

common/templates/general.go
jonas747: All discord id's are 64bit so pretty commonly used

jonas747: Ahhh yeah i see i misread it, woops, sorry man

PR #245, opened 2018-05-08
poll: New standard command for creating reaction polls

LoganK:
Usage:
-poll "Which indent?" tabs spaces "carriage returns are for posers"


jonas747:
Look good

PR #247, opened 2018-05-09
Updated addReaction template

michelhn:
Added small explanation and link to https://unicode.org/Public/emoji/11.0/emoji-test.txt


LoganK:
I'm pretty sure that if there is a fully-qualified and non-fully-qualified variant, it must be the *non*-fully-qualified variant. Feel tree to test and let me know if I'm wrong. (I was mostly testing with keycaps which perhaps behave differently?)

Also, I should just fix it to strip FEOF if found.

And add named emoji. :smile:

LoganK:
https://github.com/jonas747/discordgo/pull/2

LoganK:
Might be worth updating the template documentation to state that global emoji (Nitro) are not supported.

jonas747:
So whats the status on this?

michelhn:
Can be closed due to gitbook

PR #252, opened 2018-05-27
making things more colorful

michelhn:
the world needs to be more colorful


kennedyguy:
Hmm. It is a good idea, michelhhn, but jonas747 is also right. So its a 50-50.

PR #259, opened 2018-06-01
Added FFMPEG for Soundboard support

casKd-dev:


michelhn:
This is necessary for the soundboard.

PR #288, opened 2018-07-28
updated poll.go to show username.

mrbentarikau:
This is one way to show username if nickname is not set in Discord. Currently this if clause is executed even if nick is not set,


jonas747:
ah nice catch

jonas747:
actually can you swap them around so that it's

`err == nil && member.Nick != ""`

that way it wont panic if there's an error, and member is nil

PR #311, opened 2018-09-14
Some small but easily noticable changes

casKd-dev:


jonas747:
If you could put this in the custom.css file instead that would be better, otherwise its gonne be a pain to update bootstrap.

casKd-dev:
Did so, ready for a merge!

jonas747:
noice

PR #328, opened 2018-10-08
added target="_blank"

mrbentarikau:
Like previous html assets have, added target="_blank" to two of links here


jonas747:
noice

PR #341, opened 2018-10-20
Current Time Capitalisation

CHamburr:
It now accepts wrong capitalisation.


CHamburr:
Actually I don't think this is correct 🙁

PR #351, opened 2018-10-25
Some slight changes to Sbongebob.js.

mrbentarikau:
These changes will enable seeing what custom command is being deleted. Their ID and name.
This all can be altered later for other HTML templates.


jonas747:
could you change the text to

`Deleting x` instead of `x deleting this` 

Also having the defualt message be the same as the old one, since there are some actions that are not deletions (like reseting reputation)

mrbentarikau:
But I do understand that reputation part... did not know it was branched throw there, Songebob.js has to different parths same response is made.

jonas747:
Lokos good!

PR #352, opened 2018-10-27
Update Landing Page

CHamburr:
This includes typos, updates for YAGPDB, and incorrect links.


jonas747:
@mrbentarikau bold link tags has always worked for me

mrbentarikau:
yeah, they work both ways. still you don't put bold tag in front paragraph tag even in case you need all the paragraph bold : )

just aesthetics of HTML, nothing else

jonas747:
ah yeah true

PR #370, opened 2018-11-16
Added dogfact

CHamburr:
Are dogs better than cats?


CHamburr:
Also I got facts from [here](https://github.com/kinduff/dog-api/blob/master/db/seeds.rb) is that fine?

jonas747:
Looks good, thanks.

PR #401, opened 2019-01-15
Adds targetHasRoleId and targetHasRoleName functions

fifthfiend-kru:
This request adds the functions targetHasRoleID and targetHasRoleName, allowing for custom commands that can check any user's role information, rather than being specific to the user who called the command.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -251,7 +251,7 @@ func (c *Context) tmplTargetHasRoleID(target interface{}, roleID interface{}) bo
 		return false
 	}
 
-	ts := bot.GetMember(cs.Guild.ID, targetID)
+	ts := bot.GetMember(c.GS.ID, targetID)
 
 	role := ToInt64(roleID)
 	if role == 0 {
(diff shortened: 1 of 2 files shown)


jonas747:
Does this compile? It seems like it shouldn't?

fifthfiend-kru:
It's compiling as of now!

common/templates/context_funcs.go
```diff
@@ -241,6 +242,67 @@ func targetUserID(input interface{}) int64 {
...
+	if err != nil {
+		return false
+	}
+
+	role := ToInt64(roleID)
+	if role == 0 {
+		return false
+	}
+
+	c.GS.RLock()
```
jonas747: It's not needed to lock the guild state here, that's because bot.GetMember returns a copy from either the state or the api, not a direct reference, and while the role slice does point to the same array as the one in state, it's never modified (during MemberUpdate events the roles slice is instead completely replaced, meaning the copy would still hold the old role slice)

I notice i'm actually locking it myself in tmplHasRole but i think that was legacy code from when i directly referenced members from the state instead of copying them.

fifthfiend-kru: Fantastic, I'll fix this right away!

fifthfiend-kru:
Okay, that should fix it! Let me know if there's anything more I need to do!

jonas747:
noice

PR #409, opened 2019-02-11
idea... feature to have usability wise

mrbentarikau:
Because max cap of Custom Command characters is currently 10K and it's quite hard to keep eye on the total count of chars if one has sub-commands - this is just an idea of such implementation.


jonas747:
Good idea but should do it in JavaScript instead to have it update live

mrbentarikau:
yup, doing this client side would be more reasonable... this was just idea level PR actually, am not so eager to venture bootstrap and JS paths currently, and will close this request.

PR #412, opened 2019-02-19
temporary fix bot.go

mrbentarikau:
This is temporary fix ( and reminder : ) for displaying the list of custom commands and if custom command's trigger text is left empty (typically by choosing interval trigger type). It will not mess the list with ` characters.


jonas747:
noice

PR #418, opened 2019-03-10
PR in form of suggestion

mrbentarikau:
this could prevent thanks! spam, and then user's response "oh, it should've given rep via thanks"


jonas747:
Added a variant of this in 0c0470975b55ee8dc6e8dac48773bcda692b7024

PR #424, opened 2019-03-27
Large batch of minor changes again

ACA30:


ACA30:
Eyyy with updated commits we now have exactly 250 cat facts! :)

PR #426, opened 2019-03-28
autoMod adjustment

letto4135:
I added a third check to automod function PrepareMessageForWordCheck to leave the string as is and test the words in it as well. It continues to work for your test cases (slightly modified for third test case) as well as mine where I was trying to blacklist bot commands. It looks like you were intentionally looking at words only though so maybe there is a reason for that I'm not aware of.

Side note
If this ends up being my first successful pull request outside of school I might cry a little.
Great work on this bot. I really like it.

### DIFF
--- a/automod/automod.go
+++ b/automod/automod.go
@@ -138,10 +138,12 @@ func PrepareMessageForWordCheck(input string) string {
 		// this i imagine will solve a low of cases
 		w1 := ""
 		w2 := ""
+		w3 := ""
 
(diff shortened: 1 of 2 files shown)


jonas747:
Nice, looks good.

The reason i do this with words is to stop people being able to just throw some random symbol in the middle of the word to get around the word detection.

jonas747:
Al tough in the future i'd prefer it if you made pr's against the dev branch but eh

PR #440, opened 2019-04-29
idea for round function

mrbentarikau:
I am currently using this approach myself, maybe needs some changes


jonas747:
I would prefer it being 3 different functions instead

mrbentarikau:
Well I updated to separate round functions, left rounding to Even still there, don't know how necessary that one is... and all will return float64 - people can already convert toInt.

jonas747:
yeah can add something for that if needed in the future, but for now looks good

PR #447, opened 2019-05-11
usability suggestion for control_panel combined overview

mrbentarikau:
Rolecommands plugin has two limits, rolecommands and rolegroups count, they both are capped, It would be more informative to present both of these counts. Dunno whether on not to manually write that current max 1000 there or not,


jonas747:
add a if err != nil inbetween them

rolecommands/web.go
```diff
@@ -420,14 +420,18 @@ func (p *Plugin) LoadServerHomeWidget(w http.ResponseWriter, r *http.Request) (w
 	templateData["SettingsPath"] = "/rolecommands/"
 
 	numCommands, err := models.RoleCommands(qm.Where("guild_id = ?", g.ID)).CountG(r.Context())
```
jonas747: this error isnt handled

jonas747:
Looks good

PR #451, opened 2019-05-16
Update string to reduce confusion

buthed010203:
You cant use mod commands on users ranked the same or higher than you, changed the message to be accurate instead of saying "higher" it now says "the same as or higher"


jonas747:
Looks good

PR #455, opened 2019-05-27
Adapt to latest image

casKd-dev:
Things added/changed
- Added open ports by default (reverse proxy situations are defined in the other configurations)
- Add the ability to specify flags here ([ENTRYPOINT](https://docs.docker.com/engine/reference/#entrypoint) and [COMMAND](https://docs.docker.com/compose/compose-file/#command))

Related commit: https://github.com/casKd-dev/yagpdb/commit/8d950925d4ebad0314f24cf7d17e8c231a46d7f5

### DIFF
--- a/yagpdb_docker/docker-compose.yml
+++ b/yagpdb_docker/docker-compose.yml
@@ -13,6 +13,7 @@ services:
   app:
     image: caskd/yagpdb
     restart: unless-stopped
+    command: '-all -pa -exthttps=false -https=true'
     depends_on:
       - redis
       - db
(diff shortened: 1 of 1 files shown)


yagpdb_docker/docker-compose.yml
```diff
@@ -21,6 +22,9 @@ services:
     volumes:
       - cert_cache:/app/cert
       - soundboard:/app/soundboard
+    ports:
+      #- '80:80' # Uncomment if you want to use external reverse proxy
+      - '443:443'
```
jonas747: Why is the one above commented but not this?

casKd-dev: > Why is the one above commented but not this?

It was supposed to keep people away from using http only if they have no proxy but it should be uncommented because it redirects to https if that is the case.

PR #467, opened 2019-06-29
updated context_funcs.go footer formatting

mrbentarikau:
User's suggestion, footer does not have bold.


mrbentarikau:
sry, wanted to dev this

PR #468, opened 2019-06-29
updated context_funcs footer formatting

mrbentarikau:
User's suggestion, footer does not have bold. Now in correct branch.


mrbentarikau:
read already that this is intentional, will close

PR #474, opened 2019-07-21
docker-compose: Duplicate config across configurations, fix running

LoganK:
This effectively fixes the breakage of the proxied configuration introduced by
    3ab0453ac8f. (If you are running a reverse proxy, Docker will fail to start
    YAGPDB as the host ports will already be in use.)

    The cascading configuration was supposed to reduce the amount of maintenance,
    but does so at the cost of increased complexity. Some items, such as the port
    mappings, can't be overridden, only merged, so we either have to _always_
    require a hierarchy of configs (as-is currently the case) or we flatten the
    hierarchy and occasionally have to make changes in multiple files. As the
    former has already resulted in one breakage, we may as well go with simplicity.

    (An alternative case: a little bit of pain on the rare config change or a
    little bit of pain every time someone starts the server. This change reduces
    overall pain.)

    Other changes introduced in merging the config:
    * Change proxied to also use the caskd image
    * Fix dev ports to work like the current non-proxied config
    * Migrate command to non-abbreviated for improved diff output


LoganK:
Note that this includes the same commit from #473 to avoid silent merge failures. If there is disagreement on the decision there, this should be accepted without removal.

jonas747:
@casKd-dev Opinions?

yagpdb_docker/docker-compose.dev.yml
```diff
@@ -1,48 +1,56 @@
...
+
+services:
+  app:
+    build:
+      # We change context so that we can copy the local repo in during
+      # development
+      context: ../
+      dockerfile: yagpdb_docker/Dockerfile
+    restart: unless-stopped
+    command:
```
casKd-dev: The Dockerfile used here is different than the one used in the hub image so this is invalid as it would run `-all -pa -exthttps=false -https=true` without any executable

LoganK:
Thanks, @casKd-dev. Good catch. Fixed.

I'm personally okay just removing the dev compose config to reduce confusion at this point as I always run proxied and test via $(docker build -t caskd/yagpdb .) as it's easier than making local modifications to yet another file.

I'm personally not a fan of the entrypoint-as-app approach, it's generally against best practices, and it makes the security follow-up (mildly) more challenging. These may be previewed here:

https://github.com/LoganK/yagpdb/commit/2e7760410be5f4123f2e423bc4aebf65910b14e7
https://github.com/LoganK/yagpdb-1/commit/7e4aa8594b6a334ac730ba8a0e39b450f367f2dc

But I wanted to get back to a working config in the repo and not continue to stack changes awkwardly, first. 😄 


While you are commenting, I sent https://github.com/casKd-dev/yagpdb/pull/1 as a sort of hello to make sure the project was active. :-)

jonas747:
epic

PR #479, opened 2019-07-29
Added mod command

AlbertoMonteiro:
- I have added the mod function so we can easily calculate the modulus(Math.Mod)


jonas747:
nice

PR #488, opened 2019-08-15
toTime function for templates (name changed newDate)

mrbentarikau:
This was discussed today in # custom-command-discussion channel, albeit it's doable without specific function, maybe this one is simpler for users. I omitted nanoseconds and also *Location - left it as UTC.


jonas747:
Don't like the toTime name, rather something like, newDate maybe?

mrbentarikau:
newDate sounds OK, I'll change it

jonas747:
nice

PR #490, opened 2019-08-22
capitalize ID

mrbentarikau:
Based on consistency of the other template function names changed Id to ID.


jonas747:
Ah shoot yeah you're right, but can't change that now since it will break custom commands that uses it, instead make a duplicate with the correct version, for now atleast anyways.

I could think up some system to migrate custom commands when things like this does happen, hm.

PR #492, opened 2019-08-26
dbCount for templates

mrbentarikau:
users have requested this and I am using something similar. Optionally counts rows regarding given user_id; or all table rows of current guild.


jonas747:
Looks good

PR #493, opened 2019-08-26
revamped discord event handling revamped logging

jonas747:
There is now a proper structure for handling discord events, complete
with retrying running handlers on errors and proper logging, i've moved
around 30% of the handlers to the new system, the legacy system is still
there and i will move the rest over at a later date

Event data also contains guild state info, for guild events (minus guild
creates and deletes) the event handlers will not run if there's no state
info

I've moved from github.com/pkg/errors to emperrors, reason being
pkg/errors creates duplicate stacktraces even if there already was one
in one of the inner errors


jonas747:
need to go through every change but too lazy atm zzzzz

PR #497, opened 2019-08-30
slight changes in presenting custom commands

mrbentarikau:
the colon between .TriggerType and .TextTrigger already helps to separate default text better and increases readability. 
I added <span> for users who want to alter CSS styling further on their side - I know it deviates from bootstrap styling, but that <span> does not kill anything : )


jonas747:
use `cc-text-trigger-span` instead

mrbentarikau:
hope I understood correctly, changed span class name

mrbentarikau:
for the future, I guess .NextRun for .TriggerType 5 (time intervals) would be cool to implement, but probably javascript thingy

PR #499, opened 2019-09-01
Function to delete all message reactions

mrbentarikau:
users have mentioned the need for this, it's blunt force removal - not specific to reactionName or userID, but maybe necessary


jonas747:
add a ```go
if c.IncreaseCheckGenericAPICall() {
		return nil, ErrTooManyAPICalls
}
```
under the channel id check

mrbentarikau:
updated the code, I used tmplDelMessage as reference, that's why that APICall checker function was missing

jonas747:
looks good

PR #504, opened 2019-09-08
cc: append item to slice with size limit

Heimdall409:
Adds a template function to add an item to a slice, with size limit to prevent abuse.
The size limit uses the same logic as database entries, so that the user can retrieve database entries iteratively.


jonas747:
Multiply by 1k instead of 50

jonas747:
Ideally it would take in `interface{}` as the slice argument, then you can use reflect.Append after checking types etc

PR #509, opened 2019-09-16
function countAllBots

mrbentarikau:
There is onlineCount to get all online users. This one will distinguish bots, having optionable bool argument which by default is true and counts all bots, but set to false, counts only online bots.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -855,14 +855,16 @@ func (c *Context) tmplOnlineCountBots() (int, error) {
 	}
 
 	botCount := 0
+	
 	c.GS.RLock()
+	defer c.GS.RUnlock()
+	
 	for _, v := range c.GS.Members {
 		if v.Bot && v.PresenceSet && v.PresenceStatus != dstate.StatusOffline {
 			botCount++
 		}
 	}
-	c.GS.RUnlock()
-
+	
 	return botCount, nil
 }
 
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -849,7 +849,7 @@ func (c *Context) tmplOnlineCount() (int, error) {
 	return online, nil
 }
 
-func (c *Context) tmplBotCount() (int, error) {
+func (c *Context) tmplOnlineCountBots() (int, error) {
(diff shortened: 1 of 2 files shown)


jonas747:
this wont work on "large" servers (above 2.5k) since the bot wont have offline members cached

for now you could just do onlineCountBots

mrbentarikau:
did not know of that, basically it makes this function pointless in general concept

mrbentarikau:
so maybe this, changed the code more simple

jonas747:
change it to onlineCountBots and make sure they're online (essentially the same as onlineCount but with the bot check) while generally offline members aren't cached, they can be in some cases.

mrbentarikau:
so something like this?

common/templates/context_funcs.go
```diff
@@ -848,3 +848,21 @@ func (c *Context) tmplOnlineCount() (int, error) {
...
 	return online, nil
 }
+
+func (c *Context) tmplOnlineCountBots() (int, error) {
+	if c.IncreaseCheckCallCounter("online_bots", 1) {
+		return 0, ErrTooManyCalls
+	}
+
+	botCount := 0
+	c.GS.RLock()
```
jonas747: Add `c.GS.RUnlock()` just incase shit breaks

(and also remove the lock down below offcourse)

jonas747:
one final change

common/templates/context_funcs.go
mrbentarikau: I actually don't understand what to change here, it's straight out referenced from tmplOnlineCount .... and c.GS.RLock() before the for range  on GuildState and thereafter c.GS.Unlock().

jonas747: yeah i might have forgotten about that aswell, if it panics within the loop here it will never unlock the guild state, and the shard would eventually get stuck on trying to update it causing that shard to die, so add a defer unlock to make sure that dosen't happen and we can recover from it

mrbentarikau: I added the defer for c.GS.RUnlock() - c.GS.RLock() stays as it is?

jonas747:
looks good now

PR #518, opened 2019-09-29
square root function

mrbentarikau:
Although it can already be done using current custom command functions, "real" square root is still more reasonable. I used type switching to get always NaN answer if not a number instead of 0.


jonas747:
no need for variadic args in this function is there?

mrbentarikau:
I left the variadic there to even respond if no arguments are given with Nan. But removed it now.

jonas747:
looks good

PR #533, opened 2019-10-06
Add ttc as alias for toggletimeconversion

citizenofcode:
Self explanatory, all the aliases for toggletimeconversion are pretty long, it would be nice if there were a shorter alias such as ttc.


jonas747:
noice

PR #535, opened 2019-10-15
delete Nicknames

mrbentarikau:
An idea-level PR - enabling user to delete their own nickname records from the guild. It either deletes all nicknames till the last one recorded or if user does not have any active nicknames then everything.

This is idea-level because I don't know the general plan on handling username/nickname recordings.

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -270,3 +270,21 @@ func GetNicknames(ctx context.Context, userID, guildID int64, limit, offset int)
 		qm.Limit(limit),
 		qm.Offset(offset)).AllG(ctx)
 }
+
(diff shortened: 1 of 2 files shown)


jonas747:
Should probably be done in the control panel, in the user menu perhaps as a new entry "user data"

mrbentarikau:
but that way regular users don't have access nor control over their nickname or even username data, if it's on control panel and tied to guild admins.

jonas747:
it wouldn't be tied to a server, anyone can log in...

jonas747:
same as the premium page

mrbentarikau:
yes, I understand it now... hmm, will close this then.

PR #536, opened 2019-10-16
Replace findUserWithNickname() with findUsers() in RSVP

10by10pixel:
`findUsers()` is the same function but doesn't add nicknames. This is important for reducing ambiguity in event embeds. If `Jonas747#0001` and `(O1)#0001` both have the nickname "Narutp". The nickname version would return
```
narutp#0001
narutp#0001
```
and my version would return
```
Jonas747#0001
(O1)#0001
```


10by10pixel:
Closing this and trying again.

PR #537, opened 2019-10-17
RSVP shouldn't use nicknames

10by10pixel:
Replaces `findUserWithNickname()` with `findUsers()` which doesn't use nicknames. This makes things less ambiguous as to whether the bot is using username + discrim or nickname + discrim.


jonas747:
Rename it to `findUser` since its singular

10by10pixel:
@jonas747 Done.

jonas747:
epic

PR #539, opened 2019-10-19
Change Colors to YAGPDB Theme & Consistency Fixes

ACA30:
## Change colors:
 to 
 to 
 to 

This aligns better with the YAGPDB colour scheme while maintaining the ability to quickly see what the status of nodes is, and also makes it easier on the eyes with less neon colors.

 to 
 to 
 to 

## Consistency Fixes:
Might as well capitalize all words on the cards or none of the words on the cards for consistency, if you want the alternative to the PR, let me know.


buthed010203:
Its already impossible to read on light theme just so you know.
Changing the color to a white works fine on dark mode tho
\#b6b6b6 works well on light theme while \#1c2025 is nice on dark but then you cant read the yellow so i would honestly just use the same color for both (\#B6B6B6)

C0braD3v:
> 
> 
> those colors are a bit too dark, hard to read on low end screens such as my laptop

Definitely agree here. Less users use light theme, so this seems less of an issue then having low end screens and having the colors appear dark.

PR #543, opened 2019-10-25
Satty9361 channel topic template

Satty9361:
Added a template to allow changing of channel topic. Have reused the same counters as used in tmplEditChannelName but probably might want to increase count for edit per channel per cc or use a different per channel tracker for topic change depending on review.


jonas747:
nice

PR #546, opened 2019-11-02
Satty9361 edit nickname template

Satty9361:
Added a new template to allow editing nickname of the user triggering a command. Editing nicknames are a very popular request however ability to edit nickname of any target user has misuse potential . Hence , implemented it to only work on triggering user similar to sendDM. 

Introduced a new checker "edit_nick" which allows only two calls per cc instance similar to tmplEditChannelName . 

Also, tried to save unnecessary API call if present nickname is already same as new nickname.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -877,20 +877,14 @@ func (c *Context) tmplEditNickname(Nickname string) (string, error) {
 		if c.MS == nil {
 		return "", nil
 	}
-
-	c.GS.RLock()
-	gID := c.GS.ID
-	memberNick := c.MS.Nick
-	memberID := c.MS.ID
-	c.GS.RUnlock()
 	
-	if strings.Compare(memberNick, Nickname) == 0 {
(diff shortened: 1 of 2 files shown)


common/templates/context_funcs.go
```diff
@@ -868,3 +868,32 @@ func (c *Context) tmplOnlineCountBots() (int, error) {
...
+
+	if c.IncreaseCheckCallCounter("edit_nick", 2) {
+		return "", ErrTooManyCalls
+	}
+
+		if c.MS == nil {
+		return "", nil
+	}
+
+	c.GS.RLock()
```
jonas747: No need to lock here, guild id is immutable and c.MS is a copy of the member when the cc started, not a reference.

Satty9361: Made the changes. I was using sendDM as reference for some parts and it had the lock hence i thought it might be necessary.

jonas747: Yeah some pieces are old, used to be like that before but yeah

jonas747:
Seems fine

PR #547, opened 2019-11-05
dbCount Addon to return total count per key

Satty9361:
Made some changes to the dbCount function which presently could only return total count per "UserID" and extended to also return total count per "Key" depending on optional argument datatype.


jonas747:
Have you tested this?

Satty9361:
Yes I have tested this in my self hosted bot and it works just as expected.

PR #548, opened 2019-11-06
subtraction function

mrbentarikau:
For some reason users want this, but yeah, it completes the basic math palette, so why not : )


jonas747:
simple enough, i mean you can just use the add function with negative numbers, but alright

mrbentarikau:
Yeah, this has been discussed many times in support server, but users are still lazy and say add -1 is confusing

PR #549, opened 2019-11-07
Specifying Number of days of message to delete while executing ban command

Satty9361:
Made some changes to enable setting the number of days of messages to delete while executing the ban command.


diligamer:
Discord limits bots like yag from deleting messages that are older than 2 weeks.

Satty9361:
> Discord limits bots like yag from deleting messages that are older than 2 weeks.

Check again, I know that the endpoint only accepts values between 0 and 7 days and i have implemented a check for that @diligamer .

jonas747:
Looks good, i really need to rework the monster that is the punish function tough...

PR #552, opened 2019-11-10
Small update to not send blank DMs when response only consists of whitespace for Punishments.

Satty9361:
Dont send DM if response only consists of whitespace. This can also be used to smartly disable ban/mute/unmute/kick/warn DMs while still allowing to do any kind of processing like maintaining Records in database upon Mod Action that the user wants to do.


jonas747:
nice

PR #557, opened 2019-11-13
Clean: Optionally Ignore pinned messages

Satty9361:
Ability to optionally ignore pinned messages while executing clean command with a new "-pin" flag.
Seems to work fine for me.

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -312,7 +312,7 @@ var ModerationCommands = []*commands.YAGCommand{
 		CustomEnabled:   true,
 		CmdCategory:     commands.CategoryModeration,
 		Name:            "Clean",
-		Description:     "Delete the last number of messages from chat, optionally filtering by user, max age and regex.",
+		Description:     "Delete the last number of messages from chat, optionally filtering by user, max age and regex or ignoring pinned messages.",
 		LongDescription: "Specify a regex with \"-r regex_here\" and max age with \"-ma 1h10m\"\nNote: Will only look in the last 1k messages",
 		Aliases:         []string{"clear", "cl"},
 		RequiredArgs:    1,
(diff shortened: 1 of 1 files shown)


moderation/commands.go
```diff
@@ -778,6 +799,11 @@ func AdvancedDeleteMessages(channelID int64, filterUser int64, regex string, max
 			continue
 		}
 
+		// Check if pinned message to ignore ; no need to check if pinEnableFilter bool is true because if false , map will be empty.
+		if _, found := pinnedMessages[msgs[i].ID]; found {
```
jonas747: pretty sure this will panic if `pinFilterEnable` is false, since `pinnedMessages` will be nil

moderation/commands.go
```diff
@@ -324,6 +324,7 @@ var ModerationCommands = []*commands.YAGCommand{
 			&dcmd.ArgDef{Switch: "r", Name: "Regex", Type: dcmd.String},
 			&dcmd.ArgDef{Switch: "ma", Default: time.Duration(0), Name: "Max age", Type: &commands.DurationArg{}},
 			&dcmd.ArgDef{Switch: "i", Name: "Regex case insensitive"},
+			&dcmd.ArgDef{Switch: "pin", Name: "Ignore pinned messages"},
```
jonas747: maybe `-nopin` instead?

moderation/commands.go
Satty9361: Reading from empty maps dont cause panic iirc but writing would cause a runtime panic. Ill still wrap it up in an if statement.

moderation/commands.go
Satty9361: done

moderation/commands.go
jonas747: ah yeah you're right

PR #560, opened 2019-11-15
Satty9361 list violations

Satty9361:
Added a list violations moderation function. List all past violations of user ordering them by id , has an option to skip entries. Also included a summary mode and max age optional flags.
Permissions needed are manage messages permissions or roles allowed to give warnings.

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -1,6 +1,7 @@
 package moderation
 
 import (
+	"context"
 	"fmt"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jonas747:
if you want you can turn this into a paginated command (see https://github.com/jonas747/yagpdb/blob/master/logs/plugin_bot.go#L216 for reference), otherwise its fine.

Satty9361:
I had that in mind as well but was unsure because warnings isnt paginated. Ill work on this a bit more and make it paginated.

Satty9361:
Summary mode wouldn't work well with a paginated log so separated it into a separate command. Added a Paginated Violation Log ( list violations) command.

moderation/commands.go
```diff
@@ -602,6 +606,150 @@ var ModerationCommands = []*commands.YAGCommand{
...
+			maxAge := parsed.Switches["ma"].Value.(time.Duration)
+			skip := parsed.Switches["skip"].Int()
+			if skip < 0 {
+				skip = 0
+			}
+			
+			// retrieve Violations
+			qms := []qm.QueryMod{qm.Where("guild_id = ?", parsed.GS.ID), qm.OrderBy(order), qm.Limit(limit), qm.Offset(skip)}
+			
+			if userID != 0 {
```
jonas747: can do the same with max age, example: `qm.Where("created_at > ?", time.Now().Add(-maxAge))`

Satty9361: Yeah that looks like a much better and more consistent method. Made the changes. I plan to make some changes to the pagination method to not throw an error(Something went wrong when running this command, either discord or the bot may be having issues.) in first run when there are no results but ill put that as a separate PR later. This seems to be the case with some other commands like automod logs as well.

jonas747:
Looks good

PR #562, opened 2019-11-21
getChannel template

Satty9361:
Users have requested for this. It is already possible with execCC with channelID followed by .Channel, this makes the process less convoluted. 

The lock argument in ChannelCopy should probably be false ? ( since ChannelArg already has a Rlock) however tmplRunCC has lock set to true hence i have kept it as true for now.


jonas747:
Looks good

PR #563, opened 2019-11-23
exec/execAdmin with execCC fix

Satty9361:
made these changes so that an exec/execAdmin following an execCC runs in the new channel. 

Have not incorporated channel update for dm templates ( eg: punishment dm templates) to prevent loops (eg: warn -> warn dm execCC -> execAdmin warn ...) .

### DIFF
--- a/commands/tmplexec.go
+++ b/commands/tmplexec.go
@@ -83,7 +83,7 @@ func TmplExecCmdFuncs(ctx *templates.Context, maxExec int, dryRun bool) (userCtx
 				messageCopy.ChannelID = ctx.CS.ID
 			}
 		}
-		mc := &discordgo.MessageCreate{ctx.Msg}
+		mc := &discordgo.MessageCreate{&messageCopy}
 		if maxExec < 1 {
 			return "", errors.New("Max number of commands executed in custom command")
 		}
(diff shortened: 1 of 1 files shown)


commands/tmplexec.go
```diff
@@ -77,7 +77,13 @@ type cmdExecFunc func(cmd string, args ...interface{}) (interface{}, error)
 // Returns 2 functions to execute commands in user or bot context with limited about of commands executed
 func TmplExecCmdFuncs(ctx *templates.Context, maxExec int, dryRun bool) (userCtxCommandExec cmdExecFunc, botCtxCommandExec cmdExecFunc) {
 	execUser := func(cmd string, args ...interface{}) (interface{}, error) {
-		mc := &discordgo.MessageCreate{ctx.Msg}
+		messageCopy := *ctx.Msg
+		if ctx.Msg.ChannelID != ctx.Msg.GuildID { //punishment dms shouldnt be able to execCC followed by exec/execAdmin to prevent loops
```
jonas747: Don't think this is needed, as the moderation commands only returns when the punish dm has been sent, can there can only be 1 instance of each command per user being executed

Satty9361: Ah okay , removed the check, it is still necessary to check if CS is a nil pointer before accessing CS.ID so kept that check. It is nil pointer in some cases like direct exec/execAdmin within punishment dms ( i got a runtime panic when i was running without this condition earlier ).

Satty9361:
Do you need any other changes @jonas747  , I had made the changes you had suggested last time.

jonas747:
hmm any reason why it's only for execUser and not execAdmin?

Satty9361:
I have added for both , execUser as well as execBot( execAdmin).

jonas747:
ah my bad didn't see that

PR #565, opened 2019-11-25
updated customcommands.html

mrbentarikau:
I really think these things are necessary for CCs. 

Adding to CC Interval triggers - the time when they are next run, in server's locale. Did not use JavaScript there, can't see actual reason why. The layout formatting needs discussion due mobile screens.  (edit2: nothing to discuss actually, this is bootstrap .pull-right with frustrating-to-do vertical align and trigger text can actually be way longer than this Interval content is : )

Also because CCs response "size" is actually calculated using len function in customcommands.go returning bytes, my version has corrected character count (corrected, because Unicode emojis were not counted as one char) and byte count.

All in picture
ccUpdate

### DIFF
--- a/customcommands/assets/customcommands.html
+++ b/customcommands/assets/customcommands.html
@@ -189,7 +189,7 @@
             </div>
             <div class="row mb-2">
                 <div class="col-lg-12">
-                    <label for="responses">Response (<span class="cc-length-counter">x</span>/10000)</label>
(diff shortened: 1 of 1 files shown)


jonas747:
The size thing should be updated on the backend to count runes instead of bytes

mrbentarikau:
There are two counters in real go code yes...
First one inside customcommands.go uses len (bytes) and redirects it further to validation.go ValidateTemplateField() which actually has rune counter - this is already pointless, because len/bytes is counted first.....

and my suggestion here is to improve already broken Frontend thing

jonas747:
Noice

PR #569, opened 2019-11-26
Update handlers_general.go

buthed010203:
print max and min shard uptimes in seconds rather than ns
idk what im doing i prob broke this


jonas747:
yeahhh thats not gonna work

buthed010203:
didnt think so lmfao

PR #570, opened 2019-11-27
Pagination Error , typo & other fixes/additions

Satty9361:
Pagination Error Patch : When specifying a page number in paginated messages and there is no result, it displays the following error : 

> "Something went wrong when running this command, either discord or the bot may be having issues."

It also causes custom commands to stop execution due to error generation. This is an idea to fix this behavior. With above changes, error will not be generated during first run. Max page is also appropriately updated when going to a lower page with a simple change in paginatedmessage.go . Same method can be adopted in other paginated commands that allow specifying page like amod logs.

Pictorial Representation : left button -> left button -> left button -> right button.
image
image
image
image

Other Changes : 
Added a Total field to ListViolationsCount
Typo fix /n -> \n and Capitalization Change

### DIFF
--- a/automod/commands.go
+++ b/automod/commands.go
@@ -191,7 +191,7 @@ func (p *Plugin) AddCommands() {
 			}
 
 			for name, count := range violations {
-				out += fmt.Sprintf("Violation: %-20s count: %d\n", name, count)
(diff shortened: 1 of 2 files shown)


jonas747:
nice

PR #575, opened 2019-12-15
Satty9361 pagination patch rest

Satty9361:
Extending the paginated error patch to other similar commands ; referenced from PR #570 .


jonas747:
Looks good

PR #576, opened 2019-12-18
Idea: moderation command TopWarnings

mrbentarikau:
Command that will give ranked list of all guild's user warnings, does not create a web-based listing - not necessary for mod commands. Also, currently kept the additional function and type in separate file because really can't find better place to add the code to, maybe plugin_bot.go

### DIFF
--- a/moderation/topwarnings.go
+++ b/moderation/topwarnings.go
@@ -1,0 +1,70 @@
+package moderation
+
+import (
+	"database/sql"
+
+	"github.com/jonas747/discordgo"
+	"github.com/jonas747/dstate"
+	"github.com/jonas747/yagpdb/bot"
+	"github.com/jonas747/yagpdb/common"
+)
+
+type WarnRankEntry struct {
+	Rank      int    `json:"rank"`
+	UserID    int64  `json:"user_id"`
+	Username  string `json:"username"`
+	WarnCount int64  `json:"warn_count"`
+}
+
+func TopWarns(guildID int64, offset, limit int) ([]*WarnRankEntry, error) {
+	const query = `SELECT rank, warn_count, user_id FROM
+	(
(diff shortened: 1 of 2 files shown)


moderation/topwarnings.go
```diff
@@ -0,0 +1,70 @@
...
+	Rank      int    `json:"rank"`
+	UserID    int64  `json:"user_id"`
+	Username  string `json:"username"`
+	WarnCount int64  `json:"warn_count"`
+}
+
+func TopWarns(guildID int64, offset, limit int) ([]*WarnRankEntry, error) {
+	const query = `SELECT rank, warn_count, user_id FROM
+	(
+		SELECT RANK() OVER (ORDER BY count(message) DESC) AS rank, count(message) as warn_count, user_id
```
jonas747: why count(message) instead of count(*)?

moderation/topwarnings.go
```diff
@@ -0,0 +1,70 @@
...
+		var rank int
+		var tmp []*dstate.MemberState
+		var userID int64
+		var warncount int64
+		var err = rows.Scan(&rank, &warncount, &userID)
+		if err != nil {
+			return nil, err
+		}
+
+		tmp, err = bot.GetMembers(guildID, userID)
```
jonas747: Maybe use GetUsers instead since that works even if they leave

moderation/topwarnings.go
mrbentarikau: yeah, that was just random column name I used, will edit later to *

moderation/topwarnings.go
mrbentarikau: maybe it's more reasonable to show ranking only for active members in guild

mrbentarikau: updated for further ideas

jonas747:
lets merge this in

mrbentarikau:
roger!

PR #577, opened 2019-12-19
possible typo fix on database query type

mrbentarikau:
This might be a typo, because stringedKey value is of type string and database query wants BIGINT. Still works at present state if user presents the "key" as integer - or is this intentional?


jonas747:
good catch

PR #580, opened 2020-01-04
QoL changes

Satty9361:
targetUserID can handle mentions and getMember uses targetUserID instead of direct ToInt64.


jonas747:
looks good

PR #582, opened 2020-01-06
Slice as new Datatype and Append function update

Satty9361:
Implemented Slices as a separate datatype and reworked the Append function as a function of this new datatype. This should hopefully make it safe to have an append function without danger of mutating existing slices.

From my testing there is no problem with cembed, also the datatype is preserved after Append function or using "slice" function as well.


jonas747:
looks good

PR #586, opened 2020-01-08
Update swearwords.go

xerpiny:
its our nation's bad words


jonas747:
Not sure if i wanna extend it to swearwords in other languages

PR #587, opened 2020-01-08
quickfix for incorrect punishment DMs

mrbentarikau:
Situation: if  user was not punishable due permissions not allowing it, and YAG getting even 403 Forbidden, DM was still sent out to that user that they're kicked/banned.

So to resolve that spam case, just moving if !memberNotFound { ...
into other error checking branch could resolve this

### DIFF
--- a/moderation/punishments.go
+++ b/moderation/punishments.go
@@ -75,14 +75,7 @@ func punish(config *Config, p Punishment, guildID, channelID int64, author *disc
 	gs := bot.State.Guild(true, guildID)
 
 	member, memberNotFound := getMemberWithFallback(gs, user)
-	if !memberNotFound {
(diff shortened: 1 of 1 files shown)


jonas747:
Can't do this, there's a reason the dm is sent before they're kicked/banned and thats because the bot can't dm people it dosen't share a server with, which after they've been kicked is somewhat likely.

mrbentarikau:
those punishment cases are yeah resolved above, but currently thing is still that bot will spam DMs if there's no valid reason for it, so this needs some sort of a solution.

jonas747:
yeah would need to implement a check for bot perms and try to predict all cases where the bot can't kick them, or just notify them by sending another message in their dms that they were not kicked or something, im closing this for now tough.

PR #589, opened 2020-01-10
Clear Violations

Satty9361:
A command to clear violations of an user similar to ClearWarnings with some filters and conditions. This should more or less finish the Violation related commands. I added a GuildScopeCooldown to prevent spamming of command because this command is probably a bit database heavy? Also i first fetched the rows then deleted them because "OrderBy" didnt seem to work with direct DeleteAll.


jonas747:
Looks good, nice work

PR #590, opened 2020-01-12
addition to clean command: flag minAge

mrbentarikau:
Users have requested this quit a lot, it works well in conjunction with other flags, only thing probably is the name of the flag : )


jonas747:
Looks good, in hindsight the naming for the switches here is really bad, need to fix that, but for now call it "minage" instead since you're only removing 2 letters and adding a lot of cognitive load.

mrbentarikau:
consistency sake, because clearviolations command has "mina" flag, shouldn't this be set here as well? Should I change back from minage to mina?

jonas747:
Yeah it should be changed there aswell, ill have to imeplement a system for switches wchich keeps deprecated aliases working, so that stuff and people don't break.

mrbentarikau:
roger that!

PR #591, opened 2020-01-12
Satty9361 slice addon

Satty9361:
With slice as a new datatype, some additional functions to change value at an index of slice or appending two slices together will be helpful.


jonas747:
Looks good

PR #593, opened 2020-01-14
Comment out placeholder text.

10by10pixel:
Idea implemented based off @Satty9361 's idea.


Satty9361:
Might be a good solution yeah to at-least prevent the response from getting printed and users complaining about the bot being bugged. Jonas is ill though so this will probably take time till he reviews :)).

jonas747:
People will try to replace that comment text and start wondering why the command isn't printing that as the output, too misleading IMO

PR #594, opened 2020-01-14
Update customcommands-editcmd.html

buthed010203:
remove the ! on the !fun placeholder for the default trigger as it was confusing some people
change the default body placeholder from ^ smells to Command body here for the same reason


jonas747:
Smort

PR #595, opened 2020-01-14
addition: .CCRunCount

mrbentarikau:
Have seen users wanting this, could be useful info outside control panel, so why not add it to the custom commands dot. That +1 is there to count current "launch" as well.

currently users implement their CCcounters through database, so it lessens that a little.


jonas747:
These are cached and the cache isn't purged when the counter increases so it will show outdated  numbers, but i guess i can still merge this.

PR #598, opened 2020-01-17
Idea: Delete Role Commands from Role Groups

mrbentarikau:
This here includes HTML code (without any additional CSS styling) for delete button, which is displayed if there are role commands set in a group + counter of rolecommands in that group. 

HandleDeleteRoleCommands, when invoked will delete the rolecommands only from within that group, group itself still remains there. Also deleting from Ungrouped is possible.

Should not be database heavy.


jonas747:
epic

PR #599, opened 2020-01-18
Suggestion: Custom Commands frontpage HTML change

mrbentarikau:
This is the version I use for my fork.
It has code folding option and also working edit/delete buttons on frontpage. Line numbers are still there.
It has some forced CSS-styling with !important tags; for margins and padding - probably can be done in bootstrap. Also added back the title for delete button and span classes I had PRd before.

Probably needs to be run in YAG-testing first for feedback. Hope this concept works.

### DIFF
--- a/customcommands/assets/customcommands.html
+++ b/customcommands/assets/customcommands.html
@@ -91,7 +91,7 @@
                                 </div>
                                 <div class="form-group">
                                     <button type="submit"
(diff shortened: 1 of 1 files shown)


mrbentarikau:
updated abother HTML section, deleting group button.
The button really did not fit to mobile screen... Just moved that extra text to title tag so it's displayed on hover and also in alert. Saves screen space.

jonas747:
looks good

PR #600, opened 2020-01-18
Satty9361 punish data

Satty9361:
Some changes to make .Channel and .Message available for punishment DMs( here these are the message/channel in which the Punishment is triggered either manually or by automod). This is especially useful when people want to send an extra message to the triggered channel or perform certain actions depending on content of original message and author of triggering message or send a custom DM depending on content of triggered message so on.

Lots of small changes, I hope i didnt forget anything.


jonas747:
Looks good

PR #602, opened 2020-01-18
Format and Consistency Fixes

Satty9361:
Just some prettying up and consistency fixed.


jonas747:
normally i wouldn't let through the switch name change, but since its so new anyways i'm gonna allow it

PR #603, opened 2020-01-20
template function: exponential power and modulo edit

mrbentarikau:
We have square root, let's have some power :)
Also had to edit tmplMod "error of args length" return, because 0 is valid value for mod's return


mrbentarikau:
after reading the discussion about separate PRs - even for small changes, will not edit tmplMod here

jonas747:
Noice

PR #604, opened 2020-01-21
Satty9361 send complex

Satty9361:
A new template that creates *discordgo.MessageSend from sdict , maps etc. 

Some changes to sdict conversion now using reflection.

Enable cemebed to handle *discordgo.MessageEmbed sent as argument

Changes to SendMessage function effectively enabling all existing templates ( sendMessage , sendMessageRetID etc.. ) to send Complex messages depending of datatype of argument passed being *discordgo.MessageSend.

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -42,8 +42,8 @@ func StringKeyDictionary(values ...interface{}) (SDict, error) {
 			return nil, errors.New("Sdict: nil value passed")
 		} 		
 
-		if val.Type().String() == "templates.SDict" {	
-			return val.Interface().(SDict), nil
+		if sdict, ok := val.(SDict); ok {	
+			return sdict, nil
 		}
 		
 		switch val.Kind() {		
@@ -130,7 +130,7 @@ func CreateEmbed(values ...interface{}) (*discordgo.MessageEmbed, error) {
 	return embed, nil
 }
 
-func CreateMessage(values ...interface{}) (*discordgo.MessageSend, error) {
+func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 	if len(values) < 1 {
 		return &discordgo.MessageSend{}, nil
 	}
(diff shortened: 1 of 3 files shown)


common/templates/general.go
```diff
@@ -33,19 +35,38 @@ func Dictionary(values ...interface{}) (map[interface{}]interface{}, error) {
...
-			return SDict(mapCopy), nil		
-		default:
-			return nil, errors.New("invalid dict call")		
-		}	
+		val, isNil := indirect(reflect.ValueOf(values[0]))
+		if isNil || values[0] == nil {
+			return nil, errors.New("Sdict: nil value passed")
+		} 		
+
+		if val.Type().String() == "templates.SDict" {
```
jonas747: change to 
```go
if sdict, ok :=val.(SDict); ok {
    return sdict, nil
```

common/templates/general.go
```diff
@@ -107,6 +130,58 @@ func CreateEmbed(values ...interface{}) (*discordgo.MessageEmbed, error) {
 	return embed, nil
 }
 
+func CreateMessage(values ...interface{}) (*discordgo.MessageSend, error) {
```
jonas747: rename to "createMessageSend"

Satty9361:
Noted. I will implement these changes. Also, will split the prs in future . I wanted the sdict to be able to handle dict with string keys and so on but yeah this could be in a separate pr. Do you also want me to change the mapping to cmessagesend or cmessage is fine?

common/templates/general.go
Satty9361: I think you meant val.Interface().(SDict) because val is reflect.Value and not interface and cannot be typecasted. I have updated according to my understanding.

jonas747:
maybe complexMessage instead, no need to short it imo

Satty9361:
Ok done!

jonas747:
looks good

PR #611, opened 2020-01-24
Fix dbBottomEntries bug

Satty9361:
There was a bug since last update. dbBottomEntries should still sort values in ascending order.

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -532,7 +532,7 @@ func tmplDBCount(ctx *templates.Context) interface{} {
 func tmplDBTopEntries(ctx *templates.Context, bottom bool) interface{} {
 	orderBy := "value_num DESC, id DESC"
 	if bottom {
-		orderBy = "value_num ASC, id ASC"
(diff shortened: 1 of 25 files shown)


customcommands/tmplextensions.go
```diff
@@ -532,7 +532,7 @@ func tmplDBCount(ctx *templates.Context) interface{} {
 func tmplDBTopEntries(ctx *templates.Context, bottom bool) interface{} {
 	orderBy := "value_num DESC, id DESC"
 	if bottom {
-		orderBy = "value_num DESC, id ASC"
+		orderBy = "value_num ASC, id ASC"
```
jonas747: Maybe also invert id ASC so that its the opposit of topentries?

Satty9361:
I am probably missing something but the order is already opposite. It is id DESC in topentries and id ASC in bottomentries. Do you want me to make both same ?

jonas747:
Ah nvm then, thats's correct

Satty9361:
Done

jonas747:
sorry i already did it on my end since i wanted to get it out fast

PR #613, opened 2020-01-26
StringSlice method to convert a Slice to []String

Satty9361:
Quite a few template like joinStr, AddReactions , exec etc accept []string as arguments but not []interface{}. Rather than doing a huge number of changes, I think conversion of Slice to []slice is a good solution. This does not allow any mutation of string slices, just an easy conversion from Slices to []string. I have implemented this using a variadic boolean flag argument which is by default false but can be set to true. If true, and any value of the Slice is not string, execution is stopped and it returns nil. Otherwise it tries to generate a sub-slice of all string values in the Slice.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -2,6 +2,7 @@ package templates
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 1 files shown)


common/templates/context.go
```diff
@@ -518,3 +518,25 @@ func (s Slice) AppendSlice (slice interface{}) (interface{}, error) {
...
+	strict := false
+	if len(flag) > 0 {
+		strict = flag[0]
+	}
+
+	StringSlice := make([]string, 0, len(s))
+	
+	for _, Sliceval := range s {
+		switch t := Sliceval.(type) {
+			case string :
```
jonas747: Add a case fo fmt.Stringer aswell

Satty9361: Added case, made it work for strict = false (default) case because i guess that makes more sense.

jonas747:
looks good

PR #615, opened 2020-01-28
Tab-key for custom commands

mrbentarikau:
Like all front-end things this is "what-if"-case. This implementation does not have a keyboard "undo" but backspace still works. "Tab" creates 8 character width and this is default.

character addition to CC total counter is added

Have run this on Safari, Edge, Chrome, Firefox, Opera, Brave.

needs testing in dev and later this class could be added to other textarea sections in HTML


jonas747:
noice

PR #616, opened 2020-01-28
templates: logarithm function

mrbentarikau:
This should be the last addition for algebra palette.
It operates on base e, meaning natural logarithm. In mathematics, log_n is most often taken to be the natural logarithm and so does golang's math library.

if user gives second argument as base, return will be for that base.


jonas747:
Looks good

PR #617, opened 2020-01-30
Satty9361 soundboard reset

Satty9361:
Debug and Maintenance command for Soundboard, to reset queue and disconnect player.

This would be pretty helpful in general because currently there is no control over the bot while its in the voice channel or has many queued up files.

Hope the implementation is alright.

### DIFF
--- a/soundboard/bot.go
+++ b/soundboard/bot.go
@@ -85,7 +85,7 @@ func (p *Plugin) AddCommands() {
 		CmdCategory: commands.CategoryFun,
 		Name:        "SoundboardReset",
 		Aliases:     []string{"sbclose", "sbReset"},
-		Description: "Reset Soundboard",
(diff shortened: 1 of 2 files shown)


jonas747:
Right now this is more like a skip instead of reset, maybe also reset the queue?

Satty9361:
From my initial testing it did reset the queue as well as disconnect player. 

Flow is this:

set quit = true and broadcast.

if playing sound :
Stop playing sound and return to Run () .

Then check waitForNextElement()

here if pmstop is true 
p.exit is directly invoked.

since current vc is passed while returning to Run() from playSound, it is disconnected .

p.GuildID is deleted from players.

I assumed it is not necessary to set queue to nil anymore however ill do some more testing when im back home.

jonas747:
Looks good

PR #618, opened 2020-02-01
Satty9361 mute update

Satty9361:
Add ability to disallow reaction add while muting. (this prevents reaction spam). It can be handled by permissions but not the most efficiently in all situations.

Also bugfix : ChannelUpdate -> ChannelCreate


jonas747:
Looks good

PR #621, opened 2020-02-03
fix incorrect regex for matching.

Satty9361:
^ to \A
and $ to \z
this will always work irrespective of multi line mode being set or reset.


jonas747:
epic

PR #625, opened 2020-02-08
Tickets- Include ticket Title in transcripts

Satty9361:
It has been pointed out that finding a transcript using just the ticket number becomes difficult. Adding the ticket title in the filename would resolve this.


jonas747:
Do you know the max length of the filename?

jonas747:
Maybe limit it to say 30 chars, `using common.CutStringShort(s, len)`

Satty9361:
Even stuff like this worked for me:
attachments-2-01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567-0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789.PNG
 so i dont think it should be an issue atleast with addition of channel name which has a limited size by itself. With the attachment name maybe some limit can be imposed.

jonas747:
ah fair enough then

PR #626, opened 2020-02-09
template function: delete Message Reaction(s)

mrbentarikau:
This is very wished function - I dunno about rate-limits here, it is strictly told to run once per CC, that 20 emote deletions later might be a problem/can be changed to let's say 10. Runs fine on lesser test-server, needs further discussion and testing.


jonas747:
Maybe put a limit on the number of reactions, to avoid spam with the api, say 10 max?

mrbentarikau:
roger that, updated it to 10

PR #627, opened 2020-02-09
Idea: Member Status addition to Whois

mrbentarikau:
It's better to be in whois command than separate function for templates.


jonas747:
maybe shorten it down to just "status" instead of "member status"

mrbentarikau:
updated to Status only and tested, purrs like a kitten : )

jonas747:
noice

PR #630, opened 2020-02-12
Random noun function

iansannar:
Set up random nouns using 0% skill and 100% copy and paste. @jonas747, thanks for pointing out the exact files I needed to reference in https://github.com/jonas747/yagpdb/issues/521.

Currently there's only about 100 nouns, but this can be expanded over time.


jonas747:
In the future please make pr's against the "dev" branch. But it seems fine so ill merge it.

noice

PR #633, opened 2020-02-12
templates: comma separator for thousands

mrbentarikau:
Users frequently request this, although this can be done using custom commands. I really don't know why they want this or can't make their own CC, but they, the users cry about this feature constantly.


jonas747:
noice

PR #634, opened 2020-02-13
Automod : Ban Delete Days

Satty9361:
Add support for users to specify number of days of message to delete for ban effect through Automod V2.


jonas747:
Epic

PR #635, opened 2020-02-13
HTML: caret addition to CC card

mrbentarikau:
This is the same \f106 \f107 awesome font up-down caret used in navigation panel when folding the panels. Same here.
caret


jonas747:
epic

PR #636, opened 2020-02-14
Suggestion: replacing CC awesome icon

mrbentarikau:
Although this new icon means closed captioning, YAGPDB uses this abbreviation for custom commands and seems OK to use it here instead of just plain exclamation mark, which usually means danger : )


jonas747:
fair enough

PR #637, opened 2020-02-14
Suggestion: replace cplogs awesome icon

mrbentarikau:
Currently we have two cog icons under Core - representing different things. Also fa-database represents logs in other nav section. So consistency sake let the same database icon represent logs everywhere. 

Side note, I actually \f0c5 fa-copy even better, but let fa-database be current suggestion.


jonas747:
i guess that's more appropriate

PR #638, opened 2020-02-15
Complex message edit

Satty9361:
This function not only lets you edit both embeds and content at once but also allows for completely removing embeds by passing a nil to "embed" key in message edit builder.


Satty9361:
Added some better error handling. The above error is because the GetMarshalNil() does not exist in current version of discordgo here. Should be fixed post update.

jonas747:
assume this runs fine

PR #641, opened 2020-02-21
Fix: whois without arguments does not show status

mrbentarikau:
For some odd reason commands.ContextMS(parsed.Context()) does not register PresenceStatus from dstate, maybe due to member Copy - can't remember how it's called...

So this is one possible way solving it.

### DIFF
--- a/logs/plugin_bot.go
+++ b/logs/plugin_bot.go
@@ -82,7 +82,13 @@ var cmdWhois = &commands.YAGCommand{
 		}
 
 		member := commands.ContextMS(parsed.Context())
-		member, _ = bot.GetMember(member.Guild.ID, member.ID)
+		memberCPY := parsed.GS.MemberCopy(true, member.ID)
+		if memberCPY == nil {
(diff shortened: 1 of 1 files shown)


jonas747:
its because it uses the member object on the message.

Also get it from the state directly instead of calling bot.GetMember, since you can't get presence data from the api at all without state tracking it.

mrbentarikau:
updated that getting member part

logs/plugin_bot.go
```diff
@@ -82,6 +82,13 @@ var cmdWhois = &commands.YAGCommand{
 		}
 
 		member := commands.ContextMS(parsed.Context())
+		memberCPY := parsed.GS.MemberCopy(true, member.ID)
+		if memberCPY == nil {
```
jonas747: Maybe replace it with `if memberCPY != nil .... member = memberCPY` and drop the else clause entirely?

mrbentarikau:
updated the code

PR #642, opened 2020-02-21
Remove mute duration limits from automod v2

Satty9361:
mute limits should be also removed from automod v2 to be in sync with current changes.


jonas747:
Keep min at 1

Satty9361:
0 should be okay as discussed.

PR #644, opened 2020-02-22
Idea to revamp listroles: nomanaged switch and counter

mrbentarikau:
Has -nomana switch to exclude managed/bot roles, also a counter as response on first line.
This all can be done using custom commands, but main-command should also have it.


jonas747:
just name it -nomanaged

mrbentarikau:
Switch renamed

PR #646, opened 2020-02-24
patch timezone responding to own messages bug

Satty9361:
Dont know how this bug popped up but having this additional check is always good.


jonas747:
Hah someone probably did execAdmin with `setz` to set the bot's timezone, amazing

PR #666, opened 2020-03-22
IDEA: function tmplEditCCTriggerType

mrbentarikau:
This function changes CC trigger type - user actually have requested this = Useful for running Intervaled triggers, but I generalized it to all.
This really needs discussion on trigger type matching, and maybe not being so strict. Also checkCallCounter.

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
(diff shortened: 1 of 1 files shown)


jonas747:
This is missing some things, such as sending a pubsub event to clear cache, but overall I'm not sure if we should allow this kinda mutation of the custom commands from custom commands at all, stuff could get weird and complicated real fast, gonna close this for now.

PR #667, opened 2020-03-22
Add "tranny" and it's plural as banned words.

YaBoiJD:
The word tranny and it's plural should be banned words since it is a slur and is offensive towards transexual people.


automod_legacy/swearwords.go
```diff
@@ -60,8 +60,8 @@ var BuiltinSwearWords = map[string]bool{
 	"smegma":      true,
 	"spunk":       true,
 	"tit":         true,
-	"tranny"       true,
-	"trannies"     true,
+	"tranny":       true,
```
C0braD3v, line 63: Please indent the `true` boolean with the rest of them in the line, seems strange to have them one space out.

automod_legacy/swearwords.go
```diff
@@ -60,6 +60,8 @@ var BuiltinSwearWords = map[string]bool{
 	"smegma":      true,
 	"spunk":       true,
 	"tit":         true,
+	"tranny":       true,
+	"trannies":     true,
```
C0braD3v, line 64: Please indent the `true` boolean with the rest of them in the line, seems strange to have them one space out.

max-content:
queer is not an offensive word and should not be on this list.  As a member of the queer community, this is kind of disrespectful.

jo3-l:
I have to admit that I'm not very familiar with the history here, but:
https://en.wikipedia.org/wiki/Queer#Early_pejorative_use
This seems to suggest that queer was once used as a slur and may still be a slur (according to some other sources).

max-content:
> I have to admit that I'm not very familiar with the history here, but:
> https://en.wikipedia.org/wiki/Queer#Early_pejorative_use
> This seems to suggest that queer was once used as a slur and may still be a slur (according to some other sources).

Yes, it was a slur at one point but it was reclaimed as stated in this wikipedia article and sites the Oxford Dictionary as the source. Also, as a queer person I can tell you it's not offensive to use.

fluffyfirefly:
It's correct that the term queer used to be a slur word, But it found its place in the modern-day LGBTQ+ community and is very commonly used to refer to yourself and to identify as a non-hetersexuell person in a non-offensive way. 
There's absolutely no reason this should be listed in the banned swearwords.

jo3-l:
Apologies for the confusion on my part there then. In that case, a PR would probably be the best course of action.

PR #670, opened 2020-03-26
Stop purging messages on ban by default

ifreund:
The current behavior of deleting all messages sent by the banned user in
the past day is not documented anywhere and is undesireable as a default.

This commit sets the default back to 0 days of purged messages, returning
to the behavior of the ban command prior to #549.


Satty9361:
This used to delete messages from past day all along and changing this behavior will be undesirable now as it will break for people who expect messages to delete by default.
#549  allowed users to specify the number of days, before this messages from past day would be deleted and no change was possible.

If you really want to implement such a feature -- maybe a config setting similar to default mute duration can be a plausible solution.

ifreund:
closing this in favor of the better solution in #672

PR #672, opened 2020-03-29
Satty9361 ban ddays config

Satty9361:
This adds a default number of days of messages to delete while banning- which is much requested by users.

This is a better solution than #670  as it doesnt change behavior in existing servers while letting people set their own default per server. 

Should work reasonably fine from testing , however I am not 100% sure about the changes in validation.go for a general purpose, for this - it works as desired.


jonas747:
Looks good 👍

PR #675, opened 2020-03-30
automodV2: userStatus trigger

mrbentarikau:
Users have requested this so maybe this implementation is OK - never done amV2 code before. Uses PresenceUpdate to detect the status either on login or presence change. As much as I've tested it works, maybe needs naming changes.


mrbentarikau:
sad times... can't think of any other ways to do the same thing as well : /

PR #692, opened 2020-04-12
Lockdown commands

Pedro-Pessoa:
Simple lockdown and unlock commands.

With these, you can revoke or grant perms to send msg and add reactions from roles.

Additional -onlymsgs flag to only work with the send message perm and leave the add reaction perm as is.

Also, you can specify a role by name or ID to lock that role down. 
If no role is specified, it will affect the everyone role.

### DIFF
--- a/analytics/analytics.go
+++ b/analytics/analytics.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 
 	"github.com/jonas747/yagpdb/common"
+	"github.com/jonas747/yagpdb/common/config"
 	"github.com/mediocregopher/radix/v3"
 )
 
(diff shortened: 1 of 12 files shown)


GenryMg:
:O that would be a great feature to have!

Pedro-Pessoa:
Function explained after all updates:
Made it a full moderation module.
Users can now toggle the commands on/off at Tools&Utilities -> Moderation
Also, command usage is now logged to modlog.

-d flag added to the lockdown command to make the lockdown happen for a specific amount of time.

SendMessage permission is always affected when using the commands.
Toggleable permissions:
-reaction
-voicespeak
-voiceconnect
-all (reaction + voicespeak + voiceconnect)

Satty9361:
Additional Comments: Had also added a Role-Delete handler to clear all scheduled unlocks for the role. This perhaps is overkill? Not sure so just added for now.

PR #694, opened 2020-04-20
Complex Message Support for sendDM

Satty9361:
users have often requested for complexMessage support for DMs especially to send content + embed style messages. This extends that functionality to DMs.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -31,20 +31,39 @@ func (c *Context) tmplSendDM(s ...interface{}) string {
 	c.GS.RUnlock()
 
 	info := fmt.Sprintf("Custom Command DM From the server **%s**", gName)
-
-	// Send embed
-	if embed, ok := s[0].(*discordgo.MessageEmbed); ok {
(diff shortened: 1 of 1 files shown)


common/templates/context_funcs.go
```diff
@@ -31,20 +31,39 @@ func (c *Context) tmplSendDM(s ...interface{}) string {
...
+			}
+			if (strings.TrimSpace(msgSend.Content) == "") && (msgSend.File == nil) {
+				return ""
+			}
+			msgSend.Content = info + "\n" + msgSend.Content
+		default:
+			msgSend.Content = fmt.Sprintf("%s\n%s", info, fmt.Sprint(s...))
+	}
+	
+	channel, err := common.BotSession.UserChannelCreate(memberID)
```
jonas747, line 62: throw away the error here aswell as to not break existing cc's

Satty9361: I am throwing out the error. I just avoid an extra endpoint call in case of an error and simply return "" early.

jonas747: ah you are, my bad

PR #697, opened 2020-04-24
CC database - new functions and improvements

Satty9361:
Added:
dbDelMultiple - deleting multiple entries 
dbRank - finding rank according to a custom query (very helpful with leader board implementations, I have tested this upto 3 million entries for a given server and it took 3-5s at max even at this scale. dbCount is also equally slow at this scale.)
dbCount improvement.

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -668,6 +668,7 @@ func queryFromArg(query interface{}) (*Query, error) {
 			default:
 			return &q, errors.New("Invalid Key: " + key + " passed to query constructor")
 		}
+	}
 				
 	return &q, nil
 }
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -507,7 +507,7 @@ func tmplDBDelMultiple(ctx *templates.Context) interface{} {
 			return "", templates.ErrTooManyCalls
 		}
 		
-		q, err = queryFromArg(query)
+		q, err := queryFromArg(query)
 		if err!= nil {
 			return "", err
 		}
@@ -550,7 +550,7 @@ func tmplDBRank(ctx *templates.Context) interface{} {
 			return "", templates.ErrTooManyCalls
 		}
 		
-		q, err = queryFromArg(query)
+		q, err := queryFromArg(query)
 		if err!= nil {
 			return "", err
 		}	
@@ -615,7 +615,10 @@ func tmplDBCount(ctx *templates.Context) interface{} {
 				pattern.String = patternStr
 				pattern.Valid = true
 			default:
-				q, err = queryFromArg(query)
+				q, err := queryFromArg(query)
+				if err!= nil {
+					return "", err
+				}
 				userID = q.UserID
 				pattern = q.Pattern
 				
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -636,22 +636,39 @@ func tmplDBCount(ctx *templates.Context) interface{} {
 
 func queryFromArg(query interface{}) (*Query, error) {
 
-	dict, err := templates.StringKeyDictionary(query)
+	querySdict, err := templates.StringKeyDictionary(query)
 	if err != nil {
 		return nil, err
 	}
 	
 	var q Query
-	encoded, err := json.Marshal(dict)
-	if err != nil {
-		return nil, err
-	}
-
-	err = json.Unmarshal(encoded, &q)
-	if err != nil {
-		return nil, err
-	}
-	q.Pattern.String = limitString(q.Pattern.String, 256)
+	for key, val := range querySdict {
+		switch key {
+			case "userID":
+			switch t := val.(type) {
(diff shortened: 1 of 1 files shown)


buthed010203:
I need this in my life already

Crenshaw1312:
Hi, I was wondering what the template for dbRank looks like? I cannot seem to figure it out. Thanks.

customcommands/tmplextensions.go
```diff
@@ -488,6 +497,118 @@ func tmplDBDelById(ctx *templates.Context) interface{} {
...
+	return func(query interface{}, iAmount interface{}, iSkip interface{}) (interface{}, error) {
+		if ctx.IncreaseCheckCallCounterPremium("db_interactions", 10, 50) {
+			return "", templates.ErrTooManyCalls
+		}
+
+		if ctx.IncreaseCheckCallCounterPremium("db_multiple", 2, 10) {
+			return "", templates.ErrTooManyCalls
+		}
+
+		ctx.GS.UserCacheDel(CacheKeyDBLimits)
```
jonas747, line 510: do this at the end after we have deleted the entries, as to avoid somehow refreshing the cache with still old values

customcommands/tmplextensions.go
```diff
@@ -488,6 +497,118 @@ func tmplDBDelById(ctx *templates.Context) interface{} {
...
+			return "", templates.ErrTooManyCalls
+		}
+
+		dict, err := templates.StringKeyDictionary(query)
+		if err != nil {
+			return "", err
+		}
+
+		var q Query
+		encoded, err := json.Marshal(dict)
```
jonas747, line 569: move the query had decoding and whatnot stuff into its own function, like "queryFromDict" or something

jonas747 (changes requested):
looks good overall, a couple notes though

customcommands/tmplextensions.go
Satty9361: done

customcommands/tmplextensions.go
Satty9361: made a queryFromArg function

customcommands/tmplextensions.go
```diff
@@ -488,6 +497,97 @@ func tmplDBDelById(ctx *templates.Context) interface{} {
...
+		CASE WHEN $1 = 'DESC' THEN value_num ELSE 0 END DESC,
+		CASE WHEN $1 = 'ASC'  THEN id ELSE 0 END ASC,
+		CASE WHEN $1 = 'DESC'  THEN id ELSE 0 END DESC
+	) AS position
+FROM templates_user_database WHERE (guild_id = $2) AND ($3::bigint IS NULL OR user_id = $3) AND ($4::text IS NULL OR key LIKE $4) AND (expires_at IS NULL OR expires_at > now())
+) AS w
+WHERE user_id = $5 AND key = $6`
+
+		var rank int64
+		err = common.PQ.QueryRow(rawquery, order, ctx.GS.ID, q.UserID, q.Pattern, userID, key).Scan(&rank)
```
jonas747, line 583: why is there 2 userID's?

customcommands/tmplextensions.go
```diff
@@ -511,23 +611,51 @@ func tmplDBCount(ctx *templates.Context) interface{} {
...
+		const q = `SELECT count(*) FROM templates_user_database WHERE (guild_id = $1) AND ($2::bigint IS NULL OR user_id = $2) AND ($3::text IS NULL OR key LIKE $3) AND (expires_at IS NULL or expires_at > now())`
 
 		var count int64
-		err := common.PQ.QueryRow(q, ctx.GS.ID, userID, key).Scan(&count)
+		err := common.PQ.QueryRow(q, ctx.GS.ID, userID, pattern).Scan(&count)
 		return count, err
 	}
 }
 
+func queryFromArg(query interface{}) (*Query, error) {
```
jonas747, line 637: honestly seems kinda overkill to json serialize and deserialize, maybe just take in a dict and convert that into a quer?

jonas747 (changes requested):

customcommands/tmplextensions.go
Satty9361: so this function is used to find the rank of a particular entry (defined by userID and key) over a particular set of entries or the entire database. When no pattern/userID is mentioned, it finds the global rank. The pattern/userID can be used to further narrow down the set of entries on which the rank is to be found. Now this can be handled using an userID bool in this particular scenario, however for consistency accross functions I have reused the same query structure as being used in the other two functions. Indeed this means if the 2 userIDs here dont match, there is no point in performing a query and I have already handled this special case for optimization.

customcommands/tmplextensions.go
Satty9361: This was being used to create embeds but indeed this is an overkill for a simple structure and I will revise this 👍

Satty9361: Should be Okay now @jonas747

PR #698, opened 2020-04-26
helper function to find the Kind of a datatype

Satty9361:
We have a lot of different datatypes in CCs now and sometimes it becomes difficult to determine what broad "Kind" does one belong to. For example certain methods functions and operators can only work on some Kinds abut not others. This is a simple helper function to help identify that since we can already determine datatype using printf.


jonas747:
Conflicts her aswell

Satty9361:
Fixed here too.

PR #700, opened 2020-04-27
Update to the Clean Command

Pedro-Pessoa:
Added a -to flag to the clean command.

This allows users to clean all msgs untill a certain msg by specifying its ID. 

For example: -clean 100 -to 704152928381960284
If there are only 5 msgs until the msg 704152928381960284, it will only clean 5 msgs.


jonas747:
instead of checking for an expect match, check if `message_id < to_id`

If someone were to delete the to message just as someone sent that clean command, they would accidentally delete a lot more messages otherwise

Pedro-Pessoa:
Made the requested change @jonas747

PR #703, opened 2020-04-27
Lockdown commands

Pedro-Pessoa:
A complete new system for yag. 

Thanks @Satty9361 for all the help in making this possible.

Created a new tab in "_Tools & Utilities -> Moderation_" called **Lockdown**.

Users can enable/disable the lockdown commands, also they can choose to log lockdown actions, choose default role to be locked, choose to create a channel log when the command is used, choose which roles should be able to use the commands.

If no flags are used, the only permission affected is the _SendMessage_ one. 
These are the possible flags: _-reaction, -voicespeak, -voiceconnect, -all_

Also, there is a _-force_ flag, if this flag is used yag will force unlock permission overwrite, in other words, if users never had permission to add reactions and you use -all with -force for example, when  unlocked, users will have the add reactions perms. This also means that if the -force flag is not used, users wont get perms they never had if -all is used.

Lastly, a _-d_ flag to specify the duration of the lockdown, if not specified, its permanent. 

If server is locked, users cant toggle the permissions back, they have to use the unlock command, just like when a user is muted.

Usage example: -lock -all -force -d 1h


Pedro-Pessoa:
A few screenshots of the system working.
lockdown1
lockdown2

daalta:
This sounds very useful!

Right now I can't think of a way to easily prevents users from writing in a channel. Slowmode in ineffective for the first message, and changing role permissions is destructive. Having a lockdown command would be great.

vinceTheProgrammer:
Yo, this looks legit. Someone should approve it.

mattgeowild:
unknown-2
Bug in the server overview under Moderation.

jonas747:
I'm a bit confused as to what this does, mind explaining exactly what it does? It seems like a very complicate way of temporarily removing permissions from roles from what i understand, i feel like this could be a lot simpler.

l-zeuch:
Carl-Bot for example has a similar command, which removes the `send_messages` permission from the default `@everyone` role.
It's to lock down a server in case of a raid or similar. Only problem with Carl-Bot is, they only lock down the default role.

This will, as far as I can tell, also support roles other than the default role.

vinceTheProgrammer:
I for one would put this to use right away. It looks much more versatile than the other lockdown commands currently out there.

Wolveric:
Unless I'm mistaken, one option that doesn't make an appearance, is to set a different default role to lock down...

Seeing as I nearly universally enforce an `Everyone Surrogate` permission structure, being able to instead select this as the default would be almost always preferable, seeing as `@everyone` is effectively left as a dummy role.

Satty9361:
@Wolveric  No it does have a default lockdown role option: 
image

Wolveric:
Danke @Satty9361. 😂 UI is just a little difficult to read when you can see the final result, especially on a phone screen...

3kh0:
is there a option for a server-wide lockdown?

You should also add a option if you do `-lockdown` it lock selected channels in the dashboard. So i can set it that if i run `-lockdown` it will only lock #general and #media and nothing else.

Wolveric:
From the description, applying to select roles, server-wide, is the base behaviour of this system.

Channel-specific commands may have been contributed via another PR.

3kh0:
You should also setup different modes that the user can set.

Say if you use `-lockdown chat` it will only lock 3 selected channels in the dashboard

Pedro-Pessoa:
This needs to be updated. Converting to draft for now.

PR #704, opened 2020-04-28
Satty9361 exec template

Satty9361:
This allows for usage usage of nested templates in form of functions in much more seamless manner. Added fmt.Stringer support to returned struct which defaults to the Response string. 

addReturn allows returning a slice of upto 10 values to the parent template.


jo3-l:
Note: this may be superseded by https://github.com/jonas747/template/pull/6 if that PR is accepted.

PR #707, opened 2020-05-02
Satty9361 role edit

Satty9361:
A basic utility command to edit role color and some other properties. This is now safe because of the guild-scope-cooldown feature which prevents abuse.

ParseColor also exists in simple-embed so i wonder if it would be a good idea to move it to common.

Have kept cooldown at 15s -- wonder if it should be higher?


Satty9361:
This had those strange conflicts from circle.ci so closing this one.

PR #708, opened 2020-05-02
Edit Role Command

Satty9361:
A basic utility command to edit role color and some other properties. This is now safe because of the guild-scope-cooldown feature which prevents abuse.

ParseColor also exists in simple-embed so i wonder if it would be a good idea to move it to common.

Have kept cooldown at 15s -- wonder if it should be higher?

### DIFF
--- a/stdcommands/editrole/editrole.go
+++ b/stdcommands/editrole/editrole.go
@@ -1,0 +1,157 @@
+package editrole
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


jonas747:
I'm going to close this because the only use case is in a custom commands context in which a cc function could be made instead.

Adn im hesistant to add it to custom commands as roles is such a heavy thing on discords side and there is probably some hidden ratelimits, so the potential of abuse for this is through the roof

PR #720, opened 2020-05-21
common: Fix --all for --feeds 

LoganK:


jonas747:
What is the intended change here?

What you've done is make `-feeds` ONLY work if you pass `-all` now

LoganK:
*sigh* Last-minute, late-night changes are always bad, but this video sums it more nicely: https://youtu.be/FjYyIKkRvUU

PR #726, opened 2020-05-24
Satty9361 Add Presence Data to .Member

Satty9361:
Since bot already tracks Presence Data for all online members, it would be useful to have this available under .Member. 

I have used a new type (CtxMember) to extend *discordgo.Member and attach this additional information as it makes more sense to have this info available along with Member struct rather than having a separate function for pulling this data. This also makes it possible to make more additions in the future with a separate type.


jonas747:
Since the API for `get guild member` doesn't return presence data, there's a chance members just wont have presence available in their member state, this would be very flaky as is.

Satty9361:
Yeah I am aware and I have handled that part. In those cases Presence is just shown Unknown or 0 and Presence Game is nil which is Okay. It can be documented that It is not always available.

I also tested on offline members and this part did work as above showing PresenceStatus as  0 or unknown

jonas747:
The reason i haven't merged this and im hesistant is becuase its such unreliable data that i dont see a point really

jonas747:
Closing this for now

PR #742, opened 2020-06-30
Patch 1

JordanTheITGuy:
Added a few items for silliness.

Big fan of the bot. Figured would look through the repos and start with some small little thing to just add. Just thought it would be fun to add some more flavor on some things.


jonas747:
There's some unrelated changes to go.mod and go.sum included

PR #744, opened 2020-07-05
Sort funcs for cslices

Pedro-Pessoa:
I see a lot of users asking for a way to sort slices. You can do it in a CC using range, but why not a built in sort function?
This PR adds a sort function
What it does is create subslices based on the type/kind of the input and after that it sorts ints, strings, floats, time, slices and maps.
Slices and maps are sorted by length 
All other types/kinds are passed unchanged.

User can set flags to the function using sdict
Flags available:

1. Reverse - By default the function sorts by ascending order, this reverses it
2. Subslices - By default the function returns a single slice with all the values sorted, setting this to true will make the function return a set of sublices made based on the input type/kind
3. Emptyslices - By default the function only returns the slices that had an input to them, so, if you sort only strings, the output would be slice of strings. But with this flag the function returns all possible slices, this is helpful for indexing.

I limited this func to 1 and 3 per CC, free and premium users. Dunno if necessary.

image


common/templates/context_funcs.go
```diff
@@ -1335,3 +1336,122 @@ func (c *Context) tmplEditNickname(Nickname string) (string, error) {
...
+		if err != nil {
+			return "", err
+		}
+	}
+
+	var numberSlice, stringSlice, timeSlice, csliceSlice, mapSlice, defaultSlice, outputSlice []interface{}
+
+	for _, v := range slice {
+		switch t := v.(type) {
+		case int, int64, float64:
```
jonas747, line 1374: Lets not mix float64 with int64 as float64 only has integer level precision up to 2^53

common/templates/context_funcs.go
```diff
@@ -1335,3 +1336,122 @@ func (c *Context) tmplEditNickname(Nickname string) (string, error) {
...
 }
+
+func (c *Context) tmplSort(slice []interface{}, sortargs ...interface{}) (interface{}, error) {
+	if c.IncreaseCheckCallCounterPremium("sortfuncs", 1, 3) {
+		return "", ErrTooManyCalls
+	}
+
+	var dict SDict
+	var err error
+	switch len(sortargs) {
```
jonas747, line 1347: add some comments to explain each of these cases

jonas747 (changes requested):

Pedro-Pessoa:
Should be right now

common/templates/context_funcs.go
```diff
@@ -1335,3 +1336,154 @@ func (c *Context) tmplEditNickname(Nickname string) (string, error) {
...
+	case 0:
+		input := make(map[string]interface{}, 3)
+		input["reverse"] = false
+		input["subslices"] = false
+		input["emptyslices"] = false
+
+		dict, err = StringKeyDictionary(input)
+		if err != nil {
+			return "", err
+		}
```
jo3-l, line 1378: This could be simplified to this, right? (not tested)
```suggestion
		dict = SDict{
			"reverse": false,
			"subslices": false,
			"emptyslices": false,
		}
```

Pedro-Pessoa: Yup, you are right

jonas747:
Another problem with this, since you take in []interface for example, you can't use it with []int64 or other slices with types not being interface{},

Pedro-Pessoa:
> Another problem with this, since you take in []interface you can't use it with []int64 or other slices with types not being interface{},

That's true. But I find it not to be a problem because users can not create any other type of slice with custom commands. Unless ofc I'm missing something.

jo3-l:
A `[]string` can be produced from a `split` call, there might be a few other examples I am missing.

Pedro-Pessoa:
Aha, makes sense! Alr, will take that into account

PR #755, opened 2020-07-23
Fix unmute "bug"

Pedro-Pessoa:
Let's say a user is being muted. This user has role A and yag is set up to remove role A when muting.

Before this user is unmuted, the guilds hierarchy is changed and now yag is below role A. 

When unmuting, yag will return this error: `The bot permissions has been incorrectly set up on this server for it to run this command: Missing Permissions`

This will make it impossible to unmute the user as is. Yag would have be put higher in the hierarchy again in other to unmute the user. 

To prevent this, I added a check. It will only append roles to the `newMemberRoles` slice if the given role is below yag.
In other words: the role that was removed and is now above yag won't be added again, but the user will be unmuted nonetheless.


moderation/punishments.go
```diff
@@ -439,14 +440,25 @@ func RemoveMemberMuteRole(config *Config, id int64, currentRoles []int64, mute M
 
 	newMemberRoles := make([]string, 0, len(currentRoles)+len(config.MuteRemoveRoles))
 
+	gs := bot.State.Guild(true, config.GuildID)
+	gs.RLock()
+	defer gs.RUnlock()
```
jonas747, line 445: This would hold the guild lock during a API call, to avoid this create a new function for creating the new member roles where the state is locked and return the new roles from that.

Pedro-Pessoa: Updated the code to fix it!

moderation/punishments.go
```diff
@@ -445,7 +445,10 @@ func RemoveMemberMuteRole(config *Config, id int64, currentRoles []int64, mute M
 	defer gs.RUnlock()
 
 	botState := gs.MemberCopy(true, common.BotUser.ID)
```
DevKurka, line 455: There is a rare case which causes the bot to not be cached in dstate. Experienced that in a huge guild with ~120k members. It will cause panic in every unmute attempt.

That seems to be issue with dstate as the /:guild/botmember internal api fails as well returning "Bot Member not found". Might be worth to use bot.GetMember & double check if the botState isn't nil to prevent issues.

Pedro-Pessoa: You are right! Should be fixed now! Thanks for letting me know :)

moderation/punishments.go
```diff
@@ -439,24 +440,56 @@ func AddMemberMuteRole(config *Config, id int64, currentRoles []int64) (removedR
...
+		}
+
+		return newMemberRoles
+	}
+
+	yagHighest := bot.MemberHighestRole(gs, botState)
 
 	for _, v := range currentRoles {
-		if v != config.IntMuteRole() {
+		if v != config.IntMuteRole() && dutil.IsRoleAbove(yagHighest, gs.Role(false, v)) {
```
jonas747, line 481: If the user already had a role above the bot, wont this attempt to remove it (and therefor fail)?

jonas747:
Revivig this PR

moderation/punishments.go
Pedro-Pessoa: Oh, you are right! Will push a fix soon

Pedro-Pessoa: This should be fixed now

PR #761, opened 2020-07-27
Modify timezone setting procedure to behave more properly

### LINKED ISSUE #731: Undesired result for timezones with overlapping names
Hello,

```
America/Dawson_Creek: MST
America/Dawson: PDT

```

LRitzdorf:
Formerly, if the user specified a zone name that's a subset of another 
one, no action would occur. For example, "setz America/Dawson" would 
inform the user that zones called "America/Dawson" and 
"America/Dawson_Creek" were available, but would not set the desired 
zone. See #731 for details.

Now, if multiple matching zones are available but one is an exact match, 
it will be set, while also informing the user that other matching zones 
are available.

Resolves #731


LRitzdorf:
Ta-da! Should be ready to merge now 🎉

LRitzdorf:
@jonas747 Since there are lots of PRs here, I'm tagging you to draw attention to this one. It should be a reasonably easy fix for an important general-quality item.

LRitzdorf:
@Satty9361 @jonas747 This is the PR that was just referenced in the YAGPDB support server. It fixes the timezone-setting bug.

LRitzdorf:
Closed in favor of #772, which uses the proper feature branch as a base instead of `master`.

PR #762, opened 2020-07-28
Create slowmode command and fix amod2 slowmode effect

Pedro-Pessoa:
In amodv2, when the effect to set slow mode to a channel is triggered, after the duration is gone, instead of making the slow mode go back to it's original state, it's removed completely.
So, for example, if your channel is constantly at a 5s RL slow mode, then a rule is triggered to set it to 1h RL, when that rule expires the channel will have the slow mode removed instead of coming back to 5s RL.
This PR fixes this.

The second part of the PR is a slowmode command `-slowmode`
Syntax is: `-slowmode <rate limit> <optional delay>`

The same command can be used to remove the slowmode by just using a ratelimit of 0.


Pedro-Pessoa:
This needs to be updated. Converting to draft for now.

PR #772, opened 2020-08-28
Modify timezone setting procedure to behave more properly, attempt two

### LINKED ISSUE #731: Undesired result for timezones with overlapping names

LRitzdorf:
This PR serves the same purpose as #761, but uses the proper feature branch as a base.

Formerly, if the user specified a zone name that's a subset of another one, no action would occur. For example, `setz America/Dawson` would inform the user that zones called `America/Dawson` and `America/Dawson_Creek` were available, but would not set the desired zone. See #731 for details.

Now, if multiple matching zones are available but one is an exact match, it will be set, while also informing the user that other matching zones are available.

Resolves #731


LRitzdorf:
@jonas747 @Satty9361 It's been a while, and this simple fix hasn't been merged yet. I'd like to draw your attention here, if possible.
Thanks!

mrbentarikau:
have you even tested the code? Looking at it... it will never solve your America/Dawson issue for example.

LRitzdorf:
@mrbentarikau I have not, actually -- I lack the resources to run a full Discord bot. However, if someone else is able to test my edits, that would be greatly appreciated.

Also, I'm not terribly familiar with Go, so it's entirely possible that my changes might not solve the issue. Can you offer any tips as to what isn't working as intended?

LRitzdorf:
Okay, that *should* work now. As noted previously, I can't confirm this myself, so any further evaluation would be greatly appreciated.

PR #787, opened 2020-09-27
Addition: Stdcommands HowLongToBeat

mrbentarikau:
"hltb" queries the site howlongtobeat.com for information on how long it takes to beat a game : )
Because they do not have an actual API, this is net/html parsing. The site has been stable for at least 3 years in their results formatting, so this query/parse approach is OK to be used.

Parses the first page the site gives with max 20 search results, those are sorted by site's default - by popularity. So if user is not very specific, the first result may differ from what they actually wanted.

The Levenshtein distance usage maybe is too much, but was fun and works as intended.


buthed010203:
👀 hes back!

mrbentarikau:
dunno why circleCI redouts, but code runs fine on v1.26.0 : )

PR #794, opened 2020-10-17
docs(clean): 'clean userId amount' is invalid

Jo3-L:
This PR adds a warning to the description of the `clean` command making it clear (excuse the pun) that the combination of `clean userId mention` is invalid. This has been a source of confusion for many in the support server who have encountered the issue, especially as the auto-generated usage implies this combination is valid, when in fact it is not.
> `Clean <User:Mention/ID> <Num:Whole number>`

It is arguable whether this is a good way to resolve the issue - should the code be changed to handle this specific combination instead? In any case, this needs to be fixed or made clear in some way, hence the PR.


mrbentarikau:
don't rush your PRs to master branch

Jo3-L:
Oops, forgot about that. Will open a new PR.

PR #796, opened 2020-10-19
Update swearwords.go

designeranna1:
Please add the words, "nude" and "horny" to the list.
Also, plz disable the word, "hell" in this list.
Thanks.


Pedro-Pessoa:
Your PR syntax is wrong. 
If you want to make a suggestion you can use the discord server, instead of opening a PR.

Also, it's impossible to make a swear list that would work to every server yag is in. This is where auto mod v2 comes in handy.
You can create your own list using auto mod v2.

designeranna1:
Thanks.

PR #797, opened 2020-10-21
Fix typos and create new command

waterflamev8:
I'm not exactly sure about the path to the JSON file in line 22 so i guess somebody could help me on that. And i'm also aware that there should be error handling for lines 22 and 23 but i'm wasn't exactly sure how it's done in YAGPDB so ig somebody could fill me in on that and i could push another commit.

btw the quotes were from [here](https://github.com/dwyl/quotes/blob/master/quotes.json), is that fine?

### DIFF
--- a/stdcommands/quote/quote.go
+++ b/stdcommands/quote/quote.go
@@ -1,0 +1,36 @@
+package quote
+ 
+import (
+		"encoding/json"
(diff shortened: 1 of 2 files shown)


waterflamev8:
oops idk why "fix typos" is in the title

PR #813, opened 2020-12-10
If multiple matching timezones, inform user of partial matches

### LINKED ISSUE #731: Undesired result for timezones with overlapping names

LRitzdorf:
This PR serves the same purpose as #772, but acts on the proper development branch in the base repo.

Formerly, if the user specified a zone name that's a subset of another one, no action would occur. For example, `setz America/Dawson` would inform the user that zones called `America/Dawson` and `America/Dawson_Creek` were available, but would not set the desired zone. See #731 for details.

Now, if multiple matching zones are available but one is an exact match, it will be set, while also informing the user that other matching zones are available.

Resolves #731


LRitzdorf:
From a conversation with @mrbentarikau, I understand that I have additional work to do here. I will perform that as soon as reasonably possible, using a proper Go IDE.

PR #817, opened 2020-12-16
Update Thanks Regex

NaruDevnote:
This PR updates the thanks regex that should trigger GiveRep. It has been tested and has the same, if not better, functionality than the previous regex.

### DIFF
--- a/reputation/plugin_bot.go
+++ b/reputation/plugin_bot.go
@@ -32,7 +32,7 @@ func (p *Plugin) BotInit() {
 	eventsystem.AddHandlerAsyncLastLegacy(p, handleMessageCreate, eventsystem.EventMessageCreate)
 }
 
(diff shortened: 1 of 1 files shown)


Jo3-L (changes requested):
Unfortunately, this change causes several undesired regressions in functionality. For example, `thanks, @user` is no longer matched on. In addition, phrases like `thanks for your help @user`, `thanks anyways @user`, `thanks! @user`, `@user thanks!` etc. are no longer matched on.

For this reason, I'm not a big fan of the changes here. I feel that the flexibility of the current regex (allowing a lot of variations like `thanks :D @user` - no regex can cover all variations) makes up for its shortcomings.

Jo3-L (changes requested):
This still does not match on postfix `thanks`, i.e. `@user thanks!`. Regex is also overcomplicated but eh

PR #818, opened 2020-12-21
Change to Dicts

Pedro-Pessoa:
Small change to the `Get` method for `Dicts`.

## Dict Get Method
If you run this code: `{{ $x := dict (toInt64 1) "test" }} {{ $x.Get 1 }}`, it will return `nil`.
This code: `{{ $a := dict 1 "a" }} {{ $a.Get 1 }}` will return `a`
And this: `{{ $a := dict 1 "a" }} {{ dbSet 0 "test" $a }} {{ $b := (dbGet 0 "test").Value }} {{ $b.Get 1 }}` will return `nil`.

Imo, this might cause some confusions to users, since the `Get` method would work on the `Dict` before it was set to a DB, but it would return `nil` after fetching the DB.
My solution to this was to check if the key on the `Get` method exists. If it doesn't, it then checks the type of the key. If it's an `int`, it's converted to `int64` and vice-versa. After the conversion it tries to find the key again and returns the value.

This would break a code like this: `{{$a := dict (toInt64 1) "one"}} {{$a.Get 1}}`
Before this PR it would return `nil` and after the PR it would return `one`. 

I find it to be a simple enough solution for the problem at a small cost. What do you think?


jo3-l:
Personally, I'd say this is *mostly* obsoleted by https://github.com/jonas747/yagpdb/commit/4c6412c06f310a3495e9dde6e19780ab4ed42ac2, which removes the need for any deserialization in 90% of cases. WDYT @Pedro-Pessoa?

Pedro-Pessoa:
I thought of closing it... What made be not close it was the fact that https://github.com/jonas747/yagpdb/commit/4c6412c06f310a3495e9dde6e19780ab4ed42ac2 is not retroactive. So this could still be useful. WDYT @jo3-l?

jo3-l:
That's true. However, the new behavior will take effect once the DB value is updated with a value of custom types. Thus, all that a user needs to do to benefit from the new serialization behavior is "refresh" their database entry by retrieving the value, converting to custom types, and saving again.

I do think you should keep this PR open but with a subset of the behavior -- only the changes to `Get()`. That's a problem that the commit above fails to fix (I did not find any good way of doing so, unfortunately).

Pedro-Pessoa:
Mhm, I agree. Alright, will update

PR #824, opened 2021-01-02
Rework prefix

Pedro-Pessoa:
This PR takes the prefix logic from commands and put it on it's own module.
The reason for this is to allow `.ServerPrefix` for CCs context. Without this change, we would get circular dependency.


mrbentarikau:
For consistency sake, maybe .GuildPrefix as well to context.go

PR #825, opened 2021-01-02
Rework ParseDuration

Pedro-Pessoa:
- Removes the `ParseDuration` function from `commands/util` and leave it at `common` only. Since they were duplicated, there was no need to have 2;
- Use `case switching` instead of `else ifs`;
- When there is an error, the function returns `0` instead of `dur`;
- Make `toDuration` check for invalid duration;

Objective: the `ParseDuration` function remains permissive, but it will not return valid if an invalid duration is provided. Also, case switching should make it a little bit faster.

image

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -8,7 +8,6 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"emperror.dev/errors"
 	"github.com/jonas747/dcmd"
 	"github.com/jonas747/discordgo"
 	"github.com/jonas747/yagpdb/bot"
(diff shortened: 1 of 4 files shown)


mrbentarikau:
maybe nil would be better? because logically 0s is still a valid duration : )

Pedro-Pessoa:
> maybe nil or NaN would be better? because logically 0s is still a valid duration : )

You are right. But that would require to change the return of the function from `time.Duration` to `*time.Duration`.
I thought it was not necessary since, though technically 0 is a valid duration, on yag context it is interpreted as invalid or permanent.

What do you think? Should I change the return to properly return `nil` instead of `0`?

mrbentarikau:
I think if it's already quite a big change, let's make the math correct as well.

Pedro-Pessoa:
@mrbentarikau I updated the code as you requested. 
From my tests, it is working fine but there is a backward compatibility issue. Please see my updated initial comment.
Let me know what you think :)

mrbentarikau:
good point you make there, that's why I said big change - I myself don't like to change util+ side at all, 
I think you should revert back to 0, back.comp is more important than my maths.

Pedro-Pessoa:
Reverted it back!

PR #827, opened 2021-01-05
undelete user switch

ShadowNight573:
A user in the YAGPDB support server suggested this. Decided to PR it.

All this does is add a new switch to -undelete which is -u that allows you to only show a specific users deleted messages.
`-undelete -u userID/mention`

Edit: Thank you @Pedro-Pessoa and @jo3-l2 for the feedback.

### DIFF
--- a/stdcommands/undelete/undelete.go
+++ b/stdcommands/undelete/undelete.go
@@ -19,7 +19,7 @@ var Command = &commands.YAGCommand{
 	RequiredArgs: 0,
 	ArgSwitches: []*dcmd.ArgDef{
 		{Switch: "a", Name: "all"},
-		&dcmd.ArgDef{Switch: "u", Name: "user", Type: dcmd.UserID, Default: 0},
+		{Switch: "u", Name: "user", Type: dcmd.UserID, Default: 0},
 	},
 	RunFunc: func(data *dcmd.Data) (interface{}, error) {
 		allUsers := data.Switch("a").Value != nil && data.Switch("a").Value.(bool)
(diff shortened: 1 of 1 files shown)


stdcommands/undelete/undelete.go
```diff
@@ -15,14 +15,16 @@ var Command = &commands.YAGCommand{
 	CmdCategory:  commands.CategoryTool,
 	Name:         "Undelete",
 	Aliases:      []string{"ud"},
-	Description:  "Views your recent deleted messages, or all users deleted messages (with \"-a\" and manage messages perm) in this channel",
+	Description:  "Views the first 10 recent deleted messages. By default only the current users deleted messages will show.\n\nUse `-a` to view all users deleted messages or `-u` to view a specific users deleted messages.\nBoth `-a` and `-u` require \"Manage Messages\" permission.",
```
jo3-l2, line 18:
```suggestion
	Description:  "Views up to 10 recently deleted messages. By default only the current users deleted messages will show.\n\nUse `-a` to view all users deleted messages or `-u` to view a specific users deleted messages.\nBoth `-a` and `-u` require \"Manage Messages\" permission.",
```

stdcommands/undelete/undelete.go
```diff
@@ -15,14 +15,16 @@ var Command = &commands.YAGCommand{
...
 	RequiredArgs: 0,
 	ArgSwitches: []*dcmd.ArgDef{
 		{Switch: "a", Name: "all"},
+		&dcmd.ArgDef{Switch: "u", Name: "user", Type: dcmd.UserID, Default: 0},
```
jo3-l2, line 22: Think it might be better to be consistent here and choose one (`&dcmd.ArgDef{}` in front for both, or none for both).

stdcommands/undelete/undelete.go
```diff
@@ -32,6 +34,17 @@ var Command = &commands.YAGCommand{
 				}
 			}
 		}
+		
+		if targetUser != 0 {
+			if ok, err := bot.AdminOrPermMS(data.CS.ID, data.MS, discordgo.PermissionManageMessages); err != nil || !ok && data.MS.ID != targetUser {
```
jo3-l2, line 39: This line seems a bit complicated, it might read a little better splitting this into a few lines. No big deal though.

jo3-l2 (commented):
Just some stuff I saw reading through, there's nothing major here (just a few stylistic things + description fix).

Is there a reason you didn't choose to have the user provided as an optional positional argument instead of a flag? That seems to fit a little better in my eyes (`-ud @Joe` vs `-ud -u @Joe`). Again, no big deal, but worth considering...

ShadowNight573:
I thought a switch would work better, easier to explain and implement. (in my opinion)

PR #828, opened 2021-01-11
Update swearwords.go

7amakurd:


JordiHakimZiyech:
what is dbx 
and also, please make sure there is a tab between the word and "true"

mrbentarikau:
@7amakurd There's no need to add random acronyms that need Urbandictionary (could also be Dropbox) to explain what it is to hard-coded list. Also, don't PR stuff to master branch.

PR #830, opened 2021-01-13
Add flag to "clean" command to target only messages with attachments.

jtwhale:
Adds a new "-attachments" switch to the clean command, which if provided limits the clean operation to messages with attachments

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -933,16 +933,12 @@ func AdvancedDeleteMessages(channelID int64, filterUser int64, regex string, toI
 			continue
 		}
 
-		// If attachments only filter enabled, add only messages with attachments, else behave as usual
-		if attachmentFilterEnable {
-			if len(msgs[i].Attachments) > 0 {
-				toDelete = append(toDelete, msgs[i].ID)
-			} else {
-				continue
-			}
-		} else {
-                	toDelete = append(toDelete, msgs[i].ID)
+		// Check whether to ignore messages without attachments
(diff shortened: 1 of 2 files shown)


NaruDevnote (commented):
Wouldn't `-a` be more fitting? All the clean flags are kept short, typing out `-clean 10 -attachments` is tedious.

jtwhale:
Absolutely true, I'll shorten it

jtwhale:
All done, gave it a quick test on my self-hosted instance

moderation/commands.go
```diff
@@ -925,7 +933,17 @@ func AdvancedDeleteMessages(channelID int64, filterUser int64, regex string, toI
...
+		if attachmentFilterEnable {
+			if len(msgs[i].Attachments) > 0 {
+				toDelete = append(toDelete, msgs[i].ID)
+			} else {
+				continue
+			}
+		} else {
+                	toDelete = append(toDelete, msgs[i].ID)
+		}
+
```
DevKurka, line 946: Why like that? Could easily stick to the current style by using the power of `continue`.

```go
// Check if attachment to ignore
if attachmentFilterEnable && len(msgs[i].Attachments) == 0 {
	continue
}
```

jtwhale: @DevKurka

> 
> 
> Why like that? Could easily stick to the current style by using the power of `continue`.
> 
> ```go
> // Check if attachment to ignore
> if attachmentFilterEnable && len(msgs[i].Attachments) == 0 {
> 	continue
> }
> ```

You're 100% correct, I've amended it after testing.

PR #835, opened 2021-01-15
Year Change

GoldenXLence:
Not sure if youve realised but 2020 happened and its now 2021.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 jonas747
+Copyright (c) 2021 jonas747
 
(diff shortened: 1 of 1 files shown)


NaruDevnote:
Not a very helpful comment... wouldn't it be better to put 2019-2021 since the license was added in that year?

mrbentarikau:
License file wasn't added in 2019 but updated to 2019 back then - but from 'unknown repository' to 'master' already shows, another troll PR.

JordiHakimZiyech:
bruh people shouldn't be randomly pr'ing without knowing what it does

PR #849, opened 2021-03-11
Typo fixes, active reading, update to a message

iirccm:
Hello, I know this change might seem trivial but it's my first contribution into someone else's project on this platform so please let me know if I had an obvious mistake or a missing part mainly about the general rules (e.g. wrong use of a specific area, wrong terminology, breaking some guidelines etc.) 

Note: I use **Squash and merge** option for these kind of basic stuff to maintain a cleaner committing history so I feel like I can keep track of more important parts. Is there a similar approach like this or I'm doing something wrong?

**Question-1:** I remember I added two labels for a better classification but it seems they are not shown here. Why don't we have the permission to do this as contributors?

20210311_231837_PullRequestQuestion

**Question-2:** About the picture above, how do I see someone else's commit into my own pull request **even before I opened it**? I don't understand the system of it. How is this possible? Also keep in mind that, jonas747's commit seems to be done into a different branch from mine which was added about a month ago. Can someone explain me what's going on about this?

### DIFF
--- a/cmd/yagpdb/templates/cp_selectserver.html
+++ b/cmd/yagpdb/templates/cp_selectserver.html
@@ -12,7 +12,7 @@
             </header>
             <div class="card-body">
                 <p>You can manage this bot through a web control panel and if you need any help or want to contact the owner, you can join the <a href="https://discord.gg/4udtcA5">Support Server</a>.</p>
-                <p>I also worked on adding proper documentation and guides. The documentation might be still lacking and when you have any questions, do not hesitate to join the support server. Also keep in mind that doing a little search on your own first could be more beneficial in the beginning since you can learn in this way better. You can check our <a href="https://docs.yagpdb.xyz/">documentation here</a>.</p>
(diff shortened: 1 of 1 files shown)


NaruDevnote (changes requested):
The documentation is not complete, there are still many things that need to be documented as well as the ongoing German translation which is being written by one person.

iirccm:
> The documentation is not complete, there are still many things that need to be documented as well as the ongoing German translation which is being written by one person.
@NaruDevnote Is there a way to contribute to the documentation? So we can add related the link into the place to inform people if they want to help. Otherwise, I'm not sure how to change the sentence since it still indicates the documentation might not be complete.

NaruDevnote:
> @NaruDevnote Is there a way to contribute to the documentation? So we can add the related link into the place to inform people if they want to help. Otherwise, I'm not sure how to change the sentence since it still indicates the documentation might not be complete.

To contribute to the German translation, speak to **LZeuch#7325** in the YAGPDB Support Server. The official documentation is managed by the server staff only (documentation authors, voluntary support, and admins) so the only way to "contribute" is by suggesting changes / speaking to an author directly.

In the meantime, best to leave it as it is, the documentation will not be complete any time soon as far as I know. :)

iirccm:
I see, I have just reverted the related sentence as you noted but I also added a little message for people if they want to contribute in any ways. By this way, they would know if the project is open to help & suggestions and maybe some of them come and ask how they can help about it.

NaruDevnote (approved):
👍

PR #857, opened 2021-03-31
Remove Use Slash Commands permission from invite links

PlavorSeol:
This effectively reverts a36976d which adds **Use Slash Commands** permission to YAGPDB.xyz's invite links.

I'm highly sure that was a mistake while preparing for Slash Commands. There is no reason bots have **Use Slash Commands** permission. Registering Slash Commands doesn't require that, instead it requires `applications.commands` on authorization that @jonas747 correctly added in c1ee6dd commit. I also don't think bots can use another bot's slash commands via API, and even they could, in most cases it doesn't make sense for bots to use them, and I don't think there is a possibility that features uses them will be added to YAGPDB.xyz.

Since the commit was only added to `master` branch and not to `dev` branch, and this pull request just reverts that, I decided to send it to `master` branch.

### DIFF
--- a/cmd/yagpdb/templates/cp_nav.html
+++ b/cmd/yagpdb/templates/cp_nav.html
@@ -310,7 +310,7 @@
         {{range $index, $element := .ManagedGuilds -}}{{if not $element.Connected -}}
         <li>
             <a
-                href="https://discordapp.com/oauth2/authorize?client_id={{$clientid}}&scope=applications.commands%20bot&permissions=4294311423&guild_id={{$element.ID}}&response_type=code&redirect_uri={{joinStr "" $baseURL "/manage"}}"><i
+                href="https://discordapp.com/oauth2/authorize?client_id={{$clientid}}&scope=applications.commands%20bot&permissions=2146827775&guild_id={{$element.ID}}&response_type=code&redirect_uri={{joinStr "" $baseURL "/manage"}}"><i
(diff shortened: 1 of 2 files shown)


NaruDevnote:
Please read [CONTRIBUTING.md](https://github.com/jonas747/yagpdb/blob/master/CONTRIBUTING.md) before submitting a pull request: PRs are to be made to the dev branch, not the main. Also if I remember correctly, this change was already made recently so there is no need for this pull.

PlavorSeol:
@NaruDevnote

Please also re-read what I wrote on this pull request:
> Since the commit was only added to `master` branch and not to `dev` branch, and this pull request just reverts that, I decided to send it to `master` branch.

The commit that is expected to be reverted by this pull request was **never** added to `dev` branch but `master` branch and is currently live on the YAGPDB.xyz dashboard. In addition, I think this change is small enough.

If that's really an issue, even for this case, I can send another pull request, but in that case [some unrelated commits](https://github.com/jonas747/yagpdb/compare/dev...master) that are not on `dev` branch but `master` branch should also be added to pull request.

jo3-l:
Not quite sure what the point of this PR -- while adding that permission doesn't really add much, reverting it does nothing either. If I'm wrong in any of those assumptions, please feel free to correct, but as it stands this seems like a somewhat unnecessary revert, at least to me.

jonas747:
He's fully correct that i did this as a mistake initially

PR #861, opened 2021-04-11
Update swearwords.go

sinamostaghim:
rdgfd


JordiHakimZiyech:
PR's should be in the dev branch, not master.
Also, pr'ing non-english words is close to useless. Create your own swear word list for that

PR #865, opened 2021-04-15
templates: dbIncr should not increment expired value

jo3-l:
Currently, the SQL executed for `dbIncr` does not check whether the entry has expired, causing a bug where it increments the value for an entry which has already expired.

See the following code:
```
{{dbSetExpire 0 "test" 5 10}}
{{sendMessage nil (print (dbIncr 0 "test" 1) " " (dbGet 0 "test").Value)}}
{{sleep 15}}
{{sendMessage nil (print (dbGet 0 "test").Value " " (dbIncr 0 "test" 1))}}
```
This currently results in this:
image

As seen above, `dbIncr` increments the value of the expired entry from 6 -> 7, instead of treating it as a nonexistent entry and setting the value to 1.

This patch modifies the behavior of `dbIncr` to do this instead, treating the expired entry as nonexistent:
image

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -398,25 +398,29 @@ func tmplDBIncr(ctx *templates.Context) interface{} {
 		keyStr := limitString(templates.ToString(key), 256)
 
 		const q = `INSERT INTO templates_user_database (created_at, updated_at, guild_id, user_id, key, value_raw, value_num) 
-VALUES ($1, $1, $2, $3, $4, $5, $6)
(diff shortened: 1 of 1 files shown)


jo3-l:
Not sure if this should set `created_at` as well, will convert to draft for the moment to ensure that this isn't merged before this is resolved.

jonas747:
Instead of passing in time.Now() you can just use the postgres `now()` function which returns the time when the transaction started.
Also yeah also set created

jo3-l:
Updated to use `now()` and set `created_at` as well. Works fine as far as I can tell on my selfhost.

PR #872, opened 2021-04-24
Fix bad new lines

JantsoP:
This will fix issues which cause panic: template: logs_control_panel_html:20: unclosed action error when starting bot.

https://idrank.poisoned.coffee/u/m217nP0m.png


JantsoP:
Issue was first seen on commit 331869033ed8ee49955760f07838d3d4f1a5d100

Pedro-Pessoa:
This should not be an issue on go versions `1.16` and above

jonas747:
Yeah instead of merging this i will be bumping the minimum supported go version up to 1.16

PR #873, opened 2021-04-24
Limit reSplit and allow optional int input for reFind and reFindAllSubmatches

Pedro-Pessoa:
This PRs limits the `reSplit` function to 500 matches. What do you think about this limit, Jonas?
This is the number we came to a conclusion on the support server.

It also makes the `int` argument optional without breaking back comp and at the same time keeping a consistency with the other regex functions. 

This PR also makes `reFind` and `reFindAllSubmatches` accept an `int` argument, which is optional and limited to the current limit.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -1191,7 +1191,7 @@ func (c *Context) compileRegex(r string) (*regexp.Regexp, error) {
 	return compiled, nil
 }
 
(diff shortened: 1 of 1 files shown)


jonas747:
this has conflicts now, theres also a bunch of changes not relevant to this included somehow, and you dont seem to have run gofmt on it either.

Pedro-Pessoa:
Everything should be fine now

PR #883, opened 2021-04-29
templates: update role funcs

phenpessoa:
This PR creates these functions: `mentionRole`, `hasRole`, `targetHasRole`, `giveRole`, `addRole`, `takeRole`, `removeRole`.
To maintain back comp, all the "original" functions call these new functions.

I find it odd that we have to use `hasRoleName` and `hasRoleID` for example. Why not a `hasRole` function?
This is the case for every role function. 

Also, it was not possible to use the role functions with role mentions. So the new `FindRole` function takes all types of role input: Mentions, IDs and Names.

To find the role by name, I used `strings.EqualFold`. I don't know if I should maintain it or make it case sensitive. What do you think?

This PR also already adds #757 and #752. So, the function `getRole` is implemented amongst `optional delay support for addrole/giverole templates similar to removerole/takerole templates`.

The difference between this PR, #752 and the current implementation is that it takes multiple duration arguments and not only numbers. I also made this change to the already existing functions with an optional delay.
This should be back compatible because of the way the new func `validateDurationDelay` parses the duration.

Lastly, I moved all the role functions to the end of the file just for consistency/organization.


ashishjh-bst:
@phenpessoa  I have fixed the conflicts on this, whenever you have some time a review would be great!

PR #884, opened 2021-04-30
Trim whitespaces from prefix

l-zeuch:
[User suggestion](https://discord.com/channels/166207328570441728/356486960417734666/822955179413995580) on the server, as Discord trims leading whitespaces why not trim whitespaces from the prefix as well.
This will eliminate such a weird prefix as an issue of the bot being unresponsive.

This change trims both leading and trailing whitespaces, however it is easy to change that to only leading whitespaces.
Up to you, but I think both is good.

If I forgot to edit something, please tell me, as far as I could see that is the only function responsible for editing the prefix. Still quite new to Go and yag's code :)

### DIFF
--- a/commands/plugin_web.go
+++ b/commands/plugin_web.go
@@ -185,7 +185,7 @@ func HandleCommands(w http.ResponseWriter, r *http.Request) (web.TemplateData, e
 func HandlePostCommands(w http.ResponseWriter, r *http.Request) (web.TemplateData, error) {
 	ctx := r.Context()
 	activeGuild, templateData := web.GetBaseCPContextData(ctx)
-	newPrefix := strings.TrimLeft(r.FormValue("Prefix"))
+	newPrefix := strings.TrimLeftFunc(r.FormValue("Prefix"), unicode.IsSpace)
 	if len(newPrefix) < 1 || len(newPrefix) > 100 {
 		return templateData, web.NewPublicError("Prefix is smaller than 1 or larger than 100 characters")
 	}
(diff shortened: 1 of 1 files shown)


jonas747:
i think we should do leading whitespaces only as the trailing whitespae is used to be able to do the dollowing: `!yag help` as opposed to ``!yaghelp`

l-zeuch:
Not a problem, I'll push an update!

commands/plugin_web.go
```diff
@@ -184,7 +185,7 @@ func HandleCommands(w http.ResponseWriter, r *http.Request) (web.TemplateData, e
 func HandlePostCommands(w http.ResponseWriter, r *http.Request) (web.TemplateData, error) {
 	ctx := r.Context()
 	activeGuild, templateData := web.GetBaseCPContextData(ctx)
-	newPrefix := r.FormValue("Prefix")
+	newPrefix := strings.TrimLeft(r.FormValue("Prefix"))
```
jo3-l, line 188: `strings.TrimLeft` does not work in the same way as `strings.TrimSpace`.
Just to give you an idea of what to do (you need to import `unicode` if it isn't already):
```suggestion
	newPrefix := strings.TrimLeftFunc(r.FormValue("Prefix"), unicode.IsSpace)
```

l-zeuch: Just saw it as well, thanks though :')

jo3-l: `unicode` needs to be imported as stated in the initial review -- would suggest that you do this from an editor instead of via GitHub UI since `gofmt` is probably going to have some ideas as to where to put it. I know you've probably seen this already but just wanted to make it known here in case Jonas comes back and merges before you've edited :-)

jonas747:
👍

PR #886, opened 2021-05-02
reminders: add -a flag for delreminder to clear all

jo3-l:
As title says.
Works fine as far as I can tell on my selfhost.


jonas747:
👍

PR #887, opened 2021-05-02
moderation: make clear cmd error if missing perms

jo3-l:
Make the `clear` command check for the Manage Messages before executing the rest of the command.
From user suggestion on the server:
> Make yag display an error message when it doesn't have permission to purge messages, rather than saying it deleted a few messages when it actually didn't.

Works fine on selfhost, as far as I can see.


jonas747:
since a lot of commands require the bot to have elevated perms like this it really shouldve been in a field fo YagCommand, ill merge in this though

PR #888, opened 2021-05-02
logs: support deleting all message logs at once

jo3-l:
As title states.
Works fine as far as I can tell on selfhost.
New view of logging section
Confirmation modal
Control panel log view


jonas747:
👍

PR #889, opened 2021-05-02
moderation: -im flag to invert regex match for clear cmd

jo3-l:
This allows users to only clear messages that do *not* match the given regex.
`-im` stands for **i**nvert regex **m**atch, ideally it've just been `-i` but that was taken already. Happy to bikeshed over naming.


jonas747:
👍

PR #890, opened 2021-05-02
cc: improve error reporting

jo3-l:
This vastly improves error reporting in custom commands when possible.

**Changes**
* Instead of "ccid:line:column", use a more human readable representation
* Display the lines around where the error occurred in the error message

**Old**
Image of current error

**New**
Image of new error


jonas747:
very nice 👍

PR #891, opened 2021-05-03
moderation: check bot perms before performing kick/ban

jo3-l:
Similar to https://github.com/jonas747/yagpdb/pull/887, but for kick and ban (totally forgot about this, otherwise I would've included it in the original PR 😄). This is a slightly more annoying/confusing issue for users - kick and ban would still send a moderation DM if the bot did not have sufficient permissions to moderate the user.
Works fine on selfhost, as far as I can see.


jonas747:
this has conflicts now

jo3-l:
Should be fixed now @jonas747

jonas747:
Not needed after https://github.com/jonas747/yagpdb/commit/1772fe30bcd4a3849781e3ee92cb54b75f78e433

jo3-l:
Actually, not sure if https://github.com/jonas747/yagpdb/commit/1772fe30bcd4a3849781e3ee92cb54b75f78e433 fully resolves this -- as far as I can see the issue will still occur if the target member is higher than the bot. WDYT @jonas747? Happy to modify this PR to add that fix / keep it closed if you don't think it's worth the additional complexity 😄

jonas747:
I'll reopen, but we only need that one check then

jo3-l:
Should be good now, just tested.

ashishjh-bst:
@jo3-l  is this still needed? There is a minor conflict here

jo3-l:
I'm fairly sure the issue is still present, but don't have time to resolve the merge conflict at present; feel free to apply the patch directly. Thanks.

PR #893, opened 2021-05-04
templates: add support for switch args w/ parseArgs

jo3-l:
This PR adds support for switch args w/ parseArgs.

**Changes**
* A new function, `cswitch`, is added to enable defining switch arguments. Open to bikeshedding about the name, it's quite terrible but I was unable to think of something better. At least it's consistent with `carg`?
* `parseArgs` is slightly modified to support switch args during parsing.
* A new method, `Switch`, is added to the `ParsedArgs` struct to enable retrieving parsed switch args.

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"fmt"
+	"strings"
 	"time"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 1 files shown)


MatiasMFM2001:
> A new function, `cswitch`, is added to enable defining switch arguments. Open to bikeshedding about the name, it's quite terrible but I was unable to think of something better. At least it's consistent with `carg`?

I propose:
1) `switch`
2) `modifier`
3) `cflag` -> I like it
4) `option`
5) `copt`

PR #894, opened 2021-05-04
templates: add permission checking funcs + bitwise ops

jo3-l:
Often users want to check permissions in CC (i.e. they only want admins to run this command, but don't want to specify a role). Currently the main way of doing this is using `-viewperms` and then using string manipulation to check the permissions. This is 1) possibly unsafe, if the user does not strip the username from the resulting text, 2) easy to get wrong. Furthermore it's basically a waste of an `exec` call just to get user permissions.

This PR adds the following:
* Constants for Discord permission bits on the dot.
* `bitwise*` funcs to manipulate permissions.
* `hasPermissions/targetHasPermissions/getTargetPermissionsInChannel` funcs to get Discord permissions of users. (should `Permissions` be singular or plural here? dunno...)

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -76,7 +76,7 @@ var (
 		// bitwise ops
 		"bitwiseAnd":        tmplBitwiseAnd,
 		"bitwiseOr":         tmplBitwiseOr,
-		"bitwiseXOR":        tmplBitwiseXOR,
+		"bitwiseXor":        tmplBitwiseXor,
 		"bitwiseNot":        tmplBitwiseNot,
 		"bitwiseAndNot":     tmplBitwiseAndNot,
 		"bitwiseLeftShift":  tmplBitwiseLeftShift,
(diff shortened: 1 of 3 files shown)


mrbentarikau:
```
...
c.Data["Permissions"] = map[string]int64{
    "AttachFiles":        discordgo.PermissionAttachFiles,
    "EmbedLinks":         discordgo.PermissionEmbedLinks,
... 
}
```
would still be better than adding twenty seven Permission words as extra weight to the context data

jo3-l:
> ```
> ...
> c.Data["Permissions"] = map[string]int64{
>     "AttachFiles":        discordgo.PermissionAttachFiles,
>     "EmbedLinks":         discordgo.PermissionEmbedLinks,
> ... 
> }
> ```
> 
> would still be better than adding twenty seven Permission words as extra weight to the context data

Done in https://github.com/botlabs-gg/yagpdb/pull/894/commits/ec3872d67cfc372209bf44d3916e20fe1d6baf9d.

ashishjh-bst:
rebase is needed

jo3-l:
Rebased.

PR #896, opened 2021-05-05
cc: add ability to run interval commands from dash

jo3-l:
This adds the ability to run interval CCs immediately from the dashboard, as title states.
Mostly useful for testing (no need to wait for it to trigger / put it into another CC).
To prevent spam, there is a 5s cooldown -- let me know if you want this changed.

Image of main CC listing
Image of edit CC view
Second one looks particularly ugly but I couldn't find a better place to put it, suggestions are appreciated.

### DIFF
--- a/customcommands/assets/customcommands-editcmd.html
+++ b/customcommands/assets/customcommands-editcmd.html
@@ -361,6 +361,12 @@
                                     formaction="/manage/{{$guild}}/customcommands/commands/{{.CC.LocalID}}/delete">Delete</button>
                             </div>
                         </div>
+                        <div class="row mt-4" id="interval-cc-run-now">
+                            <div class="col">
+                                <button type="submit" class="btn btn-secondary btn-block" title="This will trigger this custom command immediately"
+                                    formaction="/manage/{{$guild}}/customcommands/commands/{{.CC.LocalID}}/run_now">Run now</button>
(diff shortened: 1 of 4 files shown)


MatiasMFM2001:
Here, button image would be 🔄 emoji. What do you think?
imagen
imagen

For your first screenshot, I think user may hit `Run now` and/or `Delete` buttons by accident, trying to click `Edit`. I suggest this:
imagen

jo3-l:
1) 🔁 emoji feels too subtle to me to be honest.
2) That's a good idea, will switch around the order in a moment...

jonas747:
This dosen't work when the webserver is not running on the same process as the bot itself, you should either make a pubsub that triggers on the bot side, or just set the next run time to immediately and update the scheduled event for it

jonas747 (changes requested):
This dosen't work when the webserver is not running on the same process as the bot itself, you should either make a pubsub that triggers on the bot side, or just set the next run time to immediately and update the scheduled event for it

jo3-l:
Switched to a pubsub event in https://github.com/jonas747/yagpdb/pull/896/commits/703d0fa9bacacd3dde6453304b067930262c530a, still works fine (as far as I can tell) but I haven't tested with the webserver/bot on different processes.

PR #897, opened 2021-05-06
Cache invites for automod

Pedro-Pessoa:
Seeing that YAG is oftenly being rate limited on the invites endpoint, caching the invites could perhaps help it.

The only issue here is that if the invite is no longer valid, amod will still flag it the invite is in the cache. 
But I think it's a fair enough drawback since the invite would be for another server anyway.


automod_legacy/rules.go
```diff
@@ -215,6 +227,34 @@ OUTER:
...
+				for _, cachedInvite := range cache {
+					if cachedInvite.Invite == id {
+						// If the invite is present on our cache, we return true even if it is not valid anymore
+						// This is to prevent making API Calls about invites which have a restrict rate limit
+						// Because otherwise we would most likely hit those limits
+						InvitesCache.RUnlock()
+						return true
+					}
+				}
+			}
```
jo3-l, line 254:
```suggestion
			for _, cachedInvite := range cache {
				if cachedInvite.Invite == id {
					InvitesCache.RUnlock()
					if guild == guildID {
						// Ignore invites to this server
						continue OUTER
					}
					// If the invite is present on our cache, we return true even if it is not valid anymore
					// This is to prevent making API Calls about invites which have a restrict rate limit
					// Because otherwise we would most likely hit those limits
					return true
				}
			}
```
This looks to be equivalent? Bit easier to understand for me as well but dunno

Pedro-Pessoa: Your are right!

automod_legacy/rules.go
```diff
@@ -238,6 +298,56 @@ OUTER:
...
+		logger.Info("Starting invites cache GC")
+
+		start := time.Now()
+		var counter int
+
+		InvitesCache.Lock()
+
+		for guild, cache := range InvitesCache.CacheMap {
+			for i, cached := range cache {
+				if time.Since(cached.CreatedAt) > (60 * time.Minute) {
```
jo3-l, line 319: Not sure if you want to use `InvitesCacheDuration` here instead of hardcoded `60`

Pedro-Pessoa: Yup

automod_legacy/rules.go
```diff
@@ -184,6 +185,17 @@ func (i *InviteRule) Check(evt *discordgo.Message, cs *dstate.ChannelState) (del
 	return
 }
 
+type CachedInvite struct {
+	CreatedAt int64
```
jonas747, line 189: use time.Time as internally it uses a monotonic clock for clock for comparisons which is gonna be more reliable

jonas747 (changes requested):

automod_legacy/rules.go
Pedro-Pessoa: Alright! Will do

PR #899, opened 2021-05-07
command_overrides: add warning when disabling cmds globally

jo3-l:
This adds a confirmation modal when disabling all commands globally via command overrides.
This is a very common mistake that newbies make - for whatever reason - causing YAGPDB to become "unresponsive" on their server. This warning box will hopefully stop most of these mistakes from being made while still allowing advanced users to disable the setting easily.

Image of warning modal


jonas747:
This has conflicts now, also has this been tested on mobile?

jo3-l:
Good point, will check on mobile in an hour or so + fix the conflicts..

jo3-l:
Looks fine on mobile (tested on an iPad and Samsung phone).

PR #903, opened 2021-05-08
cp_alerts: add warning alert when shard having issues

jo3-l:
**Summary**
Much bot functionality does not work as expected when the bot shard is having issues. However during downtime it's not uncommon for the webserver to be working meaning that users can change things on the control panel (i.e. add a custom command) but see nothing happen when they use it on their server. Of course it's not hard to check on Discord but this is slightly nicer and not a huge code change either.

Adapted from user suggestion on the server:
> Have a status :red_circle::green_circle: on the home page of your server control panel if the bot is online or offline in your server. I didn't know if it was online for testing because I have the bot on the bottom of the list. This make it faster to see if the bot it online or offline in your server without going to your server to see for yourself.

**Status**
I've tested this to the best of my ability on selfhost (since it normally gets 0 events anyways) but there are some parts that are **untested** due to not being able to find a way to kill the bot while keeping web up.

Image of warning


frontend/templates/cp_main.html
```diff
@@ -156,6 +156,58 @@ <h2 class="card-title">
...
+                    removeAlert = false;
+                } else {
+                    if (data.host_statuses) {
+                        loop: for (const host of data.host_statuses) {
+                            if (!host.Nodes) continue;
+                            for (const node of host.Nodes) {
+                                if (!node.shards) continue;
+                                for (const shard of node.shards) {
+                                    if (shard.shard_id === shardID) {
+                                        if (shard.events_per_second < 1) {
```
ashishjh-bst, line 177: For self hosts events_per_second will be less than 1, so would be better if it only checks the offline shards list.

mrbentarikau: this thing is still not fixed for self hosts

PR #906, opened 2021-05-08
slashcommands: add missing strings.ToLower()

jo3-l:
Looks like this was missed in https://github.com/jonas747/yagpdb/commit/2e09ee17ecf2dac1527acb8b00c1451f1ae62b6e.


jonas747:
not needed because of line 184

PR #907, opened 2021-05-08
Logs: Channel flag for log command

l-zeuch:
User suggestion, a channel flag for the log command.

Works fine on selfhost, as far as I can tell. Nothing much to explain on top of that 😄

### DIFF
--- a/logs/plugin_bot.go
+++ b/logs/plugin_bot.go
@@ -51,12 +51,29 @@ var cmdLogs = &commands.YAGCommand{
 	Arguments: []*dcmd.ArgDef{
 		{Name: "Count", Default: 100, Type: &dcmd.IntArg{Min: 2, Max: 250}},
 	},
+	ArgSwitches: []*dcmd.ArgDef{
+		{Name: "channel", Help: "Optional channel to log instead", Type: dcmd.Channel},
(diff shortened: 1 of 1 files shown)


jonas747:
Hm we should have a check for whether they have access to the channel or not

jonas747 (changes requested):
Hm we should have a check for whether they have access to the channel or not

l-zeuch:
You're right, we should check that. Will push an update in a moment!

jonas747 (changes requested):
We should also check the view channel and message history perms

l-zeuch:
> We should also check the view channel and message history perms
Isn't `view_channel` implied through `read_messages`? 
Can't seem to find the constant for `view_channel` at https://pkg.go.dev/github.com/jonas747/discordgo@v1.5.6?utm_source=gopls#pkg-constants ...

I'll add the message history permission, still not quite sure about the view channel perm.

jonas747:
You're right, i need to add it, i'm not 100% if it is implied or not but better be on the safe side of things and not make assumptions when working with discord.

l-zeuch:
👍 lmk when you added it

PR #909, opened 2021-05-09
rsvp: when evt create ran via interactions, send followup

jo3-l:
This fixes a visual bug where YAGPDB would be seen as infinitely "thinking" and eventually time out when the `events create` command is ran as an interaction. The issue is that on a successful invocation of the command nothing is returned from the run func (`SetupSession` handles it all, and it just sends normal messages). This PR changes `SetupSession` to send a follow-up message for the first setup message instead of a normal message.

Image of the visual bug

Works fine on selfhost (tried both running normally with prefix and with slash commands).


rsvp/setup.go
```diff
@@ -355,6 +367,29 @@ func (s *SetupSession) sendMessage(msgf string, args ...interface{}) {
 	}
 }
 
+func (s *SetupSession) sendInitialMessage(data *dcmd.Data, msgf string, args ...interface{}) {
+	switch data.TriggerType {
+	case dcmd.TriggerTypeSlashCommands:
+		content := "[RSVP Event Setup]: " + fmt.Sprintf(msgf, args...)
```
jonas747, line 373: you can use `data.SendFollowupMessage(reply interface{}, allowedMentions discordgo.AllowedMentions)` to handle things a bit easier

jonas747 (changes requested):

PR #910, opened 2021-05-09
Embed report message

l-zeuch:
User suggestion on the server, makes it consistent with the other messages and removes the annoying pings.


l-zeuch:
Rookie mistake, forgot to add error handling. Will push an update in a short moment.

PR #912, opened 2021-05-09
moderation: support anonymous reports

jo3-l:
This PR adds support for *anonymous reports*, where the reporter's name is hidden in the mod log message (but is still retrievable by staff). It's mostly useful for servers with public mod-logs where the accused user may start an unwanted argument with those who reported them.


jo3-l:
Gonna close this one since Jonas has stated he's against public report channels and this has very little value for private report channels.

PR #916, opened 2021-05-11
stdcommands: add an english dictionary lookup command

jo3-l:
> **Warning:** Though I believe a dictionary is a useful tool to have as a built-in command, there are undeniably some occasional slight issues with the output of this command due to the API used, so feel free to decline this if you feel it isn't worth it.

This adds an English dictionary command using the [Owlbot API](https://owlbot.info/). I've been testing it for a few days now and while it has some quirks it produces good output in a majority of cases. However, sometimes it does produce improper output -- for example:
Bad output example 1
Bad output example 2

I will say though that out of the dozens of words I have tried with it that's the very worst that has happened, typically it produces reasonable definitions, as an example:
Definition of owl
Definition of maligned

It does require an API key to use, but there are no limitations or bullshit required to sign up (just an email address, which they email the key to immediately). It's also completely free, and is the only dictionary API of this type I've found.

### DIFF
--- a/cmd/yagpdb/sampleenvfile
+++ b/cmd/yagpdb/sampleenvfile
@@ -71,6 +71,9 @@ YAGPDB_DISABLE_REQUEST_LOGGING=""
 YAGPDB_AYLIENAPPID="aylien app id here"
 YAGPDB_AYLIENAPPKEY="aylien app key here"
 
+# Owlbot
+YAGPDB_OWLBOT_TOKEN=
+
 # Reddit
 # You'll need to get a permanent refresh token using a script or something, i should probably make a easy way to do it but eh...
 YAGPDB_REDDIT_CLIENTID="Reddit application clientid here"
--- a/stdcommands/owldictionary/owldictionary.go
+++ b/stdcommands/owldictionary/owldictionary.go
@@ -1,0 +1,206 @@
+package owldictionary
+
+import (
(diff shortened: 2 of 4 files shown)


jo3-l:
mh yeah I agree, original thinking was: there's no precedent for maybe-added commands in `stdcommands` (the command is not registered if there's no API key) but a plugin seemed like the perfect fit. However a plugin with a single command is indeed somewhat ridiculous and I'll modify this to be in `stdcommands`.

mrbentarikau:
I mean it can still use admin/config and belong to stdcommands, no extra input necessary, just move the folders and rename paths : )

jo3-l:
dictionary is now a stdcommand as of https://github.com/jonas747/yagpdb/pull/916/commits/4de5fda404a93bab585b3ceefb2e449823955192

Engineer152:
Joe, should there be a check for swear words? I tested this on my self-host and it deleted the trigger if there was a swear word, however it did not delete the response embed.

mrbentarikau:
just a comment, it needs go 1.6+ because of io.ReadAll.... many systems maybe are not so upgraded yet and it will cause confusion.

ashishjh-bst:
@jo3-l  rebase is needed, and imports in new files have to be changed to that of monorepo format.

jo3-l:
Rebased.

PR #919, opened 2021-05-12
templates: Fix DM template bugs

l-zeuch:
This pull requests adds several needed fixes to the DM templates in YAGPDB's custom command scripting.

We now perform each time a check if the targeted member object is present. If that fails to exist, the direct message will not be sent.
This is to remedy the "abuse" of `execCC` to still be able to send a leave DM, and should be quite future-proof.

Moreover, we add the info text we can see in `sendDM` to the other DM template `sendTemplateDM` - I was not able to find out how to send everything in one message, so it gets delivered as a second, additional message. Let me know what you think about that.

Of course, this might be a small-ish breaking change (again), but I think it is more important to block leave DMs entirely.
Everything works, as far as I can tell, fine on a selfhosted instance.

Cheers!

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -139,6 +139,8 @@ type Context struct {
 
 	CurrentFrame *contextFrame
 
+	IsExecedByLeaveMessage bool
+
 	contextFuncsAdded bool
 }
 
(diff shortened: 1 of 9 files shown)


l-zeuch:
Closing this, fucked up the history. Will reopen a new one.

PR #921, opened 2021-05-12
totally disable sending DM on leave message

l-zeuch:
Like #919, but with a clean history.
This pull requests adds several needed fixes to the DM templates in YAGPDB's custom command scripting.

We simply check if the template is being executed by the leave message, if that is true, it will fail to send the message.
This is to remedy the "abuse" of `execCC` to still be able to send a leave DM, and should be quite future-proof.

Moreover, we add the info text we can see in `sendDM` to the other DM template `sendTemplateDM` - Luckily @Pedro-Pessoa helped me with that, so a big kudos to him.

Of course, this might be a small-ish breaking change (again), but I think it is more important to block leave DMs entirely.
Everything works, as far as I can tell, fine on a selfhosted instance.

Cheers!

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -22,7 +22,7 @@ var ErrTooManyCalls = errors.New("too many calls to this function")
 var ErrTooManyAPICalls = errors.New("too many potential discord api calls function")
 
 func (c *Context) tmplSendDM(s ...interface{}) string {
-	if len(s) < 1 || c.IncreaseCheckCallCounter("send_dm", 1) || c.MS == nil {
+	if len(s) < 1 || c.IncreaseCheckCallCounter("send_dm", 1) || c.IncreaseCheckGenericAPICall() || c.MS == nil || c.IsExecedByLeaveMessage {
 		return ""
 	}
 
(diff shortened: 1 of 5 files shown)


common/templates/context_funcs.go
```diff
@@ -180,6 +180,10 @@ func (c *Context) ChannelArgNoDM(v interface{}) int64 {
 }
 
 func (c *Context) tmplSendTemplateDM(name string, data ...interface{}) (interface{}, error) {
+	if c.IsExecedByLeaveMessage {
+		return "", errors.New("Can't use sendDM on leave msg")
```
jo3-l, line 184:
```suggestion
		return "", errors.New("Can't use sendTemplateDM on leave msg")
```

l-zeuch: 😮 guess I shouldn't be too surprised

l-zeuch:
Has conflicts now, I'll resolve them in a moment

l-zeuch:
Not sure why it redouts now, investigating.

BlackWolfWoof:
i'll be really happy to see this being implemented... 😈

PR #922, opened 2021-05-12
moderation: make clean consistently ignore trigger msg

jo3-l:
Make `clean` ignore the triggering message, so `-cl 5` will remove 5 messages -- as opposed to 4 messages plus the trigger -- from the perspective of the user.

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -499,8 +499,20 @@ var ModerationCommands = []*commands.YAGCommand{
 			userFilter := parsed.Args[1].Int64()
 
 			num := parsed.Args[0].Int()
-			if (userFilter == 0 || userFilter == parsed.Author.ID) && parsed.Source != 0 {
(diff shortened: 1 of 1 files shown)


jo3-l:
Addressed in https://github.com/jonas747/yagpdb/pull/922/commits/a4b3e4dad6562fb4b604e3fb90b8bb46b456a494, trigger is now not included in the final count.
Also looked like this actually still deletes even if triggered by `exec/execAdmin` so I went ahead and added a check for the `ExecutedByCC` context key.

jonas747:
What i meant was take in a "authorID" param for the delete function, and just silently include that

Increasing the count by 1 is kind of a hack and could potentially include a extra message if the message was deleted before the command was handled

jo3-l:
Ah OK. Just to clarify, would you like this to manually delete the trigger message in `AdvancedDeleteMessages` instead of including it in the bulk delete, or only include it in the bulk delete if the message ID is present in the slice of fetched messages?

PR #926, opened 2021-05-13
add triggering channel to report embed

l-zeuch:
Noticed that it was quite handy to just jump to the channel with the old report message, so this pull request will also add the channel and link it.

We also mention the target user, as quickly calling up the profile might also come in handy, despite the classic Discord bug with `@invalid-user`.

Thanks in advance!


l-zeuch:
Closing, as you're already working on this.

PR #928, opened 2021-05-14
New: StdCommands: Covid statistics

mrbentarikau:
Because this virus still keeps going on, this command maybe is helpful. Have used it for a half year, works.


mrbentarikau:
I really don't know why it reds out, everyting seems to be OK

jo3-l:
looks like there's an unrelated change lumped in there:
Image of commits

mrbentarikau:
> 
> 
> looks like there's an unrelated change lumped in there:
> Image of commits

not a problem

PR #940, opened 2021-05-22
Auto comments for PRs against master

l-zeuch:
Due to the not infrequent PRs for `swearwords.go` against master and the like, I was so free and set up a workflow which automatically comments on such pull requests.

A pull request to `master` without changes to `swearwords.go` will mention the user to please not do that and instead change the base branch.

A pull request updating `swearwords.go` (again against `master`) will tell the user that we will **not** update legacy automoderator, however if they still wish to request critical changes, that they please change the base branch.

Both messages include a short description on how to change the base branch.

Was only able to test it on a private repository with limited capacities, but it works fine as far as I can tell.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,28 @@
+name: Comment PR
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    name: Comment on Pull Request
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - if: ${{ !hashFiles('**/swearwords.go') }}
+        name: Comment
+        uses: thollander/actions-comment-pull-request@master
+        with:
+          message: "Hey @${{ github.actor }}! Thank you for your pull request.\nHowever, please do not open pull requests against the `master` branch, as stated in [CONTRIBUTING.md](../../CONTRIBUTING.md):\n> push the changes you make to the dev branch (Jonas747/yagpdb/dev).\n\nIf you still wish to keep this pull request open, please change the base branch.\nFor that, simply click the edit button next to the title of this pull request and select the new base;\nTo prevent conflicts, I suggest you first `git merge upstream/dev` on your local branch and push these new changes. Alternatively, `git rebase upstream/dev` on your local branch and then force-push."
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - if: ${{ hashFiles('**/swearwords.go') }}
+        name: Comment
+        uses: thollander/actions-comment-pull-request@master
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/action.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: Comment on Pull Request
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      - if: ${{ !hashFiles('**/swearwords.go') }}
+        name: Comment
+        uses: thollander/actions-comment-pull-request@master
+        with:
+          message: "Hey @${{ github.actor }}! Thank you for your pull request.\nHowever, please do not open pull requests against the `master` branch, as stated in [CONTRIBUTING.md](../../CONTRIBUTING.md):\n> push the changes you make to the dev branch (Jonas747/yagpdb/dev).\n\nIf you still wish to keep this pull request open, please change the base branch.\nFor that, simply click the edit button next to the title of this pull request and select the new base;\nTo prevent conflicts, I suggest you first `git merge upstream/dev` on your local branch and push these new changes. Alternatively, `git rebase upstream/dev` on your local branch and then force-push."
```
jo3-l, line 20: Personally found this a bit hard to read (long line), YAML has multiline strings which you could look into for this.

.github/workflows/action.yml
```diff
@@ -0,0 +1,28 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - if: ${{ hashFiles('**/swearwords.go') }}
+        name: Comment
+        uses: thollander/actions-comment-pull-request@master
+        with:
+          message: "Hey @${{github.actor }}! Thank you for your interest in contributing to YAGPDB.\nUnfortunately, basic automoderator will not receive **any** updates. To be able to use a custom list of swearwords, please use [automoderator v2](https://docs.yagpdb.xyz/tools-and-utilities/automoderator-v2).\n\nAdditionally, please do not open pull requests against the `master` branch, as stated in [CONTRIBUTING.md](../../CONTRIBUTING.md):\n> push the changes you make to the dev branch (Jonas747/yagpdb/dev).\n\nIf you still wish to keep this pull request open, please change the base branch.\nFor that, simply click the edit button next to the title of this pull request and select the new base;\nTo prevent conflicts, I suggest you first `git merge upstream/dev` on your local branch and push these new changes. Alternatively, `git rebase upstream/dev` on your local branch and then force-push."
```
jo3-l, line 27: ditto

l-zeuch: I'll look into those, wasn't entirely sure how the action takes it. Shouldn't be too hard though -- iirc it was something like

```yaml
key: >-
  Text goes here
```
?

jo3-l: I recall there's multiple ways, but the one I'm personally familiar with is this:
```yaml
key: |
  line one
  line two
  line three
```

l-zeuch: Got it! Thanks  : )

l-zeuch:
Now using multiline mode. Had to also use a workaround to get the modified files, because `hashFiles()` works differently than I initially understood.

.github/workflows/action.yml
```diff
@@ -0,0 +1,50 @@
...
+            Hey @${{ github.actor }}! Thank you for your pull request.
+            However, please do not open pull requests against the `master` branch, as stated in [CONTRIBUTING.md](../../CONTRIBUTING.md):
+            > push the changes you make to the dev branch (Jonas747/yagpdb/dev).
+            
+            If you still wish to keep this pull request open, please change the base branch.
+            For that, simply click the edit button next to the title of this pull request and select the new base;
+            To prevent conflicts, I suggest you first `git merge upstream/dev` on your local branch and push these new changes. Alternatively, `git rebase upstream/dev` on your local branch and then force-push.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - if: ${{ contains(steps.changed-files.outputs.modified_files, 'noCopy.txt') }}
```
jo3-l, line 34:
```suggestion
      - if: ${{ contains(steps.changed-files.outputs.modified_files, 'swearwords.go') }}
```
copy-paste mistake? :-)

l-zeuch: hah, yes

PR #942, opened 2021-05-23
Fix app.env

nexxydev:
fixed an issue


l-zeuch:
I don't see how that is a fix? It works even with the Bot prefix. Plus, please don't PR to master...

Wolveric:
This... seems like a reductive change... as it's removing an informative comment, just to match up the format for the least line with the rest of them...

fluffyfirefly:
It's not an issue, it works with or without "Bot", due to yagpdb/common/common.go line 136.
Although it can cause issues if you misplace quotation marks or leave out space. This was cause for issues in #selfhosting-discussion (support server) multiple times now.
Merging this PR is not really necessary, but the commit itself could potentially help reduce further issues (by reducing user error).
But in that case you'd also have to update yagpdb/cmd/yagpdb/sampleenvfile to reflect the same edit.

nexxydev:
Thaught it was an issue, according to ppl it wasnt, its not to match the format, its because if me and a friend include bot, it doesnt boot.

PR #948, opened 2021-05-27
moderation: Limit ban and kick reason length

l-zeuch:
As title states.
Because Discord only allows for a max reason length of 512 characters, we limit the reason to 470 characters, as we prepend it with the username and discriminator as well as two formatting characters.
As it stands, YAGPDB just responds with 
> Something went wrong when running this command, either discord or the bot may be having issues.

This pull request adds a check and responds with
> Error: Reason too long (can be max 470 characters).

Should the reason's length cause it to exceed max character count.

Relevant [suggestion](https://discord.com/channels/166207328570441728/356486960417734666/844968308360216646) on the support server.

Thanks in advance!

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"emperror.dev/errors"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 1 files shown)


moderation/commands.go
```diff
@@ -149,6 +149,10 @@ var ModerationCommands = []*commands.YAGCommand{
 				return nil, err
 			}
 
+			if len(reason) > 470 {
```
jonas747, line 152: use utf8.RuneCount instead, as that's also what discord goes by afaik

moderation/commands.go
```diff
@@ -238,6 +242,10 @@ var ModerationCommands = []*commands.YAGCommand{
...
```
jonas747, line 245: here too

jonas747 (changes requested):

l-zeuch:
Should be good now 👍

PR #950, opened 2021-05-31
reddit: disallow all mentions

l-zeuch:
Sometimes, when reddit users think they're funny, they use the underlying format for role and/or user mentions, which of course then mentions that role/user.

As webhooks also have an `allowed_mentions` property, setting this to empty should fix this kind of behavior - this is what this pull request aims to accomplish.
I wasn't entirely sure how webhooks behave, but I think it's better to be safe than sorry.

Couldn't test it on my selfhosted instance, as I host locally and reddit wants a legit URI for my application, different from `localhost`.

Thanks in advance!


l-zeuch:
Nice, CI hitting concurrency limit ._.

PR #952, opened 2021-06-01
templates: add reQuoteMeta func to escape metachars

jo3-l:
Make `regexp.QuoteMeta` accessible in templates.
Often users want to embed strings into regexps to be matched literally (i.e. a server prefix), currently they have to do one of two things: use `\Q...\E`, or escape all metacharacters using `reReplace`. Both are easy to get wrong: it's easy to overlook an important detail when using the former approach or forget a metacharacter when using the latter.


ashishjh-bst:
Has no conflicts, but please rebase to master as it shows commits with changes which are already in master.

PR #954, opened 2021-06-01
reminders: add -channel flag to set reminder channel

### LINKED ISSUE #953: Reminders in Other Channel
It would be nice if we could set reminders for other channels. That way all of the reminder setting could be done in a Bots channel to "hide" the clutter.

This would most likely just require a tweak to the existing Reminder interface, and since the bot is already able to do things across channels, it should already have the code to do it. So just a bit of copy/paste/frankensteining should be needed.

jo3-l:
Allow users to set the channel in which their reminder is sent, provided they have read and send message permission in said channel.
This was a user suggestion on the server, also fixes #953.
Works fine on selfhost.


ashishjh-bst:
Rebase is needed

jo3-l:
Rebased.

PR #956, opened 2021-06-02
Create reminders in other channels;

grantemj:
Also allowed  to list reminders in other channels

issue #953


grantemj:
Yup, my bad. I just realized I did the wrong branch. Imma fix it

PR #969, opened 2021-06-10
templates: indirect more values

jo3-l:
Continuation of #963, but with a few more ones that were missed.
This should be all, I checked through all the code in templates/ that might be affected and there's nothing else that might pose an issue as far as I see.
Works as expected on selfhost.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -247,8 +247,20 @@ func (c *Context) checkSafeStringDictNoRecursion(d SDict, n int) bool {
 			}
 		}
 
+		if cast, ok := v.(*Dict); ok {
+			if !c.checkSafeDictNoRecursion(*cast, n+1) {
+				return false
+			}
+		}
+
 		if cast, ok := v.(SDict); ok {
 			if !c.checkSafeStringDictNoRecursion(cast, n+1) {
+				return false
+			}
+		}
+
+		if cast, ok := v.(*SDict); ok {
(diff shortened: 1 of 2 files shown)


common/templates/context_funcs.go
```diff
@@ -277,12 +277,24 @@ func (c *Context) checkSafeStringDictNoRecursion(d SDict, n int) bool {
 			}
 		}
 
+		if cast, ok := v.(*Dict); ok {
+			if !c.checkSafeDictNoRecursion(*cast, n+1) {
+				return false
+			}
+		}
+
```
jo3-l, line 285: This whole block of code would probably be better written as a type switch given the number of type casts, but I didn't want to accidentally cause more breakage : )

common/templates/general.go
```diff
@@ -716,14 +718,11 @@ func shuffle(seq interface{}) (interface{}, error) {
 		return nil, errors.New("can't iterate over a nil value")
 	}
 
-	switch seqv.Kind() {
-	case reflect.Array, reflect.Slice, reflect.String:
-		// okay
-	default:
+	if seqv.Kind() != reflect.Slice {
```
jo3-l, line 721: This was changed as shuffle never worked with strings or arrays in the first place.
MakeSlice panics if the passed Type does not have kind Slice.

jo3-l:
Friendly bump @nitinr-bst :) would be nice to get this in.

PR #970, opened 2021-06-11
show server icon besides server name in server selection

Spongerooski:
suppose I have 2 servers with the same name but different icons, this helps differentiate between them.
plus, I think it looks better than just a simple wrench

I have tested this on a selfhost and it seems to work fine 😃

### DIFF
--- a/cmd/yagpdb/templates/cp_nav.html
+++ b/cmd/yagpdb/templates/cp_nav.html
@@ -92,7 +92,7 @@
                     </li>
                 </ul>
             </div>
-        </div> 
+        </div>
 
         <script>
         $(function(){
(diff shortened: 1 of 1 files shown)


l-zeuch:
+1 on this, I can see it adding a better experience with server navigation (I sometimes happen to misclick the server, an icon could help)

jonas747:
Has this been tested on mobile?

fluffyfirefly:
Seems to work as intended on mobile, tested with Firefox mobile on a selfhosted YAG instance.

Spongerooski:
^, works fine on mobile
image

Spongerooski:
Do you think it would be a better idea to show the question mark image instead of the wrench if a server has no icon?

l-zeuch:
Would leave it as wrench to be fair, people are used to that now. 
A question mark *may* confuse users into thinking that their server is unavailable or something along the line of that.

PR #971, opened 2021-06-11
workflows: Remove unnecessary stuff, fix minor mistakes

l-zeuch:
Found a small-ish mistake from #940. This pull request fixes that.

Sorry for the inconvenience :)

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -17,7 +17,7 @@ jobs:
         id: changed-files
         uses: tj-actions/changed-files@v2.1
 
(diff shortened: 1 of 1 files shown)


l-zeuch:
Addendum:
As most `swearwords.go` PRs are made by first-time contributors, this workflow requires prior approval. This kind of defeats the purpose, hence I removed it entirely. It'll now only comment on PRs against master in general (for contributors).

This limitation is imposed by GitHub themselves, which is rather annoying, but it is how it is.

PR #977, opened 2021-06-14
Fix potential panic from whois

Pedro-Pessoa:
`Whois` was dereferencing `member.Member` without checking if it was nil first, thus, in some cases where the `Member` fields is not present in the `dstate.MemberState` object, the bot would panic.


BlackWolfWoof:
Me who uses exec "whois" 780~ times per hour be like 😳

PR #978, opened 2021-06-15
templates: qol improvements to setRoles

jo3-l:
Some QoL improvements to setRoles to make it nicer to use based on user feedback.

1. Deduplicate input slice of roles, this makes it easier to write code that adds multiple roles to users in a single go. Let's say a user had 249 roles, and we want to add 5 roles to them. If we just add those 5 roles to their member's role slice, it's going to error currently. It would be nice if we filtered out duplicates in the input slice so it "just worked".
2. Add in managed roles automatically. You can't remove managed roles, so currently users have to take into account managed roles and add them to the input slice manually. We can do that for them instead.
3. Support `nil` for member parameter (uses the context member).
4. Support role structs as input, this is just something nice to have.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -696,45 +696,89 @@ func (c *Context) takeRole(targetID int64, roleID int64, delay time.Duration) st
 	return ""
 }
 
-func (c *Context) tmplSetRoles(target interface{}, roleSlice interface{}) (string, error) {
+const DiscordRoleLimit = 250
+
+func (c *Context) tmplSetRoles(target interface{}, input interface{}) (string, error) {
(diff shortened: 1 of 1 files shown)


jonas747 (changes requested):
If were using the context member, on success we should maybe clone the member and update the roles and swap out the context member with the updated clone

as remember, the context member will not get updates automatically from state.

jo3-l:
If you'd like to me to do that I'm happy to, but I'm curious as to why this should be done in `setRoles` only? In particular, I don't think any of the other role functions do this.

jonas747:
Hm yeah good point, the other role funcs should do this, for now ill merge this i guess

PR #979, opened 2021-06-15
stdcommands: more answers to advice

l-zeuch:
Quite an old [suggestion](https://discord.com/channels/166207328570441728/356486960417734666/692984043389845524) which apparently was not implemented or PRd yet.

This pull request adds a little sugar to the advice command: Instead of always indexing the first item upon a successful search, we take a random item of the returned array of advice slips. Not a huge change, however I tested it as usual on a selfhosted instance and the results speak for themselves:
screenshot
(sorry for compact mode - ATBDB is my selfhosted instance, LAGPDB is the officially hosted instance).

As you can clearly see, there is now more "advice" to receive, the command still functions the same. I was also so free to remove the unnecessarily redundant type declaration for the ArgDef.

Cheers : )

### DIFF
--- a/stdcommands/advice/advice.go
+++ b/stdcommands/advice/advice.go
@@ -2,6 +2,7 @@ package advice
 
 import (
 	"encoding/json"
+	"math/rand"
 	"net/http"
 	"net/url"
 
@@ -15,7 +16,7 @@ var Command = &commands.YAGCommand{
 	Name:        "Advice",
 	Description: "Don't be afraid to ask for advice!",
 	Arguments: []*dcmd.ArgDef{
-		&dcmd.ArgDef{Name: "What", Type: dcmd.String},
+		{Name: "What", Type: dcmd.String},
 	},
 	DefaultEnabled:      true,
 	SlashCommandEnabled: true,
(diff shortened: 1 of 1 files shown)


Pedro-Pessoa:
Shouldn't you `rand.Seed` before?

l-zeuch:
Never heard of that tbh... What does it do and what benefit would it bring? Isn't it already seeded when we make the call to the function?

Pedro-Pessoa:
From the GoDocs:
`Seed uses the provided seed value to initialize the default Source to a deterministic state. If Seed is not called, the generator behaves as if seeded by Seed(1). Seed values that have the same remainder when divided by 2³¹-1 generate the same pseudo-random sequence. Seed, unlike the Rand.Seed method, is safe for concurrent use.`

If you don't seed the result will always be the same if `len` is the same. 

Usually we seed time as it an always changing value: `rand.Seed(time.Now().Unix())`

l-zeuch:
Ah, gotcha. Will push changes in a minute!
(You learn something new everyday)

jonas747:
Seeding rand here is not needed at all, there are so many other things calling rand that its random enough, and 
> If you don't seed the result will always be the same if len is the same.

is also not true, https://play.golang.org/p/WyXoa0VaTt-

PR #980, opened 2021-06-16
logs: show underlying JSON of embeds

l-zeuch:
As title states.
This was also a fairly old [suggestion](https://discord.com/channels/166207328570441728/356486960417734666/754098618620706906) on the support server:
> For logging with the `-logs` command, specifically logging embeds, use the JSON version of the embed so that there is at least some form of embed logging, rather than "(x embeds is not shown)".

This pull request adds this kind of behavior, although I was not able to indent it beautifully for some reason. 
Let me know what needs to be changed for it to do that, or if this is intentional.
I decided to take this action on all embeds of each message, if they exist. Feedback regarding this is appreciated, as well.

As usual, I tested this on an selfhosted instance and as far as I can tell, it works fine:
screenshot of logs on selfhost

Cheers!

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -4,6 +4,7 @@ package logs
 
 import (
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


jonas747 (changes requested):
I don't think we should use marshalindent here, as were not even using the indentation.

l-zeuch:
True. I think I've left it in there for testing purposes and forgot to change it back. Will push changes in a hot second 👍

l-zeuch:
Pushed the changes, should be good now. Let me know if I missed something else!

PR #981, opened 2021-06-16
Reddit: disable feeds option

mrbentarikau:
To disable and not just to delete a Reddit feed has been requested several times... Looking at the back end of reddit code, it looks like nothing dramatical needs to be changed; None/0 channels in database are already avoided and disabled. So this little change enables front-end possibility for users to disable their feed.

### DIFF
--- a/reddit/assets/reddit.html
+++ b/reddit/assets/reddit.html
@@ -54,6 +54,7 @@
             for that.{{end}}</p>
         <p>The subreddit field is just the name of the subreddit (no /r/ in front of it), examples: "games",
             "multicopter"</p>
+        <p><b>If Server Channel is set to "None" the added or already active Reddit feed will be disabled.</b></p>
     </div>
     <div class="col-md-6">
         <h3>New feed</h3>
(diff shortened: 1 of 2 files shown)


Wolveric:
I feel like a command to toggle these setting from within Discord, would be a notable inclusion for this feature.

mrbentarikau:
reddit is just totally outside of discord, only manageable through control panel

Wolveric:
It's just I think enabling such functionality, would give the feature a greater use.

mrbentarikau:
This all needs Jonas' opinion anyway because there are other ways to disable a feed, like a button : ) And how disabling a feed will affect the re-enabled one later on.

l-zeuch:
-1 on a new command that's likely to be never used.
Control panel is fine.
> YAGPDB is nearly exclusively configured through its control panel. If you want to make any major configuration changes, you'll have to go through there.

: )

jonas747:
Maybe add a check on the webserver to set the `disabled` field if it's 0, i know this happens somewhere down the line in mqueue or somwhere but this would be better IMO

mrbentarikau:
> Maybe add a check on the webserver to set the `disabled` field if it's 0, i know this happens somewhere down the line in mqueue or somwhere but this would be better IMO

Currently it yeah goes automatically disabled if next post comes in and no channel is found.

mrbentarikau:
I hope you meant something like this ^

jonas747:
looks about right

Wolveric:
> -1 on a new command that's likely to be never used.
> Control panel is fine.
> > YAGPDB is nearly exclusively configured through its control panel. If you want to make any major configuration changes, you'll have to go through there.
> 
> Feel like disabling a reddit feed is a major configuration.
Listen here you little-

PR #982, opened 2021-06-16
Docker Improvements

fluffyfirefly:
This PR is intended to improve the current Docker support.

Dockerfile:
Added Entrypoint and modified CMD:
- The current Dockerfile does not match the syntax of docker-compose.yml and docker-compose.proxied.yml.
Those have currently no specification of the executable path and thus would fail to start a container from an image build by the current Dockerfile. 
Additional, the current CMD is written in shell format, which should be exec format instead.
Since the yagpdb_app container is only supposed to run an executable and nothing else, it would be reasonable to change the current shell format CMD to an exec format Entrypoint (/app/yagpdb) + CMD for additional commands.

Removed ENV extra_flags:
- There should be no reason to have such an ENV in the running container, since you can already specify those with CMD/command

Moved copy operations to builder:
- Using /appbuild/yagpdb/ as workdir for mod download and go build. Move all the necessary files + binary to /bin/yagpdb so we can copy all files in one image layer to the final image. And importantly exclude the unnecessary files in yagpdb/cmd/yagpdb/ (build.sh; copytemplates.sh; main.go; sampleenvfile) since they are not required inside the image and only take up unnecessary space. 

Formating changes:
- Removed white space in front of the "RUN apk" lines since it will result in a weird syntax on docker hub
edit: changed "RUN apk" and "volume" to one-liners.

docker-compose.yml:
- Added commented executable path in command, since it is currently used in the public community image by teyker/teyloll

docker-compose.proxied.yml:
- Updated to match docker-compose.yml

docker-compose.dev.yml:
- Commented out the executable path in order to match the new Dockerfile

### DIFF
--- a/yagpdb_docker/Dockerfile
+++ b/yagpdb_docker/Dockerfile
@@ -1,47 +1,40 @@
 FROM golang:stretch as builder
 
-WORKDIR /appbuild/yagpdb/
-
-COPY go.mod .
-COPY go.sum .
-
+WORKDIR /appbuild/yagpdb
+COPY go.mod go.sum .
 RUN go mod download
-
 COPY . .
 
-WORKDIR /appbuild/yagpdb/cmd/yagpdb
+WORKDIR /bin/yagpdb
+COPY cmd/yagpdb/posts posts/
+COPY cmd/yagpdb/static static/
+COPY cmd/yagpdb/templates templates/
+# Handle templates for plugins automatically
+COPY */assets/*.html templates/plugins/
 
-RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-X github.com/jonas747/yagpdb/common.VERSION=$(git describe --tags)" -v
+WORKDIR /appbuild/yagpdb/cmd/yagpdb
+RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-X github.com/jonas747/yagpdb/common.VERSION=$(git describe --tags)" -v -o /bin/yagpdb
+
+
 
 FROM alpine:latest
 
 WORKDIR /app
-VOLUME /app/soundboard \
+VOLUME \
+  /app/soundboard \
   /app/cert
 EXPOSE 80 443
 
 # We need the X.509 certificates for client TLS to work.
-# Also install tzdata, needed for Go timezone support, 
+# Also install tzdata, needed for Go timezone support,
 # and ffmpeg for soundboard support,
-# as the alpine image does not come with them by default. 
-RUN apk --no-cache add ca-certificates \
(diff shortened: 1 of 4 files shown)


l-zeuch:
Just a general question: #904 was meant to decrease the overhead when building the image - that is still the case?
I'm not very proficient with reading Dockerfiles 🤔

fluffyfirefly:
#904 was meant to implement build caching by caching the already build go-module image-layer and thus improve build times. This is still the case, I merely changed the copy lines for go.mod and go.sum into a single line.

yagpdb_docker/Dockerfile
```diff
@@ -1,47 +1,40 @@
...
-COPY --from=builder /appbuild/yagpdb/cmd/yagpdb/templates templates/
-COPY --from=builder /appbuild/yagpdb/cmd/yagpdb/posts posts/
-COPY --from=builder /appbuild/yagpdb/cmd/yagpdb/static static/
-
-COPY --from=builder /appbuild/yagpdb/cmd/yagpdb .
+# as the alpine image does not come with them by default.
+RUN apk --no-cache add \
+ca-certificates \
+ffmpeg \
+tzdata
```
jo3-l, line 35:
```suggestion
  ca-certificates \
  ffmpeg \
  tzdata
```
To be consistent with the rest of the multi-line instructions here.

l-zeuch:
Ah, gotcha. Thanks for the answer 👍

jo3-l (commented):
very minor formatting suggestion

yagpdb_docker/Dockerfile
fluffyfirefly: I would generally agree, but that is what I meant by "weird syntax on docker hub".
Docker hub lists the contents of the image layers in a single line. 
If we add the white space in front of the dependencies, then that would result in double whitespace in the detail pane on docker hub.
Changing it into a one-liner instead might be the better resolution with only 3 dependencies. 

Image of docker hub with white space:
image

without white space:
image

jo3-l: Fair enough, TIL : )

fluffyfirefly: removing the white space at the end of the line instead would also work:
```suggestion
RUN apk --no-cache add\
 ca-certificates\
 ffmpeg\
 tzdata
```

or just write it all in one line:
```suggestion
RUN apk --no-cache add ca-certificates ffmpeg tzdata
```

Really not sure on that one

fluffyfirefly: Changed it into a one-liner

PR #983, opened 2021-06-16
amod: create Send Message Effect

Pedro-Pessoa:
This feature has been requested a lot on the support server, so I decided to make it.
This is a Send Message Effect for AutomodV2.

It has no custom command involved whatsoever. Users will only be able to send a string response.
I added an option to ping the triggering user.

I have made some decisions and would like opinions on them:
- I made the message length limit be 150;
- This will not be triggered by other bots.

What do you think of these limitations?


l-zeuch:
I feel like 500 (or 512 :^) might be better. Best example is Twitter, you mostly cannot quite express yourself with such a hefty limitation.

Not being triggered by bots sounds good.

Pedro-Pessoa:
That's fair enough. 512 sounds reasonable enough to me. Will wait for more inputs.

mrbentarikau:
> I feel like 500 (or 512 :^) might be better. Best example is Twitter, you mostly cannot quite express yourself with such a hefty limitation.
> 
> Not being triggered by bots sounds good.

AMv2 should not be a writing club, so automated responses should not be 512 long, Twitter limit of 140 is quite fair

l-zeuch:
> AMv2 should not be a writing club, so automated responses should not be 512 long, Twitter limit of 280 is quite fair

Fair enough. 280 sounds like a good meeting point

Pedro-Pessoa:
Increased the limit to 280 chars!

PR #989, opened 2021-06-18
templates: improve randInt error when start >= stop

jo3-l:
A rather cryptic error pops up when you run `randInt a b` where `a >= b`: `invalid argument to Int63n`. Improve it a bit so that the error is easier to understand without needing to look through Go docs for what `Int63n` is.


jo3-l:
Rebased.

PR #993, opened 2021-06-20
Docker Improvements again

fluffyfirefly:
Dockerfile:
- Revert the move of copy operations to builder from the last Docker Improvements PR  #982, and instead run them in the final image again to make use of image layer download caching. 
Having separate image layers for static/ and templates/ in the final image allows us to cache these when pulling new images from Dockerhub. This means we only need to re-download the image layer with the binary and don't need to re-download the image layers for static/ and templates/ if they weren't changed.
This reduces the total download traffic when pulling new images, due to static being close to the size of the binary itself.
- However, we leave the copy-templates operation in the builder so we only have one image layer for all templates.
- And we remove yagpdb/cmd/yagpdb/posts due to it being obsolete.

- Reduced the wall of text for the RUN apk comment to acceptable levels.
- Moved the "-v" option in docker build in front of "-ldflags" for better readability.

app.example.env:
- Commented out Twitter credentials, due to them causing the bot to error out if not edited and run with the default placeholder keys. (cmd/yagpdb/sampleenvfile already has it like this)

### DIFF
--- a/yagpdb_docker/Dockerfile
+++ b/yagpdb_docker/Dockerfile
@@ -3,17 +3,13 @@ FROM golang:stretch as builder
 WORKDIR /appbuild/yagpdb
 COPY go.mod go.sum .
 RUN go mod download
+
+# Handle templates for plugins automatically
+COPY */assets/*.html cmd/yagpdb/templates/plugins/
 COPY . .
 
-WORKDIR /bin/yagpdb
-COPY cmd/yagpdb/posts posts/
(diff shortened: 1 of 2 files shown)


fluffyfirefly:
Changed the "COPY go.mod go.sum" line from "." to "./" since docker build requires it while docker buildx can run without it.
"When using COPY with more than one source file, the destination must be a directory and end with a /"

PR #1000, opened 2021-07-08
templates: Expose `HasPrefix` and `HasSuffix` from the Go library

l-zeuch:
This was a suggestion on the support server, exposing the inbuilt functions `HasPrefix` and `HasSuffix` from the Go `strings` library.

As far as I can tell, it works fine on selfhost - this is the first time that I added things to the scripting language, so any advice / tips are highly appreciated.

Thanks in advance!

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -42,6 +42,8 @@ var (
 		"urlescape": url.PathEscape,
 		"split":     strings.Split,
 		"title":     strings.Title,
+		"hasPrefix": strings.HasPrefix,
+		"hasSuffix": strings.HasSuffix,
 
 		// math
 		"add":               add,


common/templates/context.go
```diff
@@ -68,6 +68,8 @@ var (
 		"complexMessage":     CreateMessageSend,
 		"complexMessageEdit": CreateMessageEdit,
 		"kindOf":             KindOf,
+		"hasPrefix":          HasPrefix,
+		"hasSuffix":          HasSuffix,
```
jo3-l, line 72:
```suggestion
		"hasPrefix":          strings.HasPrefix,
		"hasSuffix":          strings.HasSuffix,
```

common/templates/general.go
```diff
@@ -131,6 +131,14 @@ func KindOf(input interface{}, flag ...bool) (string, error) { //flag used only
...
 }
 
+func HasPrefix(input, prefix string) bool {
+	return strings.HasPrefix(input, prefix)
+}
+
+func HasSuffix(input, prefix string) bool {
+	return strings.HasSuffix(input, prefix)
+}
+
```
jo3-l, line 141:
```suggestion
```

jo3-l (commented):
untested, but could this not simply be:

common/templates/general.go
l-zeuch: sounds like it could work, i'll test it tomorrow :ye:

l-zeuch:
Yup, works fine and we actually already use some things like that from the `strings` library :+1:

PR #1002, opened 2021-07-14
stdcommands: add switch to weather cmd to change the unit system

l-zeuch:
Small one, but fairly frequent request to please use imperial units for the weather command.
Reading through the API's documentation, this is more than doable: just change the unit parameter from `m` to `u`.

This pull request will add a switch `imperial` to the command, which then allows users to change the unit system from metric to imperial. Backwards compatibility is therefore kept.

As far as I can tell, everything works fine on a selfhost:

screenshot

### DIFF
--- a/stdcommands/weather/weather.go
+++ b/stdcommands/weather/weather.go
@@ -23,14 +23,24 @@ var Command = &commands.YAGCommand{
 	RunInDM:      true,
 	RequiredArgs: 1,
 	Arguments: []*dcmd.ArgDef{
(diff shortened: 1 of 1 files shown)


jonas747:
So there already was code in to show the temperature in both units, but it seems like they serve 2 different formats so i've gone ahead and added a fix for that in https://github.com/jonas747/yagpdb/commit/2bc9ffd7974e6b4fd3d99ef019a17dbdc98fa9a7

image

l-zeuch:
Gotcha - thanks for taking a look though!

PR #1015, opened 2021-07-31
commands: add a note explaining the effects of prefix change on flags

LemmeCry:
Users often get confused and think flags also change along with the prefix. This PR adds a note talking about this, and also uses `rolemenu create` instead of `ping` in the example, to show the usage of flags with the user's set prefix.

image


jo3-l (changes requested):
The following line also needs to be changed.
https://github.com/jonas747/yagpdb/blob/7ac10f4ab71c4cc35fb423e02dcc3034dbb5af0d/commands/assets/commands.html#L133

PR #1035, opened 2021-09-03
Remove payment methods other than Patreon

PlavorSeol:
Since payment methods other than Patreon were removed from [Premium setup](https://yagpdb.xyz/premium) page (#1003) as they are no longer accepted, let's remove leftovers of mentions of them on other pages as well.


3kh0:
That is a good idea ngl

PR #1037, opened 2021-09-05
web: use bigints for channel permissions

jo3-l:
Use bigints for channel permissions, as discussed on Discord. This fixes the issue where the control panel claims that YAGPDB's missing some permissions despite it actually having said permissions.

Tested on a few control panel pages with various bot permissions on selfhost, and it appears to work fine.

### DIFF
--- a/frontend/static/js/spongebob.js
+++ b/frontend/static/js/spongebob.js
@@ -343,15 +343,15 @@ function initPlugins(partial) {
 var discordPermissions = {
 	read: {
 		name: "Read Messages",
-		perm: 0x400
+		perm: BigInt(0x400),
 	},
 	send: {
 		name: "Send Messages",
-		perm: 0x800
+		perm: BigInt(0x800),
 	},
 	embed: {
 		name: "Embed Links",
-		perm: 0x4000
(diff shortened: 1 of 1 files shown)


nitinr-bst (commented):
LGTM, will be merged when we plan our next release :)
However there is still an issue where even if bot is given Admin perm but some Role is assigned to bot lets say role1 which overrides send msg perm in a channel, it doesn't work, but it should. 
This fix doesn't solve that issue I believe?

3kh0:
Seems stable

jo3-l:
Hi @nitinr-bst, thank you for the review. I'm admittedly unfamiliar with the issue you refer to, but to be clear, this patch only fixes the frontend such that large permission bit values are interpreted correctly and not truncated, while the issue you refer to would likely need to be fixed in the backend.

Side note: I have not tested this, but the code that handles the `channelperms` endpoint appears to already take channels into account:
https://github.com/botlabs-gg/yagpdb/blob/10dcb5154cbeab9d723fac61ad358bdfa66983ed/bot/botrest/server.go#L202

nitinr-bst (approved):

PR #1038, opened 2021-09-12
Added case "image" to createMessageSend declaration so yag can post l…

mangowhite:
This was made by @Spongerooski to make our self-host be able to post image attachments. It simply adds a new case to complexMessage, so the usage would be:
(complexMessage "image" "link") 
Yag will then post the given link as an attachment. This is useful for a lot of things, such as logging; It is also a very good work around for broken images (which happens when the attachment is deleted and you have a YAG message with it).
[Example](https://media.discordapp.net/attachments/808023716847681567/885096788468576276/AmEXUzjkGD.gif)

I hope my PR was understandable, if you have any questions please feel free to DM me in discord or ping me in the YAGPDB support server: @mangowhite#0001

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -244,7 +244,7 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 
 		case "image":
 			if val == nil {
-				return nil, errors.New("need an arguement for the image")
+				return nil, errors.New("need an argument for the image")
 			}
 			str := fmt.Sprint(val)
 			img, err := http.Get(str)
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"math"
 	"math/rand"
 	"reflect"
@@ -240,6 +241,29 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 				ContentType: "text/plain",
 				Reader:      &buf,
 			}
+
+		case "image":
+			if val == nil {
+				return nil, errors.New("need an arguement for the image")
+			}
+			str := fmt.Sprint(val)
+			img, err := http.Get(str)
+			if err != nil {
+				return nil, errors.New("error while trying to get the image")
+			}
+
+			imageName := "image.png"
+			imageType := "image/png"
+			if strings.HasSuffix(strings.ToLower(str), ".gif") {
+				imageName = "image.gif"
+				imageType = "image/gif"
+			}
+			msg.File = &discordgo.File{
+				Name: imageName,
+				ContentType: imageType,
+				Reader: img.Body,
+			}
+
 		default:
 			return nil, errors.New(`invalid key "` + key + `" passed to send message builder`)
 		}


common/templates/general.go
```diff
@@ -240,6 +241,29 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 				ContentType: "text/plain",
 				Reader:      &buf,
 			}
+
+		case "image":
+			if val == nil {
+				return nil, errors.New("need an arguement for the image")
```
jo3-l, line 247:
```suggestion
				return nil, errors.New("need an argument for the image")
```
Small typo.

Side note: I don't really feel that making YAGPDB fetch potentially very large user-provided images is a great idea, but obviously the decision regarding whether or not this PR will be accepted isn't up to me : )

common/templates/general.go
```diff
@@ -240,6 +241,29 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
...
+			if strings.HasSuffix(strings.ToLower(str), ".gif") {
+				imageName = "image.gif"
+				imageType = "image/gif"
+			}
+			msg.File = &discordgo.File{
+				Name: imageName,
+				ContentType: imageType,
+				Reader: img.Body,
+			}
+
```
Pedro-Pessoa, line 266: I think you should img.Body.Close() here.

common/templates/general.go
```diff
@@ -240,6 +241,29 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
...
+			imageName := "image.png"
+			imageType := "image/png"
+			if strings.HasSuffix(strings.ToLower(str), ".gif") {
+				imageName = "image.gif"
+				imageType = "image/gif"
+			}
+			msg.File = &discordgo.File{
+				Name: imageName,
+				ContentType: imageType,
+				Reader: img.Body,
```
Pedro-Pessoa, line 264: Have you checked what happens if the attachment is not an image? 
Perhaps this should be handled. 
Dunno if this can be maliciously used.

common/templates/general.go
```diff
@@ -240,6 +241,29 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 				ContentType: "text/plain",
 				Reader:      &buf,
 			}
+
+		case "image":
+			if val == nil {
+				return nil, errors.New("need an argument for the image")
+			}
+			str := fmt.Sprint(val)
+			img, err := http.Get(str)
```
Pedro-Pessoa, line 250: You can perhaps check if `str != ""`? 
http.Get would of course handle that and error, but I think this is a better approach. Totally subjective though.

Pedro-Pessoa (commented):
A few things I found worth noting :)

common/templates/general.go
mangowhite: Why though? It works without it! :)

common/templates/general.go
mangowhite: It literally creates a file named image.png without any content. Can't be used maliciously. Tried youtube links and direct links to compressed files that are often used as a virus.
image

common/templates/general.go
Pedro-Pessoa: Just because "it works", it doesn't mean it is right :)

As per Go's http package: `The client *must* close the response body when finished with it`

common/templates/general.go
Pedro-Pessoa: I don't think this is ideal, even if it can't be used maliciously. 
Bot should definitely respond with and error, in my opinion.

common/templates/general.go
mangowhite: indeed. I'll do it then! Thanks Pedro.

common/templates/general.go
mangowhite: Alright, I'll handle that. Thanks!

LemmeCry:
-1 on this. As discussed on multiple occasions before, fetching, parsing and uploading files as attachments is resource intensive, especially with how big YAG already is. Given the choice, a vast majority of people will send images as attachments instead of URLs, and this doesn't seem like a good idea. The fact that there's no major moderation bot that already does this for the sake of logging, goes to show this. All in all, cool thing to have on a private bot limited to a few servers, but very resource intensive at the scale YAG is at.

devnote-dev (changes requested):
(Unfortunately couldn't get the code reference for this)
As far as I'm aware, Yag is unable to send multiple files in one message currently, and this code sets `msg.File` to the image in place of the `file` option which can cause unexpected functionality. While only a minor issue, I think this should wait until Yag officially supports sending multiple files.

ashishjh-bst (changes requested):
Please make changes as suggested by Pedro

ashishjh-bst:
Closing this, just sending a link to the image from the bot instead of making an http call to get the image and then upload it as an attachment, good enough for a selfhost, but for the primary bot that would lead to a lot of extra memory usage.

PR #1045, opened 2021-09-24
reputation: start using IDs in reputation commands

lemm-e:
Not being able to use `GiveRep`, `TakeRep` and `Rep` commands with IDs and having to resolve to user mentions instead, has been an inconvenience when using the reputation system. This PR aims to fix this issue by using `dcmd.UserID` instead of `dcmd.User`. Tested it on a self-hosted instance and it seems to work fine, wouldn't have PRd if it wasn't for @mrbentarikau showing how their bot does it.


jo3-l:
Does `userid` work with user (nick)names? For example, `-rep lemmecry` works currently (and it should keep working with this change, imo).

ashishjh-bst:
This was done in #1198 so closing.

PR #1050, opened 2021-10-04
youtube unique always 0

devjalo:
Selecting the correct key to get the unique youtube channels


Pedro-Pessoa:
This looks right, but why not use the const `RedisKeyWebSubChannels` instead of the raw string?

devjalo:
created a new one cause of missing permissions to edit the pr. The new pullrequest is using the const.
https://github.com/botlabs-gg/yagpdb/pull/1052

PR #1051, opened 2021-10-04
common/templates: add urlunescape template

lemm-e:
Does exactly what it says- We already have `urlescape` as a template, this PR adds `urlunescape` to unescape a url string.


ashishjh-bst:
@lemm-e  please resolve the conflicts.

PR #1052, opened 2021-10-05
youtube unique always 0

devjalo:


l-zeuch:
For context: essentially just like #1050, but using the `const` instead of the raw string.

LGTM, but that missing permission thing is quite weird.

Please don't PR to master though, dev is the base branch to be used.

PR #1065, opened 2021-11-15
common/templates: add templates to pin and unpin messages

LemmeCry:
Being able to pin/unpin messages with the bot has been a common request. This PR does just that by adding custom command templates which allow pinning/unpinning a message.
The syntax for the template usage will be:
- `pinMessage channel messageID`
- `unpinMessage channel messageID`

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -499,6 +499,8 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("sendMessageNoEscapeRetID", c.tmplSendMessage(false, true))
 	c.addContextFunc("editMessage", c.tmplEditMessage(true))
 	c.addContextFunc("editMessageNoEscape", c.tmplEditMessage(false))
+	c.addContextFunc("pinMessage", c.tmplPinMessage)
+	c.addContextFunc("unpinMessage", c.tmplUnpinMessage)
 
 	// Mentions
 	c.addContextFunc("mentionEveryone", c.tmplMentionEveryone)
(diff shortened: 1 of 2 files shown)


LemmeCry:
Those errors are handled automatically. Here is the error returned when you try to pin after the limit:
image

Wolveric:
Yeah, that's the case I mean. Because it's difficult to track that count, I'd rather it not automatically panic because of that. I'd prefer the option to gracefully handle the error, by checking the return, indicating whether it failed.

common/templates/context_funcs.go
```diff
@@ -402,6 +402,32 @@ func (c *Context) tmplEditMessage(filterSpecialMentions bool) func(channel inter
 	}
 }
 
+func (c *Context) tmplHandlePinMessage(mode bool) func(channel, msgID interface{}) (string, error) {
```
jo3-l, line 405:
```suggestion
func (c *Context) tmplPinMessage(mode bool) func(channel, msgID interface{}) (string, error) {
```
For consistency with similar higher-order functions.

common/templates/context.go
```diff
@@ -499,6 +499,8 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("sendMessageNoEscapeRetID", c.tmplSendMessage(false, true))
 	c.addContextFunc("editMessage", c.tmplEditMessage(true))
 	c.addContextFunc("editMessageNoEscape", c.tmplEditMessage(false))
+	c.addContextFunc("pinMessage", c.tmplPinMessage(true))
```
jo3-l, line 502: The values are 
```suggestion
	c.addContextFunc("pinMessage", c.tmplPinMessage(false))
```

The parameter specifies whether it should unpin.

common/templates/context.go
```diff
@@ -499,6 +499,8 @@ func baseContextFuncs(c *Context) {
...
+	c.addContextFunc("unpinMessage", c.tmplPinMessage(false))
```
jo3-l, line 503:
```suggestion
	c.addContextFunc("unpinMessage", c.tmplPinMessage(true))
```
Ditto.

jo3-l (changes requested):

PR #1067, opened 2021-11-15
common/templates: allow file renaming

l-zeuch:
User suggestion on the server, to allow for more flexibility when outputting files as custom command response.

* We add a new key to `complexMessage`; `"filename"` - its value will be the file's new name instead of `Attachment`.
    * Because this new key is optional, we will default to `Attachment`, if no filename given.
    * The custom filename will be limited to 64 characters, this seems like a reasonable limit.
* The file's extension will be hard-coded to `*.txt` as countermeasures against abuse of `*.bat`, `*.wav` or other wonky fonky file formats.

I decided on no-op when a filename but no file is given. Just doesn't make sense to give a filename without a file, and when it does happen it seemed to much to stop the CC execution over something that has no effect whatsoever.

As far as I can tell, this works fine on selfhost.
handy dandy screenshot

Thanks to Joe and mrbentarikau for helping me and giving input, and as usual thanks in advance!

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -216,11 +216,14 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 
 	msg := &discordgo.MessageSend{}
 
+	// Default filename
+	filename := "attachment_" + time.Now().Format("2006-01-02_15-04-05")
+
 	for key, val := range messageSdict {
 
 		switch key {
 		case "content":
-			msg.Content = fmt.Sprint(val)
+			msg.Content = ToString(val)
 		case "embed":
 			if val == nil {
 				continue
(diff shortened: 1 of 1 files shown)


mrbentarikau:
maybe add date to default filename, would differentiate more and no overwrite troubles
something like > "attachment_" + time.Now().UTC().Format("20060102_150405") + ".txt"

l-zeuch:
Sounds good, I'll push an update :+1:

mrbentarikau:
hehe, also... no capital letters : )

l-zeuch:
Format of default filename is now `attachment_2006-01-02_15-04-05` with https://pkg.go.dev/time#pkg-constants formatting.

Pedro-Pessoa:
I'm in favor of using `ToString` instead of `fmt.Sprint`.
Otherwise LGTM :)

l-zeuch:
What would be the benefit here? Shouldn't they behave the same?

Pedro-Pessoa:
The end result would be the same, yes. 
But `fmt.Sprint` is slower because it uses reflection and is less efficient, unlike `ToString` which uses the "proper" conversion method. 

But tbf it would probably be negligible here. It is more for consistency. If we have a func to convert types to string why not use that? :)

l-zeuch:
Consistency is a valid point. I'll push an update :+1:

PR #1069, opened 2021-11-16
common/templates: add getPinCount template

LemmeCry:
This adds the template `getPinCount channel` which returns the total number of pins in a channel. The command is limited to 2 calls per CC for non-premium and 4 calls per CC for premium because it has to get all pinned messages in the channel to be able to find number of pins.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -536,6 +536,7 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("getChannel", c.tmplGetChannel)
 	c.addContextFunc("getThread", c.tmplGetThread)
 	c.addContextFunc("getChannelOrThread", c.tmplGetChannelOrThread)
+	c.addContextFunc("getPinCount", c.tmplGetChannelPinCount)
 	c.addContextFunc("getRole", c.tmplGetRole)
 	c.addContextFunc("addReactions", c.tmplAddReactions)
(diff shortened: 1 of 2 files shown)


jo3-l:
If this requires fetching all pinned messages, why not simply have `getPinnedMessages` and let the user call `len` on the result if they wish? (Exposing such a function would have other uses as well, I would imagine.)

common/templates/context.go
```diff
@@ -536,6 +536,7 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("getChannel", c.tmplGetChannel)
 	c.addContextFunc("getThread", c.tmplGetThread)
 	c.addContextFunc("getChannelOrThread", c.tmplGetChannelOrThread)
+	c.addContextFunc("getPinCount",  c.tmplGetChannelPinCount)
```
jo3-l, line 539:
```suggestion
	c.addContextFunc("getPinCount", c.tmplGetChannelPinCount)
```

LemmeCry:
True, this was discussed for a while on the support server, before everybody felt returning just the length was better, provided no other template gives users access to a bulk of messages.

ashishjh-bst:
Hi @LemmeCry, can you add an example usage of this template in CC?

LemmeCry:
@ashishjh-bst it'll simply be `{{getPinCount nil}}` or any channel ID instead of "nil"

PR #1070, opened 2021-11-19
Update swearwords.go

martinbndr:
Added a few swear words


BlackWolfWoof:
I personally think that adding some swearwords from other languages (for example German) will be a good idea but I do not think this will get accepted because you can simply just add those in your own list on your server and adding swearwords from every language would be quite a challange.

martinbndr:
> I personally think that adding some swearwords from other languages (for example German) will be a good idea but I do not think this will get accepted because you can simply just add those in your own list on your server and adding swearwords from every language would be quite a challange.

A okay yea I see your point. Just had the idea suddenly 😂 Sry then

PR #1077, opened 2021-11-27
FIX: templates > add/delete MessageReactions having nil arguments

mrbentarikau:
This fixes message-reaction functions for templating system if they get nil arguments either as messageID or userID... causing *reflect: call of reflect.Value.Interface on zero Value* soft-panic.


mrbentarikau:
you use those custom commands before this patch and see bot panicking versus after applying those patches.

PR #1097, opened 2021-12-22
reddit/web: check if subreddit is available

l-zeuch:
Request on the support server to check whether
a given subreddit does not exist, and to fail if so.

Thanks to @mrbentarikau for poking at Reddit's API and finding a better endpoint to use, that way this is the only PR required for this functionality.


reddit/reddit.go
```diff
@@ -109,3 +112,43 @@ func MaxFeedForCtx(ctx context.Context) int {
...
 }
+
+func CheckSubreddit(data string) (bool, error) {
+	var redditStruct struct {
+		Data struct {
+			Dist     int `json:"dist"`
+			Children []struct {
+			}
+		} `json:"data"`
+	}
```
jo3-l, line 123: Perhaps name this `redditData`? Generally naming variables with their types (Hungarian notation) is discouraged (e.g., you didn't use `dataString` as the name of the first parameter.)

reddit/reddit.go
```diff
@@ -109,3 +112,43 @@ func MaxFeedForCtx(ctx context.Context) int {
 
 	return GuildMaxFeedsNormal
 }
+
+func CheckSubreddit(data string) (bool, error) {
```
jo3-l, line 116: `name` seems like it describes this parameter more appropriately.
```suggestion
func CheckSubreddit(name string) (bool, error) {
```
...and ditto for usages.

reddit/plugin_web.go
```diff
@@ -142,6 +142,10 @@ func HandleNew(w http.ResponseWriter, r *http.Request) interface{} {
 		Disabled:   false,
 	}
 
+	if ok, _ := CheckSubreddit(newElem.Subreddit); !ok {
```
jo3-l, line 145: Is there a reason `CheckSubreddit` returns `(bool, error)` rather than simply `bool`, given that the only callsite only uses the first return value?

(If you do change `CheckSubreddit` to return a single value, perhaps call it `SubredditExists` or something along those lines to suggest that it returns a boolean.)

reddit/reddit.go
```diff
@@ -109,3 +112,43 @@ func MaxFeedForCtx(ctx context.Context) int {
...
+	queryErr := json.Unmarshal(body, &redditStruct)
+	if queryErr != nil {
+		return false, err
+	}
+
+	if redditStruct.Data.Dist == 0 || len(redditStruct.Data.Children) == 0 {
+		return false, nil
+	}
+
+	return true, nil
```
jo3-l, line 153:
```suggestion
	return redditStruct.Data.Dist > 0 && len(redditStruct.Data.Children) > 0, nil
```

reddit/reddit.go
```diff
@@ -109,3 +112,43 @@ func MaxFeedForCtx(ctx context.Context) int {
...
+		return false, err
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return false, err
+	}
+
+	queryErr := json.Unmarshal(body, &redditStruct)
```
jo3-l, line 144: Is there a reason this doesn't use the name `err`?

jo3-l (commented):
Couple of nits.

l-zeuch:
Addressed the feedback, thanks!

reddit/plugin_web.go
```diff
@@ -142,6 +142,10 @@ func HandleNew(w http.ResponseWriter, r *http.Request) interface{} {
 		Disabled:   false,
 	}
 
+	if ok := CheckSubreddit(newElem.Subreddit); !ok {
+		return templateData.AddAlerts(web.ErrorAlert(fmt.Sprintf("No such subreddit %s", newElem.Subreddit)))
+	}
```
jo3-l, line 147:
```suggestion
	if !CheckSubreddit(newElem.Subreddit) {
		return templateData.AddAlerts(web.ErrorAlert(fmt.Sprintf("No such subreddit %s", newElem.Subreddit)))
	}
```

reddit/reddit.go
```diff
@@ -109,3 +112,39 @@ func MaxFeedForCtx(ctx context.Context) int {
...
+
+	req.Header.Set("User-Agent", UserAgent())
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
jo3-l, line 139:
```suggestion
	body, err := io.ReadAll(resp.Body)
```
`ioutil` is deprecated.

l-zeuch:
As it stands currently, this is still all wonky and based on maybes etc., also reddit does not like bots calling endpoints without proper authorization (i.e. bearer token).
Converting to draft for now, until I come around and doing it properly with token authorization.

l-zeuch:
Closing, I don't have the time nor energy to further work on this. If anyone else is interested, then please submit a patch!

PR #1105, opened 2022-01-07
FIX: customcommands: fix execCC in threads

zaggi-iwnl:
Any execCC calls currently break the command they're used in if said command is executed in a thread. This should hopefully fix that!

image


mrbentarikau:
there's also scheduleUniqueCC to "fix" and next time PR to dev. thnx

zaggi-iwnl:
Thanks for the heads-up! sorry about that, been way too long since I did anything on github

PR #1113, opened 2022-01-16
Added antiphishing protection

ashishjh-bst:
In this PR I've added protection for common scam links 
They can be enabled by a single click in Basic Automoderator, or can be configured via a trigger in AMV2

### DIFF
--- a/antiphishing/antiphishing.go
+++ b/antiphishing/antiphishing.go
@@ -31,12 +31,22 @@ type BitFlowAntiFishResponse struct {
 	} `json:"matches"`
 }
 
+type SinkingYachtsRecentDomainsResponse struct {
+	Type    string   `json:"type"`
+	Domains []string `json:"domains"`
+}
+
 var (
-	hyperphishURL      = "https://api.hyperphish.com/gimme-domains"
-	bitflowAntiFishURL = "https://anti-fish.bitflow.dev/check"
+	phishingDomainApiBaseUrl          = "https://phish.sinking.yachts/v2"
+	getAllPhishingDomainsUrl          = fmt.Sprintf("%s/%s", phishingDomainApiBaseUrl, "all")
+	getRecentPhishingDomainUpdatesUrl = fmt.Sprintf("%s/%s", phishingDomainApiBaseUrl, "recent")
 )
 
-const RedisKeyHyperfishDomains = "hyperfish_domains"
+var (
+	fallbackPhishingUrlCheckAPI = "https://anti-fish.bitflow.dev/check"
+)
+
+const RedisKeyPhishingDomains = "phishing_domains"
 
 var logger = common.GetPluginLogger(&Plugin{})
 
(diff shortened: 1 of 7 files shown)


antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func saveHyperfishDomains() ([]string, error) {
+	domains, err := fetchHyperfishDomains()
+	if err != nil {
+		return nil, err
+	}
+
+	//clear old domains incase there was a false positive
+	redisDelErr := common.RedisPool.Do(radix.Cmd(nil, "DEL", "hyperfish_domains"))
```
jo3-l, line 76: Maybe make `hyperfish_domains` a const, e.g., `RedisKeyHyperfishDomains`, to be consistent with existing code and reduce magic values.

Also:
```suggestion
	err = common.RedisPool.Do(radix.Cmd(nil, "DEL", "hyperfish_domains"))
```
Typically in Go the idiomatic pattern is to shadow the `err` variable as opposed to using fresh names. (Ditto elsewhere.)

antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+	if redisDelErr != nil {
+		return nil, redisDelErr
+	}
+	//save new domains to set
+	args := append([]string{"hyperfish_domains"}, domains...)
+	redisSaveErr := common.RedisPool.Do(radix.Cmd(nil, "SADD", args...))
```
jo3-l, line 82: (Not tested) -- is there any reason this isn't done thusly?
```suggestion
	redisSaveErr := common.RedisPool.Do(radix.Cmd(nil, "SADD", "hyperfish_domains", domains...))
```

antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+	return domains, nil
+}
+
+func saveHyperfishDomains() ([]string, error) {
+	domains, err := fetchHyperfishDomains()
+	if err != nil {
+		return nil, err
+	}
+
+	//clear old domains incase there was a false positive
```
jo3-l, line 75: Style suggestion; really nitpicky:
```suggestion
	// clear old domains incase there was a false positive
```

antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+	i := 0
+	for range input {
+		isPhishingLink, err := queryHyperFish(input[i])
+		if err != nil {
+			return nil, err
+		}
+		if *isPhishingLink {
+			return &input[i], err
+		}
+	}
```
jo3-l, line 163:
```suggestion
	for i, d := input {
		isPhishingLink, err := queryHyperFish(d)
		if err != nil {
			return nil, err
		}
		if *isPhishingLink {
			return &d, err
		}
	}
```
(More idiomatic. Also, presently the code appears incorrect as `i` is always `0` as opposed to iterating through `input`.)

antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+	//save new domains to set
+	args := append([]string{"hyperfish_domains"}, domains...)
+	redisSaveErr := common.RedisPool.Do(radix.Cmd(nil, "SADD", args...))
+	if redisSaveErr != nil {
+		return nil, redisSaveErr
+	}
+	return domains, nil
+}
+
+func queryHyperFish(input string) (*bool, error) {
```
jo3-l, line 89:
```suggestion
func queryHyperFish(input string) (bool, error) {
```
Typically in Go we do not return pointers to primitives when not necessary. In this case it suffices to return the zero value (`false`) upon error.

antiphishing/antiphishing.go
```diff
@@ -0,0 +1,183 @@
...
+	jsonErr := json.Unmarshal(bytes, &bitflowAntifishResponse)
+	if err != nil {
+		logrus.WithError(jsonErr).Error(("[antiphishing] Error parsing JSON from bitflowAntiFish API"))
+		return nil, jsonErr
+	}
+
+	return &bitflowAntifishResponse, nil
+}
+
+func checkPhishingDomains(input []string) (*string, error) {
```
jo3-l, line 153: Ditto here -- `string` as opposed to `*string`.

automod_legacy/assets/automod_legacy.html
```diff
@@ -186,31 +186,13 @@ <h3>Basic automod</h3>
...
-        <div class="checkbox">
-            <label>
-                <input type="checkbox" name="Sites.BuiltinBadSites" {{if .AutomodConfig.Sites.BuiltinBadSites }} checked{{end}} disabled> Enable safe browsing unwanted software site detection
-            </label>
-        </div>
-        <div class="checkbox">
-            <label>
-                <input type="checkbox" name="Sites.BuiltinBadSites" {{if .AutomodConfig.Sites.BuiltinBadSites }} checked{{end}} disabled> Enable safe browsing potentially harmfull application site detection
-            </label>
-        </div> -->
```
jo3-l, line 213: Can I suggest that this commented code be simply removed if it is no longer needed? (Keeping it there confuses future readers of the code -- is it not working, not useful anymore, or meant to be removed completely?) Those that found it useful can always search back in Git history.

jo3-l (commented):
Brief review -- mainly style/Go idioms-related things, in addition to one possible bug (or at least very confusingly-written code.)

automod_legacy/assets/automod_legacy.html
ashishjh-bst: This is already removed in this PR! :D

simoniz0r:
Just gonna throw this out there... using the hyperphish API seems like not a great idea.  It seems like ***anyone*** can ***add and remove*** domains from their list ***without authentication***.  I just tested it myself, and I was able to add and remove `bob.com` without authenticating at all.  That means anyone could add any domain and make it trigger as a phishing domain or anyone could remove all of the domains.

The other API you're using is actually using https://phish.sinking.yachts as a backend which allows you to fetch a list of all of the domains in a similar fashion to how you're using hyperphish if you wanted to replace hyperphish with something a bit more secure.

BlackWolfWoof:
UHHHHHHHHHHHHH

BlackWolfWoof:
@simoniz0r the author of the api is going to add an auth within the next few hours *sweats*

simoniz0r:
Good to hear lol

vcokltfre:
Hi, dev here! The add/remove endpoints are now inaccessible on the current, we're moving to a new version of the API anyway (this implementation shouldn't notice any changes as the downstream, `sinking.yachts` will handle it for you) which has token-based auth, so this will cease to be an issue at all

PR #1127, opened 2022-02-21
soundboard: actually switch voice channels

jo3-l:
While I have tested this change fairly thoroughly, the soundboard code is quite complicated due to it being concurrent, with many possible paths through the code. I am not fully certain that this patch does not introduce new race conditions as I am unfamiliar with the surrounding code. Further testing under different scenarios (i.e., multiple servers using soundboard) would be appreciated.

---

When a user runs soundboard in a voice channel — but not the one the bot is in — it is expected that the bot switches voice channels when the sound played comes up. This does not occur currently -- the bot does not change channels with the user, though the code is written with the intention to do so.

<details><summary>Reproduction steps</summary>

1. Have two voice channels. We will refer to them as VC 1 and VC 2 for convenience.
2. Join VC 1 and run `-sb sound_1`.
3. Wait for the bot to join and play `sound_1`.
4. Move to VC 2 and run `-sb sound_1` again.
5. Observe that while the bot does play the sound, it stays in VC 1 instead of moving to VC 2.

</details>

The reason for this behavior lies in the `ChangeChannel` method being used. More specifically, it has several critical bugs causing it to not actually change channels, such not waiting for the new voice connection to become ready, the channel ID being switched with the guild ID, and the channel ID not being the one requested.

Though these issues with `ChangeChannel` could be fixed, this PR takes a different approach, opting instead to use the `ChannelVoiceJoin` method recommended by the documentation of `ChangeChannel`.

### DIFF
--- a/soundboard/player.go
+++ b/soundboard/player.go
@@ -115,13 +115,8 @@ func (p *Player) Run() {
 		p.timeLastPlay = time.Now()
 		playersmu.L.Unlock()
 
-		// do a changechannel if we are in another channel and also voice is connected
-		if changeChannel && p.vc != nil {
-			p.vc.ChangeChannel(item.ChannelID, false, true)
-		}
-
 		var err error
-		p.vc, err = playSound(p, p.vc, bot.ShardManager.SessionForGuild(p.GuildID), item)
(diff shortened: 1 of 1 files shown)


jo3-l:
Some oddities came up -- with others not being to reproduce the issue at all and strange errors being logged after this change -- so marking this as draft for now.

PR #1130, opened 2022-03-03
Added validations for Autodelete trigger and response deletion delay

ankurp-bst:
**YAGPDB Control Panel > Core > Command Settings**
User was able to add and save negative and very large values for "Autodelete Trigger after" and "Autodelete Response after".
This pull request adds range validation for these fields in both Channel and Command Override.

### DIFF
--- a/commands/assets/commands.html
+++ b/commands/assets/commands.html
@@ -238,7 +238,7 @@
                             {{if .Override.AutodeleteTrigger}}checked{{end}}>
                     </span>
                 </span>
-                <input type="number" min="0" max="3600" class="form-control" placeholder="Seconds..."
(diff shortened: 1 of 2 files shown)


l-zeuch:
Is there any reason this is such an arbitrarily low upper bound?

For example servers utilising the `poll` command en masse would definitely not benefit from such a low limit.

I for myself use polls often enough and I autodelete them after a day, sometimes after a week. This patch will break exactly that.

ankurp-bst:
> Is there any reason this is such an arbitrarily low upper bound?
> 
> For example servers utilising the `poll` command en masse would definitely not benefit from such a low limit.
> 
> I for myself use polls often enough and I autodelete them after a day, sometimes after a week. This patch will break exactly that.

Discussed with Des. Updating the delay to 1 month

PR #1131, opened 2022-03-07
Fix calc dependency to include a correct implementation of acot

ei14:
One dependency is changed: from alfredxing/calc to ei14/calc.

Compared to alfredxing/calc, ei14/calc is different by 1 line.

Prior to this fix, `-c acot(0)` returns 90 despite the fact that every other trig function is in radians. Furthermore, `-c acot(1e100)` returns 90 - pi/2 instead of a near-zero value as it should. These discrepancies are caused by `acot` being calculated as `90 - atan` instead of `pi/2 - atan`.

ei14/calc has a correct implementation of `acot`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/PuerkitoBio/goquery v1.5.0
 	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
-	github.com/alfredxing/calc v0.0.0-20180827002445-77daf576f976
+	github.com/ei14/calc v0.0.0-20220307064635-62ca623f5da1
 	github.com/bwmarrin/snowflake v0.3.0
 	github.com/dghubble/oauth1 v0.6.0
 	github.com/didip/tollbooth v4.0.2+incompatible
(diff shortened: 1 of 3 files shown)


ashishjh-bst:
You should rather raise a PR to github.com/alfredxing/calc for this.

l-zeuch:
[alfredxing/calc](https://github.com/alfredxing/calc) is archived by the author and thus does not accept any pull requests and the like.

Updating the dependency to a maintained fork is the ideal way of getting correct behaviour.

ashishjh-bst:
Oh okay, I'll in that case switch to this fork then. Thanks @l-zeuch

jo3-l:
Unless I'm mistaken, this PR breaks the build. I'm getting several errors along the lines of `no required module provides package github.com/alfredxing/calc/constants` with this commit. It seems like this issue was fixed in https://github.com/ei14/calc/commit/adbe43bdf8019134486df43b1f2a807dd63a3e34, but the version in go.mod points to the previous commit.

ashishjh-bst:
> Unless I'm mistaken, this PR breaks the build; I'm getting several errors along the lines of `no required module provides package github.com/alfredxing/calc/constants` with this commit. It seems like this issue was fixed in [ei14/calc@adbe43b](https://github.com/ei14/calc/commit/adbe43bdf8019134486df43b1f2a807dd63a3e34), but the version in go.mod points to the previous commit.

@jo3-l  fixed in #1168

SeniorMars:
Thank you for this fix. This has made the bot way better.

PR #1138, opened 2022-03-15
Update middleware validator to handle only min value, add min value validations in various forms

ankurp-bst:
1. Updated the readMinMax function to handle empty max value so that only min value can be provided (for example: a field that requires positive number, hence min=0 and no max value). For this field the validation will be described as:
```
type Example struct {
    PositiveField int `valid:"0,"`
}
```
Here only min value (0) will be validated, and if negative value is entered, it will show an error:
```
Positive Field: should be at least 0
```

2. Several forms on the control panel didn't have proper validations on the backend, this PR adds those validations.

### DIFF
--- a/web/validation.go
+++ b/web/validation.go
@@ -114,8 +114,8 @@ func ValidateForm(guild *dstate.GuildSet, tmpl TemplateData, form interface{}) b
 		// Perform validation based on value type
 		switch cv := vField.Interface().(type) {
 		case int:
-			min, max := readMinMax(validationTag)
-			err = ValidateIntMinMaxField(int64(cv), int64(min), int64(max))
+			min, max, onlyMin := readMinMax(validationTag)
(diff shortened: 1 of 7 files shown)


web/validation.go
```diff
@@ -219,12 +219,15 @@ func ValidateForm(guild *dstate.GuildSet, tmpl TemplateData, form interface{}) b
...
+func readMinMax(valid *ValidationTag) (float64, float64, bool) {
+	onlyMin := false
 	min, _ := valid.Float(0)
 	max, _ := valid.Float(1)
+	if valid.values[1] == "" {
+		onlyMin = true
+	}
 
-	return min, max
+	return min, max, onlyMin
```
jo3-l, line 230: It's possible to simplify this thusly: `return min, max, valid.values[1] == ""`. Whether this is a good idea is debatable, but figured I'd throw it out there.

PR #1165, opened 2022-04-07
Update autorole to add Membership screening required toggle, and add its handling

ankurp-bst:
Members were being assigned roles from Autorole even if their membership screening was pending.
This PR adds a toggle to make membership screening required before assigning autrole to that member.

### DIFF
--- a/common/scheduledevents2/backgroundworker.go
+++ b/common/scheduledevents2/backgroundworker.go
@@ -104,7 +104,7 @@ func runFlushNewEvents() error {
 // flushEventToRedis flushes an event to redis, this is done as a performance improvement as the postgres db is only queried as often
 func flushEventToRedis(c radix.Client, evt *models.ScheduledEvent) error {
 	v := fmt.Sprintf("%d:%d", evt.ID, evt.GuildID)
-	err := c.Do(radix.Cmd(nil, "ZADD", "scheduled_events_soon", strconv.FormatInt(evt.TriggersAt.UTC().Unix(), 10), v))
+	err := c.Do(radix.Cmd(nil, "ZADD", "scheduled_events_soon", strconv.FormatInt(evt.TriggersAt.UTC().UnixMicro(), 10), v))
 	if err != nil {
 		return err
 	}
--- a/common/scheduledevents2/scheduledevents.go
+++ b/common/scheduledevents2/scheduledevents.go
@@ -160,7 +160,7 @@ func (se *ScheduledEvents) check() {
 	defer se.currentlyProcessingMU.Unlock()
 
 	var pairs []string
-	err := common.RedisPool.Do(radix.FlatCmd(&pairs, "ZRANGEBYSCORE", "scheduled_events_soon", "-inf", time.Now().UTC().Unix()))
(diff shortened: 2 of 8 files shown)


common/scheduledevents2/backgroundworker.go
```diff
@@ -104,7 +104,7 @@ func runFlushNewEvents() error {
 // flushEventToRedis flushes an event to redis, this is done as a performance improvement as the postgres db is only queried as often
 func flushEventToRedis(c radix.Client, evt *models.ScheduledEvent) error {
 	v := fmt.Sprintf("%d:%d", evt.ID, evt.GuildID)
-	err := c.Do(radix.Cmd(nil, "ZADD", "scheduled_events_soon", strconv.FormatInt(evt.TriggersAt.UTC().Unix(), 10), v))
+	err := c.Do(radix.Cmd(nil, "ZADD", "scheduled_events_soon", strconv.FormatInt(evt.TriggersAt.UTC().UnixMicro(), 10), v))
```
jo3-l, line 107: Out of curiosity, what's up with this change?

common/scheduledevents2/scheduledevents.go
```diff
@@ -160,7 +160,7 @@ func (se *ScheduledEvents) check() {
 	defer se.currentlyProcessingMU.Unlock()
 
 	var pairs []string
-	err := common.RedisPool.Do(radix.FlatCmd(&pairs, "ZRANGEBYSCORE", "scheduled_events_soon", "-inf", time.Now().UTC().Unix()))
+	err := common.RedisPool.Do(radix.FlatCmd(&pairs, "ZRANGEBYSCORE", "scheduled_events_soon", "-inf", time.Now().UTC().UnixMicro()))
```
jo3-l, line 163: Ditto.

common/scheduledevents2/backgroundworker.go
ankurp-bst: Since, `.Unix()` does not consider microseconds, the events were being triggered a few microseconds before they were supposed to. This change allows the events to trigger more precisely.
(The reason for updating this is that some events were being re-scheduled, because there was still some microseconds left before they were actually supposed to get triggered)

jo3-l: Interesting! Seems like a reasonable change. I was thinking about what occurs when there are entries in `scheduled_events_soon` with second-precision scores (i.e., added before this change) in addition to microsecond-precision scores (i.e., after this change) and whether this would lead to issues due to the different format, but it turns out the existing code should handle this just fine. Nonetheless, perhaps testing such a scenario might be prudent (to make sure events added before still work), if you haven't done so already.

mrbentarikau:
YAGPDB is in modules and in circleCI conf still using go v1.16 and that version does not have UnixMicro as time method

common/scheduledevents2/backgroundworker.go
ankurp-bst: Just after deployment, the entries in `scheduled_events_soon` will be scheduled (because their score will be less than the time generated by UnixMicro), but it will be a one time thing and yag can handle it elegantly.

PR #1178, opened 2022-04-15
Add Alias "yeet" to stdcommand "throw"

savage4618:
Suggestion ID : 416362327789666307
Suggestion Message ID: 741166631312556103


savage4618:
Des said "yeet" and "throw" have different definitions.

PR #1187, opened 2022-04-23
add dadjoke command.

savage4618:
This command uses an API (icanhazdadjoke) to pull in a random dadjoke. 

I had contacted the original creator of this code to get it re-done. They had confirmed to me it still worked and they used it on their self-host. I have tested the code on my own self-host now.

### DIFF
--- a/stdcommands/dadjoke/dadjoke.go
+++ b/stdcommands/dadjoke/dadjoke.go
@@ -1,0 +1,56 @@
+package dadjoke
+
+import (
+	"io/ioutil"
+	"net/http"
+	"encoding/json"
+	"github.com/jonas747/dcmd/v4"
+	"github.com/botlabs-gg/yagpdb/commands"
+)
+
+//Create the struct that we will serialize the api respone into.
+type Joke struct {
+	ID string `json:"id"`
+	JOKE string `json:"joke"`
+	STATUS int `json:"status"`
+}
+
+var Command = &commands.YAGCommand{
+	Cooldown:    5,
+	CmdCategory: commands.CategoryFun,
+	Name:        "DadJoke",
+	Description: "Generates a dad joke using the API from icanhazdadjoke.",
+	DefaultEnabled: 		true,
+	SlashCommandEnabled: 	true,
+	RunFunc: func(data *dcmd.Data) (interface{}, error) {
+		//Define the request and website we will navigate to. 
+		req, err := http.NewRequest("GET", "https://icanhazdadjoke.com", nil)
+		if err != nil {
+			return nil , err
+		}
+		//Set the headers that will be sent to the API to determine the response.
+		req.Header.Set("Accept", "application/json")
+
+		client := &http.Client{}
+		apiResp, err := client.Do(req)
+		if err != nil {
+			return nil , err
+		}
+		//Once the rest of the function is done close our connection the API.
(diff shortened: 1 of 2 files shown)


mrbentarikau (changes requested):
Header,Set takes only two string arguments, this will error out on compile

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		req, err := http.NewRequest("GET", "https://icanhazdadjoke.com", nil)
+		if err != nil {
+			return nil , err
+		}
+		//Set the headers that will be sent to the API to determine the response.
+		req.Header.Set("Accept", "application/json")
+		req.Header.Add("User-Agent", "YAGPDB.xyz (https://github.com/botlabs-gg/yagpdb)")
+
+		client := &http.Client{}
+		apiResp, err := client.Do(req)
```
jo3-l, line 36:
```suggestion
		apiResp, err := http.DefaultClient.Do(req)
```

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return nil , err
+		}
+		//Once the rest of the function is done close our connection the API.
+		defer apiResp.Body.Close()
+		//Read the API response.
+		bytes, err := ioutil.ReadAll(apiResp.Body)
```
jo3-l, line 43: Package ioutil is deprecated. Use `io.ReadAll` instead.

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return nil , err
+		}
+		//Create our struct and unmarshal the content into it.
+		joke1 := Joke{}
```
jo3-l, line 48:
```suggestion
		var joke Joke
```

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		jsonErr := json.Unmarshal(bytes,&joke1)
```
jo3-l, line 49:
```suggestion
		err = json.Unmarshal(bytes,&joke1)
```

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		if jsonErr != nil {
+			return nil , jsonErr
+		}
```
jo3-l, line 52:
```suggestion
		if err != nil {
			return nil, err
		}
```

stdcommands/dadjoke/dadjoke.go
```diff
@@ -0,0 +1,57 @@
...
+		//Return the joke - the other pieces are unneeded and ignored. 
+		resp := joke1.JOKE
+		return resp , nil
```
jo3-l, line 55:
```suggestion
		return joke.Joke, nil
```

jo3-l (commented):
Some overarching suggestions:

- Run `gofmt` on this file.
- Many of the comments seem superfluous; consider removing.

savage4618:
Thank you Joe for reviewing. I fixed all suggested lines and tested with success.

PR #1196, opened 2022-05-05
reputation: default to command author if no arguments are provided to the replog command

lemm-e:
This PR allows using the `replog` command without having to provide the user argument by making it optional, in which case, it defaults to the user using the command.


lemm-e:
Marking as draft because `replog 10`, for example, will not show page 10 of the user invoking the command.

jo3-l:
Now that #1218 has been merged (which implements this change a different way), this can be closed.

PR #1207, opened 2022-05-08
all: provide more suggestions on invalid user input

jo3-l:
Supercedes #901.

This PR implements a fuzzy matching package using the Jaro-Winkler similarity metric and changes a couple of bot commands to provide suggestions on invalid user input using the aforementioned package. Studies have shown that [Jaro-Winkler strikes a good balance between performance and accuracy](https://www.cs.cmu.edu/afs/cs/Web/People/wcohen/postscript/kdd-2003-match-ws.pdf).

A complete list of commands that have been changed follows:

- `howlongtobeat`: now uses Jaro-Winkler instead of Levenshtein distance. Some quick tests indicate that the results are not affected much.
- `rolemenu create`: now provides suggestions for role group names when an invalid one is provided.
- `cah create`: now provides suggestions for card packs when an invalid one is provided.


mrbentarikau:
Maybe call the lib "jarowinkler", not just "fuzzy".

jo3-l:
Interesting idea. I'm not completely opposed to it, but I think `fuzzy.Select` reads better than `jarowinkler.Select`, though `fuzzy.Similarity` is definitely worse, so it's a bit of a tradeoff.

I will consider it; in the meantime, other thoughts and opinions on the naming of the package are most welcome.

mrbentarikau:
to me fuzzy does not read better, because it does not tell you anything

jo3-l:
The idea I had in mind was that `fuzzy.Select` implies fuzzy selection.

l-zeuch:
Personally I don't think `jarowinkler` is suitable either, what if more matching algorithms are implemented? That'll be inaccurate then.

To me `fuzzy` seems... weird to pop up in a codebase; though I lack the words to suggest a better alternative :-)

mrbentarikau:
> The idea I had in mind was that `fuzzy.Select` implies fuzzy selection.

maybe jarowinkler.FuzzySelect then

jo3-l:
Alternately, we could rename `Similarity` to `JaroWinklerSimilarity`, which then leads to `fuzzy.Select` (fuzzy selection) and `fuzzy.JaroWinklerSimilarity` (self-explanatory.) `jarowinkler.FuzzySelect` and `jarowinkler.Similarity` are also nice though.

l-zeuch:
> Alternately, we could rename `Similarity` to `JaroWinklerSimilarity`, which then leads to `fuzzy.Select` (fuzzy selection) and `fuzzy.JaroWinklerSimilarity` (self-explanatory.) `jarowinkler.FuzzySelect` and `jarowinkler.Similarity` are also nice though.

I like the first part, though I still find `fuzzy` a rather weird word to use.

mrbentarikau:
sounds better already, but definitely something different than "lib/fuzzy"

l-zeuch:
Maybe something like `matching`? That conveys the message and is in line with the intention behind `fuzzy`.

mrbentarikau:
hehe, and on review side... works well, and HLTB even gives more "accurate" results

jo3-l:
> I like the first part, though I still find `fuzzy` a rather weird word to use.

"Fuzzy string matching" is a well-established term; see, e.g., [approximate string matching](https://en.wikipedia.org/wiki/Approximate_string_matching), [fuzzy logic](https://en.wikipedia.org/wiki/Fuzzy_logic), the popular [thefuzz](https://github.com/seatgeek/thefuzz) Python library, [O'Reilly's _Taming Text_](https://www.oreilly.com/library/view/taming-text-how/9781933988382/kindle_split_012.html), and [academia](https://scholar.google.ca/scholar?hl=en&as_sdt=0%2C5&q=fuzzy+string+matching&btnG=).

jo3-l:
Renamed package to `jarowinkler` per further discussion on Discord.

Dhmee:
StringMetric
StringMatching

PR #1211, opened 2022-05-08
customcommands: improve editor ui

jo3-l:
-  For command triggers, display the prefix to the left of the actual trigger. The prefix will be hidden for all other trigger types. Inspiration for this change comes from a fairly popular suggestion (18 upvotes, 1 downvote) on the server from a while back.
- Show the last run time and the trigger length.
- Add <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd> as a shortcut to save the command.

Before:
Screenshot of editor before changes

After:
Screenshot of editor after changes


jo3-l:
Mr. T brought up a couple of other improvements to the CC editor during review that I'd like to add in here as well, so will draft this for now.

PR #1214, opened 2022-05-09
Update User arg in dcmd to accept ID, use this in all rep related commands

ankurp-bst:
Updated UserArg in dcmd to accept the user id.
Use dcmd.User in all places in reputation related commands for consistency.

### DIFF
--- a/customcommands/tmplextensions.go
+++ b/customcommands/tmplextensions.go
@@ -73,7 +73,7 @@ func tmplCArg(typ string, name string, opts ...interface{}) (*dcmd.ArgDef, error
 	case "string":
 		def.Type = dcmd.String
 	case "user":
-		def.Type = dcmd.UserReqMention
+		def.Type = dcmd.User
 	case "userid":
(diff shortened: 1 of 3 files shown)


jo3-l:
I haven't looked at this very closely, but doesn't this patch render modifying the reputation of users no longer in the server impossible?

ankurp-bst:
> I haven't looked at this very closely, but doesn't this patch render modifying the reputation of users no longer in the server impossible?

Updated the code to allow modifying the rep of any user no longer in the server, only if the user has a record in rep logs.
`replog` and `delrep` can also be used for users no longer in server.

jo3-l:
While rebasing another PR to accommodate this change, I realized that `carg "user"` in custom commands now accepts searches by username. Is this intended? Particularly, has the impact of performing searches by username been considered, provided that it iterates over all members (and is thus fairly computationally expensive)? (Similarly, is changing all user arguments to search by username an intended change?)

PR #1218, opened 2022-05-10
dcmd: refactor combo selection algorithm

jo3-l:
Presently, `-clean user_id n` (e.g., `-clean 204255221017214977 1`) fails with 
> Invalid arguments provided: Num is too big (has to be within 1 - 100)

The issue is that the user ID argument is erroneously interpreted as the number of arguments to be deleted. That is, the argument combo with the user argument first and the number second is selected, while it should be the other way around.

This PR fixes the issue with the `clean` command and additionally supercedes #1196 through using argument combos in a way that was not previously possible.

## Implementation details and discussion

The underlying cause of the issue is that the integer argument type does not take its range into account when matching (the float argument type has a similar issue.) In other words, given an integer type bounded to the range `[1, 10]`, the value `11` would match but fail during parsing. The obvious fix then seems to be to make matching for integer types take range into account, so `11` does not match an integer type with range `[1, 10]`.

Unfortunately, though this fix addresses the original issue with the clean command, it introduces a regression; when the number of messages to delete is greater than 100, e.g., `-clean 204255221017214977 101`, the cryptic error
> Invalid arguments provided: No matching combo found

is produced, replacing the original, more helpful message
>  Invalid arguments provided: Num is too big (has to be within 1 - 100)

In order to avoid this regression, this PR takes a different, more involved approach. Instead of having input matching returning a binary result (either the input matches or it doesn't), it may now return three types of results: the input can be incompatible with the type; it can have poor compatibility with the type; and the input can have good compatibility with the type. (Some examples: "abc" is incompatible with an integer type; 10 has poor compatibility with an integer type with range `[0, 5]`; 100 has good compatibility with an unbounded integer type.) 

Accordingly, the argument combo selection algorithm is modified such that it prefers argument combos with better overall compatibility as opposed to just length. Argument combos with incompatible components will be rejected, but combos having components with poor compatibility will be accepted if there is no better alternative. This retains the helpful error message in cases like out-of-range integers.

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -21,19 +21,22 @@ type DurationArg struct {
 
 var _ dcmd.ArgType = (*DurationArg)(nil)
 
-func (d *DurationArg) Matches(def *dcmd.ArgDef, part string) bool {
+func (d *DurationArg) CheckCompatibility(def *dcmd.ArgDef, part string) dcmd.CompatibilityResult {
 	if len(part) < 1 {
-		return false
+		return dcmd.Incompatible
 	}
 
 	// We "need" the first character to be a number
 	r, _ := utf8.DecodeRuneInString(part)
 	if !unicode.IsNumber(r) {
-		return false
+		return dcmd.Incompatible
 	}
 
 	_, err := common.ParseDuration(part)
-	return err == nil
+	if err != nil {
+		return dcmd.Incompatible
+	}
+	return dcmd.CompatibilityGood
 }
 
 func (d *DurationArg) ParseFromMessage(def *dcmd.ArgDef, part string, data *dcmd.Data) (interface{}, error) {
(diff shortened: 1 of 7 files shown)


jo3-l:
This PR has been updated to remove username/nickname support from user arguments. Without doing this, it is impossible to differentiate between a user argument and any other argument (since nicknames/usernames can essentially be any string modulo length restrictions.) Though this is technically a breaking change, I speculate that it will have little impact as the username/nickname search function is typically foregone by users in favour of mentions/IDs. 

(Note that if username/nickname support is still desirable in certain cases, `dcmd.AdvUserArg` may be used.)

jo3-l:
Going to tentatively mark this as ready for review -- I tested it more today and did not find any issues/regressions.

PR #1222, opened 2022-05-13
Add count and limit on CC control panel, prevent creating new CC if limit is exhausted

ankurp-bst:
This PR adds a text message before the `Create a new Custom Command` button to show how many commands are created, and disables the button if the limit is exhausted.
See the screenshots below:
Screenshot 2022-05-13 at 3 45 35 PM
Screenshot 2022-05-13 at 3 46 14 PM

### DIFF
--- a/customcommands/assets/customcommands.html
+++ b/customcommands/assets/customcommands.html
@@ -62,7 +62,7 @@
                     <form class="form-horizontal" method="post"
                         action="/manage/{{.ActiveGuild.ID}}/customcommands/groups/{{.CurrentCommandGroup.ID}}/update"
                         data-async-form>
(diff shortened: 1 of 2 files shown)


mrbentarikau:
This one here does not take account if all guilds are set to premium through /admin/config or via Redis directly.

ankurp-bst:
> This one here does not take account if all guilds are set to premium through /admin/config or via Redis directly.

This can be done by using `premium.IsGuildPremium()` function as the condition for checking whether the guild is premium or not. This handles the case when all guilds are set to premium through the env variable, but I saw some issues when setting premium through the /admin/config. So I'll update this once the issue is figured out.

PR #1234, opened 2022-05-22
custom commands main view reactions clearer

mrbentarikau:
This PR makes it clearer UI-wise what action triggers the trigger for reactions.
"reaction added + removed" needs discussion, maybe "or" is better. Although CC-edit view uses plus.

image


ashishjh-bst:
or would definitely sounds better.

mrbentarikau:
updated to or

PR #1246, opened 2022-05-25
Updated thanks reputation regex to include some more multilingual exp…

redja150:
…ressions

### DIFF
--- a/reputation/plugin_bot.go
+++ b/reputation/plugin_bot.go
@@ -32,7 +32,7 @@ func (p *Plugin) BotInit() {
 	eventsystem.AddHandlerAsyncLastLegacy(p, handleMessageCreate, eventsystem.EventMessageCreate)
 }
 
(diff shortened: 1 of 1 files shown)


lemm-e:
Personally against this PR because it creates an expectation and reason for others to request adding the thanks equivalent in the language they speak. If the regex can be changed to add French or Spanish, why wouldn't Portugese be added? What about other languages users might use on their server? Such things can already be achieved via Custom commands and that, according to me, is the best option available to make any such additions. The regex should only include English and its variations, but it won't be possible for us to remove something in order to maintain backward compatibility. An alternative could be to have an option to set a custom thanks regex per-server under the reputation settings on the bot panel, but as it stands right now, custom commands should be the way to go with what's available.

ashishjh-bst:
I agree with lemme on this, instead of adding keywords to the regex, we'd be better off with a dedicated regex per server and letting them configure it, as @mrbentarikau  had suggested.

ashishjh-bst:
Closing this, as this is not the approach we intend to take.

PR #1258, opened 2022-06-07
Add interaction related structs and functions, use buttons in paginated commands

ankurp-bst:
This PR
1. Adds structs and functions required for interactions.
2. Updates paginated messages to use interaction buttons instead of reactions.


mrbentarikau:
maybe rename /lib/discordgo/utils.go to util.go
no need to have it in plural and keep it consistent with other such files all over the code.

PR #1277, opened 2022-06-20
follow up to #903, removed block to check for online shards with less events

ashishjh-bst:


mrbentarikau:
problem with self hosts is that they are as offline shards by default and this yellow error is always there
image

ashishjh-bst:
> problem with self hosts is that they are as offline shards by default and this yellow error is always there image

That doesn't show up on my selfhost

Screenshot 2022-06-20 at 6 17 34 PM

mrbentarikau:
then it depends probably how one launches the bot and the system
happens here with the same dev branch and FreeBSD
`> ./yagpdb -all -backgroundworkers`

does not happen on linux

mrbentarikau:
all good... the Redis on my end had leftover dshardorchestrator keys and that made it like the screenshot has.

PR #1285, opened 2022-06-27
Fix empty name while saving form issue in amv2

ankurp-bst:


mrbentarikau:
after this "update" whole web-page jumps to the top after pressing normal-save button, not the overlay save

PR #1295, opened 2022-07-01
autorole: add notice about Discord's protections

l-zeuch:
Seeing as Discord's protective means like Verification Levels and Rule
Screening do not apply to members with a role present, adding a role
automatically on join bypasses these means.

This commit adds a notice to the autorole page pointing this exact thing
out.

### DIFF
--- a/autorole/assets/autorole.html
+++ b/autorole/assets/autorole.html
@@ -42,6 +42,12 @@
                             </select>
                         </div>
                         <div class="form-group">
+                            <h4>Notice:</h4>
+                            <p>
+                            Automatically adding any role on join bypasses Discord's protections such as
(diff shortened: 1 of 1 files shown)


autorole/assets/autorole.html
```diff
@@ -42,6 +42,12 @@ <h2>Autorole</h2>
                             </select>
                         </div>
                         <div class="form-group">
+                            <h4>Notice:</h4>
+                            <p>
+                            Automatically adding any role on join bypasses Discord's protections such as
+                            <a href="https://support.discord.com/hc/en-us/articles/216679607-What-are-Verification-levels-"><b>Verification Levels</b></a>
```
ashishjh-bst, line 48: add target="_blank" to the anchor tags to make sure they open in a new window.

l-zeuch: Done in 1b54652, thanks for the review!

PR #1306, opened 2022-08-02
Added functions to reply to a message using custom command

ankurp-bst:
Added key `reply` in `complexMessage` which can be used to reply to a message by providing its message id.
Example:
`{{ sendMessage nil (complexMessage "content" "This is a reply" "reply" .Message.ID) }}`

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -259,8 +259,12 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 			// Cut the filename to a reasonable length if it's too long
 			filename = common.CutStringShort(ToString(val), 64)
 		case "reply":
+			msgID := ToInt64(val)
+			if msgID <= 0 {
+				return nil, errors.New(fmt.Sprintf("invalid message id '%s' provided to reply.", ToString(val)))
+			}
 			msg.Reference = &discordgo.MessageReference{
-				MessageID: val.(int64),
+				MessageID: msgID,
 			}
 		default:
 			return nil, errors.New(`invalid key "` + key + `" passed to send message builder`)
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -258,6 +258,10 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 		case "filename":
 			// Cut the filename to a reasonable length if it's too long
 			filename = common.CutStringShort(ToString(val), 64)
+		case "reply":
+			msg.Reference = &discordgo.MessageReference{
+				MessageID: val.(int64),
+			}
 		default:
 			return nil, errors.New(`invalid key "` + key + `" passed to send message builder`)
 		}
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -565,10 +565,6 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("editMessageNoEscape", c.tmplEditMessage(false))
 	c.addContextFunc("pinMessage", c.tmplPinMessage(false))
 	c.addContextFunc("unpinMessage", c.tmplPinMessage(true))
-	c.addContextFunc("sendMessageReply", c.tmplSendMessageReply(true, false))
-	c.addContextFunc("sendMessageReplyRetID", c.tmplSendMessageReply(true, true))
-	c.addContextFunc("sendMessageReplyNoEscape", c.tmplSendMessageReply(false, false))
-	c.addContextFunc("sendMessageReplyNoEscapeRetID", c.tmplSendMessageReply(false, true))
 
 	// Mentions
 	c.addContextFunc("mentionEveryone", c.tmplMentionEveryone)
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -565,6 +565,10 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("editMessageNoEscape", c.tmplEditMessage(false))
 	c.addContextFunc("pinMessage", c.tmplPinMessage(false))
 	c.addContextFunc("unpinMessage", c.tmplPinMessage(true))
+	c.addContextFunc("sendMessageReply", c.tmplSendMessageReply(true, false))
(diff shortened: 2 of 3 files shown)


mrbentarikau:
why not use complexMessage and add "reply" field there, you already have sendMessage function, why duplicate?

ankurp-bst:
> why not use complexMessage and add "reply" field there, you already have sendMessage function, why duplicate?

Could not use `sendMessage` because to reply to a message, we need to provide ID of the original message as the one of the arguments which was not possible with current sendMessage function.
Will try if I can use the `complexMessage` one. Thanks for suggestion.

mrbentarikau:
yes you could do then
`{{ sendMessage nil (complexMessage "content" "DZ likes cookies" "reply" .Message.ID) }}`

Pedro-Pessoa:
I agree with mrben here. I don't think a new func is a good solution. Adding it to complexMessage looks fine.

You could also make it an optional argument to sendMessage. But I don't know if it is a good solution or not.

```go
func (c *Context) tmplSendMessagee(filterSpecialMentions bool, returnID bool) func(channel interface{}, msg interface{}, replyTo ...interface{}) interface{} {}
```

l-zeuch:
I prefer seeing it being added to `complexMessage`, it is sensible there.

Adding more functions just with slightly different behaviour is in my eyes a bit uncool.

common/templates/general.go
```diff
@@ -258,6 +258,10 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 		case "filename":
 			// Cut the filename to a reasonable length if it's too long
 			filename = common.CutStringShort(ToString(val), 64)
+		case "reply":
+			msg.Reference = &discordgo.MessageReference{
+				MessageID: val.(int64),
```
l-zeuch, line 263:
```suggestion
				MessageID: ToInt64(val),
```

Consider this change instead, otherwise it will fail with a bare int, like in this code:

```go
{{/* any valid message ID, hardcoded */}}
{{ $id := 1004375006253424680 }}
{{ sendMessage nil ( complexMessage "content" "hello" "reply" $id ) }}
```

Without this suggested change, users will have to cast `toInt64` themselves, which is not immediately obvious.

Pedro-Pessoa, line 263: It is easy to cause a panic with this implementation. And the error might look weird to the end user: `error calling complexMessage: interface conversion: interface {} is T, not int64`, T being the type provided.

What if we handle those errors?
```go
case "reply":
	mID := ToInt64(val)
	if mID <= 0 {
		return nil, errors.New("invalid message id provided to reply: " + ToString(val))
	}
	msg.Reference = &discordgo.MessageReference{
		MessageID: mID,
	}
```
This way the error to the user would be: `error calling complexMessage: invalid message id provided to reply: V`, v being the value of whatever they provided. And this is not a panic as well.

l-zeuch (changes requested):
Looks OK, just one thing:

l-zeuch:
0a1c839 fixes it well.

One small question: Discord provides a `fail_if_not_exists` field to their `referenced_message` object, as listed here: https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure

What are your thoughts on implementing some sort of `force` attribute, that will send the message either way? As of now, it fails silently if the message is not found. That of course would mean we'd have to invert the boolean value from a `force:true` to a `fail_if_not_exists:false`, but that should be relatively simple.

If there's an alternative way you're thinking of, do let us know!

----
P.S.: I might just be misinterpreting Discord's documentation a little bit, as `referenced_message` is used for multiple things. I could not make test patches to work as described above.

Pedro-Pessoa:
I don't think this is quite related to this PR. Implementing that field wouldn't change anything with the current implementation of `tmplSendMessage`.
The default value for `fail_if_not_exists` is `true`, so it should be erroring out already. 

The thing here is that `tmplSendMessage` fails silently if an error occurs while making the request. We'd need to make it return an error as well.

l-zeuch:
> I don't think this is quite related to this PR.

It was a question to handle the case of "referenced message does not exist".
 
> The thing here is that `tmplSendMessage` fails silently if an error occurs while making the request. We'd need to make it return an error as well.

If we can make it return an error, then I'm fine with that. Users can `catch` that at-will just fine.

Pedro-Pessoa:
+1 from me.
Since we have try catch, I think that all context_funcs that might return an error, should return an error.

Pedro-Pessoa:
> that's out of scope of this PR

That was my point, adding the `force` attribute to this PR wouldn't make a change.
We could maybe implement it already, but users should be aware that it is not yet working.

> we have to consider how much of a breaking change this would be

I think @ashishjh-bst is in favor: https://discord.com/channels/166207328570441728/920615133993107467/956301200171622462

common/templates/general.go
```diff
@@ -258,6 +258,14 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
 		case "filename":
 			// Cut the filename to a reasonable length if it's too long
 			filename = common.CutStringShort(ToString(val), 64)
+		case "reply":
+			msgID := ToInt64(val)
+			if msgID <= 0 {
+				return nil, errors.New(fmt.Sprintf("invalid message id '%s' provided to reply.", ToString(val)))
+			}
```
jo3-l, line 265: This is not common practice within template functions; typically, we cast and let it error later. (Additionally, if we ever make `sendMessage` return API errors as is being suggested, this check will be redundant.) At any rate, if you would prefer to keep this, I would suggest writing it thusly:
```suggestion
			if msgID <= 0 {
				return nil, fmt.Errorf("invalid message reference ID %d", val)
			}
```

jo3-l (commented):
IMO passing a message structure directly should be supported. Also, one comment regarding error checking:

common/templates/general.go
Pedro-Pessoa: Why `%d`? `val` might not be an int.

jo3-l: Good point, I overlooked that. Edited.

PR #1307, opened 2022-08-04
amod_legacy: rewrite cached invites logic

Pedro-Pessoa:
This PR aims to rewrite the logic of the cached invites in the automod_legacy package.
The old logic was messy, I find this new one much cleaner.

Other than that, this logic is orders of magnitude faster.

---- Edit
Removed the benchmarks because of the discussion in the community server, see here: https://discord.com/channels/166207328570441728/920615133993107467/1004781523557429349

### DIFF
--- a/automod_legacy/rules.go
+++ b/automod_legacy/rules.go
@@ -186,16 +186,63 @@ func (i *InviteRule) Check(evt *discordgo.Message, cs *dstate.ChannelState) (del
 	return
 }
 
-type CachedInvite struct {
-	CreatedAt time.Time
-	Invite    string // we only store the invite code, no need to waste memory on the entire invite data
+type cachedInvite struct {
+	createdAt time.Time
+	guildID   int64
 }
 
-var InvitesCache = struct {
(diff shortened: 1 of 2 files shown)


jo3-l:
I have no doubt that this is faster, but the benchmark appears flawed in multiple ways to me. For one, it never retrieves an entry that is actually in the map -- instead, it looks up the same entry over and over, which is not a representative scenario. In the case that there are no hash collisions, the branches in the map access call should be predictable and thus avoid the cost of incorrect branch prediction.

More importantly, the benchmark is not concurrent and does not interleave writes with reads. Instead, it performs a large number of writes and then a large number of reads on a single thread. These two factors cause the cost of the mutex to essentially not be factored in at all as there is no possible contention, starving, etc.

Ultimately, while the increased performance of this implementation is not in question, I recommend that the benchmark be either removed or rewritten to address these concerns. As it stands, it does not provide an accurate picture of the absolute performance of this implementation and may mislead future contributors who attempt to optimize this code further (given that it is a hot spot.)

PR #1317, opened 2022-08-12
CC: getChannelPins function

mrbentarikau:
Function `getChannelPins` will return a slice of pinned message IDs... This is still open for discussion, to return full messages versus only IDs as in this PR and later on use `getMessage`. Also context function branches are set to alphabetical order.


jo3-l:
(Note: I've already stated this privately — this comment is just for the benefit of possible reviewers considering the problem at hand.)

My thoughts on whether full messages or IDs should be returned: IMO providing direct access to messages is the way to go. Given that the API returns full messages, deliberately transforming the response to just contain IDs seems counterproductive — all it does is add a layer of indirection for users, who will still be able to use `getMessage` if needed. (Furthermore, if users do end up calling `getMessage`, we will end up making multiple API calls when one could have sufficed.)

Pedro-Pessoa:
+1 to returning the entire message.

mrbentarikau:
with full messages comes another question, returning slice of pointers []*discordgo.Message, or a slice of dereferenced messages... First would take less message space because 50 full sized pinned messages surely will not fit 2K limit.

mrbentarikau:
changed it to return complete message objects

PR #1324, opened 2022-08-23
templates: jsonToSdict

mrbentarikau:
Unmarshals given JSON to templates.SDict. sdict is commonly used and also useful for cembed.

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -671,8 +671,10 @@ func tmplMin(argX, argY interface{}) float64 {
 	return math.Min(xySlice[0], xySlice[1])
 }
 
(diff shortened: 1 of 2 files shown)


common/templates/general.go
```diff
@@ -1082,6 +1084,16 @@ func tmplJson(v interface{}) (string, error) {
 	return string(b), nil
 }
 
+func tmplJSONToSDict(v interface{}) (SDict, error) {
+	var toSDict SDict
+	err := json.Unmarshal([]byte(v.(string)), &toSDict)
```
Pedro-Pessoa, line 1089: This is a panic if v is not a string.

PR #1336, opened 2022-08-30
customcommands: limit execCC call chains

jo3-l:
**This change is as of yet completely untested; it exists to facilitate discussion regarding whether the approach taken is viable. If we agree that this is the right way to proceed, I will conduct the appropriate tests prior to undrafting this.**

---

`execCC` abuse has ran rampant ever since the function's introduction, with current limits proving futile. In certain causes, commands (mis)using `execCC` have caused whole shards to be impacted. This PR attempts to prevent such issues while minimally impacting reasonable usage. 

The key idea is to track relevant parts of `execCC` call chains (herein "chains") and limit calls based on that. A chain is created when a custom command executes another via `execCC`. If the executed command calls `execCC` again, then the chain is extended. Since chains can get very long without necessarily being abusive (e.g., consider a chain of `execCC`s with a 24h delay, which shouldn't be an issue), we only remember calls that occurred less than 5 minutes ago. If, at any time, the number of elements in the chain exceeds 20, we issue an error.

Put differently, this change limits `execCC` chains to not exceed a depth of 20 in a span of less than 5 minutes.

Feedback is appreciated and encouraged. This has the potential to be a major breaking change, and we want to get it right the first time, so there's no rush.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -165,6 +165,8 @@ type Context struct {
 
 	CurrentFrame *ContextFrame
 
+	// See DelayedRunCCData.CallChain.
+	ExecCallChain          []time.Time
 	IsExecedByLeaveMessage bool
 
 	contextFuncsAdded bool
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -105,9 +105,16 @@ func handleCustomCommandsRunNow(event *pubsub.Event) {
 }
 
 type DelayedRunCCData struct {
-	ChannelID int64  `json:"channel_id"`
-	CmdID     int64  `json:"cmd_id"`
-	UserData  []byte `json:"data"`
+	// CallChain is a rolling window tracking the previous execution times in
+	// the execCC call chain for ratelimiting purposes. For example, if command
+	// 1 calls command 2 at time t1 which then proceeds to call command 3 at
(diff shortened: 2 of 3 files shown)


BlackWolfWoof:
Idea sounds good.
That way you can still chain stuff together and it shouldn't be an issue for people not abusing it.
If i look at my code nothing should be impacted with that change (except the 1h cc, but that was already talked about with des)

ashishjh-bst:
I had an idea on this. 

We should generate a unique key on every fresh instance of execCC and pass it down the chain, if the key exists in execCC metadata then the key is not generated, and only keeps a track of chain length . 

If chainlength goes higher than x, then we throw an error "Max execCC chain length reached"

What would be the optimal value of X is something we can discuss in the support server. 

We have to put an end to the tail-recursion of CCs which bypass the min execution duration of interval CCs to do things.

jo3-l:
I'm sure I'm missing something, but I'm not clear on what benefits the key-based system confers over the proposed change.

jo3-l:
I'm going to re-close this PR for housekeeping purposes. We can discuss what, if anything, we want to change about `execCC` limits internally and adjust afterward; it's not clear that this PR is the right approach.

PR #1345, opened 2022-09-16
Implementation of Suggestion1981

savage4618:
"Violation: " removed from `/automod listviolationscount` embed. 
With a gentle shove from Mr. Ben, truncation of long Violation names was added. Emoji removal/unicode replacement was not implemented (seemed to be deemed unnecessary)

### DIFF
--- a/automod/commands.go
+++ b/automod/commands.go
@@ -205,7 +205,7 @@ func (p *Plugin) AddCommands() {
 				return "No Violations found with specified conditions", nil
 			}
 
-			out = "```" + out + fmt.Sprintf("\n%-31s Count: %d\n", "*Total*", len(listViolations)) + "```"
+			out = "```" + out + fmt.Sprintf("\n%-31s Count: %d\n", "Total", len(listViolations)) + "```"
 			return &discordgo.MessageEmbed{
 				Title:       "Violations Summary",
 				Description: out,
--- a/automod/commands.go
+++ b/automod/commands.go
@@ -199,7 +199,7 @@ func (p *Plugin) AddCommands() {
 
 			for name, count := range violations {
 				if len([]rune(name)) > 31 {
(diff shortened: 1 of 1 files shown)


automod/commands.go
```diff
@@ -198,14 +198,18 @@ func (p *Plugin) AddCommands() {
 			}
 
 			for name, count := range violations {
-				out += fmt.Sprintf("Violation: %-20s Count: %d\n", name, count)
+				if len([]rune(name)) > 31 {
+					out += fmt.Sprintf("%-30s_ Count: %d\n", string([]rune(name)[:30]), count)
+				} else {
+					out += fmt.Sprintf("%-31s Count: %d\n", name, count)
+				}
```
jo3-l, line 205: Consider using `CutStringShort`. Note that it does not have precisely the same effect -- it uses `...` to mark truncated text as opposed to `_` -- but IMO this is not a major issue; violation names should generally be short.

ashishjh-bst: I am in favour of this too, "..." is generally a globally understood term for a truncated text.

savage4618: changes reflected in 51aec6e

automod/commands.go
```diff
@@ -198,14 +198,18 @@ func (p *Plugin) AddCommands() {
 			}
 
 			for name, count := range violations {
-				out += fmt.Sprintf("Violation: %-20s Count: %d\n", name, count)
+				if len([]rune(name)) > 31 {
+					out += fmt.Sprintf("%-31s Count: %d\n", common.CutStringShort(name, 30), count)
+				} else {
+					out += fmt.Sprintf("%-31s Count: %d\n", name, count)
+				}
```
jo3-l, line 205: The conditional branch seems unnecessary now that we're using `CutStringShort`.

savage4618: You make a fair point sir. I will address when I get back to the desk.

PR #1346, opened 2022-09-17
Modified inspire.go and added exchange.go.

Shadow21AR:
Added exchange.go


lemm-e:
I don't see why a single PR is opened to make changes/additions to two separate things. Addition of a new command should be a PR of its own.

Shadow21AR:
;)

PR #1357, opened 2022-10-11
Link to Role Commands page

Wolveric:
Similarly to linking to the moderation page when moderation commands are disabled, this adds a link to the Role Commands page, when there's a configuration issue, such as YAGPDB being unable to find a specified Role Group, or the Role Group having no Role Commands.


Wolveric:
Somehow didn't anticipate that... should be a simple enough change, and removing the constant keyword from it should be fine, right...?

mrbentarikau:
don't PR if you do not test code changes out yourself

PR #1360, opened 2022-10-26
Added another thing to throw

savage4618:
Added "beans" to things to throw


BlackWolfWoof:
*happy furry noises 😹*

PR #1361, opened 2022-11-02
added nil checks on is role above

ashishjh-bst:


mrbentarikau:
what if a is not nil and b is nil, it clearly is above and not false

ashishjh-bst:
> what if a is not nil and b is nil, it clearly is above and not false

If any of the role is nill, the comparision is incorrect, because nil is not a role, hence the output should be false, as none is above the other.

ashishjh-bst:
both have to be roles for a comparison, otherwise. A isn't Above B. B isn't Above A, as they are incomparable.

PR #1362, opened 2022-11-06
feat(rep): Add links to config page

Wolveric:
The current iteration of the system, verbally directs users to the control panel, when commands produce an error because the rep system is disabled, in the assumption they'll know how to locate the correct page.

This will update this behaviour, to directly link to the relevant configuration page.


Wolveric:
Crap- I didn't mean to PR this yet... sorry...

PR #1364, opened 2022-11-06
Config page links

Wolveric:
Updating the rep and tickets system with links to their config pages in the control panel, when they error out because the system is disabled.

### DIFF
--- a/reputation/plugin_bot.go
+++ b/reputation/plugin_bot.go
@@ -158,7 +158,7 @@ var cmds = []*commands.YAGCommand{
 			}
 
 			if !conf.Enabled {
-				return repDisabledError(parsed.GuildData), nil
+				return createRepDisabledError(parsed.GuildData), nil
 			}
 
 			if !IsAdmin(parsed.GuildData.GS, parsed.GuildData.MS, conf) {
@@ -205,7 +205,7 @@ var cmds = []*commands.YAGCommand{
 			}
 
 			if !conf.Enabled {
-				return repDisabledError(parsed.GuildData), nil
+				return createRepDisabledError(parsed.GuildData), nil
 			}
 
 			if !IsAdmin(parsed.GuildData.GS, parsed.GuildData.MS, conf) {
@@ -464,7 +464,7 @@ func CmdGiveRep(parsed *dcmd.Data) (interface{}, error) {
 	}
 
 	if !conf.Enabled {
-		return repDisabledError(parsed.GuildData), nil
+		return createRepDisabledError(parsed.GuildData), nil
 	}
 
 	pointsName := conf.PointsName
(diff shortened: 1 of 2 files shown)


reputation/plugin_bot.go
```diff
@@ -35,7 +35,9 @@ func (p *Plugin) BotInit() {
 
 var thanksRegex = regexp.MustCompile(`(?i)( |\n|^)(thanks?\pP*|danks|ty|thx|\+rep|\+ ?\<\@[0-9]*\>)( |\n|$)`)
 
-var repDisabledError = "**Rep command is disabled on this server. Enable it from the control panel.**"
+func repDisabledError(guild *dcmd.GuildContextData) string {
```
jo3-l, line 38: Perhaps rename this to `createRepDisabledError`?

tickets/tickets_commands.go
```diff
@@ -27,6 +27,10 @@ const InTicketPerms = discordgo.PermissionReadMessageHistory | discordgo.Permiss
 
 var _ commands.CommandProvider = (*Plugin)(nil)
 
+func TicketsDisabledError(guild *dcmd.GuildContextData) string {
```
jo3-l, line 30: This probably shouldn't be exported. Also, consider renaming this to `createTicketsDisabledError`, similar to above.

jo3-l:
You renamed the functions, but not any of the references. I suggest that at the very least you use an editor/IDE with Go support so these issues are caught before committing.

reputation/plugin_bot.go
```diff
@@ -35,7 +35,9 @@ func (p *Plugin) BotInit() {
 
 var thanksRegex = regexp.MustCompile(`(?i)( |\n|^)(thanks?\pP*|danks|ty|thx|\+rep|\+ ?\<\@[0-9]*\>)( |\n|$)`)
 
-var repDisabledError = "**Rep command is disabled on this server. Enable it from the control panel.**"
+func createRepDisabledError(guild *dcmd.GuildContextData) string {
+	return fmt.Sprintf("**The reputation system is disabled for this server.** Enable it at: <https://yagpdb.xyz/manage/%d/reputation>.", guild.GS.ID)
```
ashishjh-bst, line 39: instead of using https://yagpdb.xyz, use `web.BaseURL()`

tickets/tickets_commands.go
```diff
@@ -27,6 +27,10 @@ const InTicketPerms = discordgo.PermissionReadMessageHistory | discordgo.Permiss
 
 var _ commands.CommandProvider = (*Plugin)(nil)
 
+func createTicketsDisabledError(guild *dcmd.GuildContextData) string {
+	return fmt.Sprintf("**The tickets system is disabled for this server.** Enable it at: <https://yagpdb.xyz/manage/%d/tickets/settings>.", guild.GS.ID)
```
ashishjh-bst, line 31: instead of using https://yagpdb.xyz/, use `web.BaseURL()`

ashishjh-bst (changes requested):

PR #1378, opened 2022-11-22
slashcommands: force NSFW to false if not present

SoggySaussages:
Currently, if a command has previously been set to NSFW when the bot is run, subsequently removing the NSFW flag does NOT update the slash command's NSFW status in Discord. This results in said command continually being restricted to NSFW channels exclusively by Discord.

I have tested this to the best of my ability, and I have personally concluded that removing `omitempty` from the `CreateApplicationCommandRequest` struct seems not to make a difference. I am open to being proven wrong. 

My tests involved setting a command to NSFW, running the bot and testing the command, then removing the NSFW from the command, running the bot, restarting Discord client, logging in and out of Discord client to be 100% certain, and running the command again. These tests concluded that removing `omitempty` did not affect Discord's limiting of the slash command to NSFW channels.

This commit simply forces the NSFW flag to be set to false during conversion to `CreateApplicationCommandRequest` if the command doesn't have the NSFW flag set. Similar testing has concluded this to be effective.

### DIFF
--- a/commands/slashcommands.go
+++ b/commands/slashcommands.go
@@ -162,6 +162,12 @@ func (p *Plugin) yagCommandToSlashCommand(cmd *dcmd.RegisteredCommand) *discordg
 		// not enabled for slash commands
 		return nil
 	}
+
+	nsfw := cast.NSFW
+	if !cast.NSFW {
+		nsfw = false // force NSFW to false if not present
+	}
+
 	t := true
 
 	_, opts := cast.slashCommandOptions()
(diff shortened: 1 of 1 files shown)


jo3-l (commented):
Does this actually make any difference? The branch `!cast.NSFW` is hit iff `cast.NSFW == false`, so `nsfw == cast.NSFW` in all possible scenarios.

SoggySaussages:
Totally agree with you. Even printf'ing it with %v and %T shows that even if the flag is not set, it's interpreted as a false bool value. 

[Des is of the impression](https://discord.com/channels/166207328570441728/166207328570441728/1044705334276673637) that it's totally coincidental and caused by Discord not handling their own NSFW slash commands properly. I believe that my testing results were correlational at best, but not cause and effect :)

PR #1381, opened 2022-11-23
forex: float support

SoggySaussages:
Adds support for float input to forex

Before:
Screenshot 2022-11-23 at 10 00 07

After:
Screenshot 2022-11-23 at 10 00 15

### DIFF
--- a/lib/dcmd/arg.go
+++ b/lib/dcmd/arg.go
@@ -150,9 +150,6 @@ func (p *ParsedArg) Float64() float64 {
 		return float64(t)
 	case uint64:
 		return float64(t)
-	case string:
-		parsed, _ := strconv.ParseFloat(t, 64)
-		return parsed
 	default:
 		return 0
 	}
--- a/lib/dcmd/arg.go
+++ b/lib/dcmd/arg.go
@@ -128,6 +128,36 @@ func (p *ParsedArg) Int64() int64 {
 	}
 }
 
+func (p *ParsedArg) Float64() float64 {
+	if p.Value == nil {
+		return 0
+	}
+	
+	switch t := p.Value.(type) {
+	case int:
+		return float64(t)
+	case int32:
+		return float64(t)
+	case int64:
+		return float64(t)
+	case float32:
+		return float64(t)
+	case float64:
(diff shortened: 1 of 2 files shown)


SoggySaussages:
Noticing now that float conversion error should return error as message instead of throwing generic command error

SoggySaussages:
Added custom response :)

jo3-l:
Is there a reason you aren't using `dcmd.Float`?

lib/dcmd/arg.go
```diff
@@ -128,6 +128,36 @@ func (p *ParsedArg) Int64() int64 {
...
+		return float64(t)
+	case uint:
+		return float64(t)
+	case uint32:
+		return float64(t)
+	case uint64:
+		return float64(t)
+	case string:
+		parsed, _ := strconv.ParseFloat(t, 64)
+		return parsed
```
jo3-l, line 155: Remove this case for consistency with `ParsedArg.Int`: it's not `Float64`'s responsibility to parse string values, As is, this also discards any parse errors, which is inadvisable.
```suggestion
```

stdcommands/forex/forex.go
```diff
@@ -41,7 +42,7 @@ var Command = &commands.YAGCommand{
 		if (to == nil) || (from == nil) {
 			return "Invalid currency code.\nCheck out available codes on: <https://api.exchangerate.host/symbols>", nil
 		}
-		output, err := requestAPI("https://api.exchangerate.host/convert?from=" + from.Code + "&to=" + to.Code + "&amount=" + amount.Str())
+		output, err := requestAPI("https://api.exchangerate.host/convert?from=" + from.Code + "&to=" + to.Code + "&amount=" + strconv.FormatFloat(amount, 'f', -1, 64))
```
jo3-l, line 45: This may be better written as a call to Sprintf.

jo3-l (changes requested):

stdcommands/forex/forex.go
SoggySaussages: Sounds good, thanks

PR #1385, opened 2022-11-28
paginatedmessages: switch to MessageSend

SoggySaussages:
Previous paginated messages were limited to message embeds only. This pr switches them over to MessageSend struct, allowing full customization.

If the paginated message contains embeds, the page number and timecode are added to the final embed. If the message contains no embeds, the page number is appended to the message content.

The paginate test command is added to demonstrate this functionality, with an optional arg to control page count, and a switch for demonstrating messages without embeds.

Previous paginated commands are also updated to use MessageSend instead.

Screenshot 2022-11-28 at 01 18 59

### DIFF
--- a/automod/commands.go
+++ b/automod/commands.go
@@ -100,7 +100,7 @@ func (p *Plugin) AddCommands() {
 			{Name: "user", Type: dcmd.UserID},
 		},
 		RequireDiscordPerms: []int64{discordgo.PermissionManageServer, discordgo.PermissionAdministrator, discordgo.PermissionBanMembers},
-		RunFunc: paginatedmessages.PaginatedCommand(0, func(data *dcmd.Data, p *paginatedmessages.PaginatedMessage, page int) (*discordgo.MessageEmbed, error) {
+		RunFunc: paginatedmessages.PaginatedCommand(0, func(data *dcmd.Data, p *paginatedmessages.PaginatedMessage, page int) (*discordgo.MessageSend, error) {
 			skip := (page - 1) * 15
 			userID := data.Switch("user").Int64()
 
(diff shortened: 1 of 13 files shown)


lemm-e:
I don't quite understand the point of PR'ing a test command to the official repository.

stdcommands/paginatetest/paginatetest.go
```diff
@@ -0,0 +1,49 @@
+package paginatetest
```
ashishjh-bst, line 1: @SoggySaussages this isn't needed.

stdcommands/stdcommands.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/botlabs-gg/yagpdb/v2/stdcommands/memstats"
 	"github.com/botlabs-gg/yagpdb/v2/stdcommands/owldictionary"
 	"github.com/botlabs-gg/yagpdb/v2/stdcommands/ping"
+	"github.com/botlabs-gg/yagpdb/v2/stdcommands/paginatetest"
```
ashishjh-bst, line 36: remove this to along with the paginatetest command.

SoggySaussages: Sounds good, thanks

mrbentarikau:
I still do not see any use cases for plain text page numbers here, only bloating already existing code with additional code, examples like `return &discordgo.MessageSend{Embeds: []*discordgo.MessageEmbed{paginatedEmbed}}, nil` instead of `return paginatedEmbed. nil`

ashishjh-bst:
The plain text pagination can be handy in commands like cc, topcommands, topservers. 
I'll try out some experiments with it before v2.14.0

mrbentarikau:
but why change all embed responses thing for this, add new pagination funcs for content messages... these are not even that necessary, you can embed-paginate all the commands you listed

ashishjh-bst:
I'll be closing this because there isn't really a need to move to sendmessages as mrben has said, and the original issue for which this was created has been fixed in a different PR.

PR #1387, opened 2022-11-29
stdcommands: Define/Urban Dictionary embed fields fix for author field

mrbentarikau:
When Urban Dictionary returns results without an author, this would error out the embed with no data for field value. This quickfix prevents this and will not remove Example, Author, Votes structure.


savv-pixel (commented):
N/a 
Making of marauders map

PR #1402, opened 2022-12-20
Update Inspire command, added `season` arg for season variants of response! (xmas)

Shadow21AR:
new flag on inspire commanf for XMAS season.
Screenshot to show it gives images related to xmas: 
image


lemm-e:
-1 on this one. Seasonal commands don't make a lot of sense imo

Shadow21AR:
Well yeah it is seasonal, but it's not like it gonna get removed from API. _and it's just 1 flag so shouldn't matter much._ v:

mrbentarikau:
Flag it as season then accepting string argument, because maybe they'll add more seasons to their query > `&season=xmas`

PR #1403, opened 2023-01-02
fixed `Forex` Command bug, where more than a million in input was considered as 1 by API cuz float.

Shadow21AR:
fixed the bug where querying huge numbers (like more than 1,000,000 bananas) gives result for only 1 banana.
before: 
image
now: 
image

### DIFF
--- a/stdcommands/forex/forex.go
+++ b/stdcommands/forex/forex.go
@@ -41,7 +41,7 @@ var Command = &commands.YAGCommand{
 		if (to == nil) || (from == nil) {
 			return "Invalid currency code.\nCheck out available codes on: <https://api.exchangerate.host/symbols>", nil
 		}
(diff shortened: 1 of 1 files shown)


stdcommands/forex/forex.go
```diff
@@ -41,7 +41,7 @@ var Command = &commands.YAGCommand{
 		if (to == nil) || (from == nil) {
 			return "Invalid currency code.\nCheck out available codes on: <https://api.exchangerate.host/symbols>", nil
 		}
-		output, err := requestAPI(fmt.Sprintf("https://api.exchangerate.host/convert?from=%s&to=%s&amount=%g", from.Code, to.Code, amount))
+		output, err := requestAPI(fmt.Sprintf("https://api.exchangerate.host/convert?from=%s&to=%s&amount=%d", from.Code, to.Code, 1))
```
SoggySaussages, line 44:
```suggestion
		output, err := requestAPI(fmt.Sprintf("https://api.exchangerate.host/convert?from=%s&to=%s&amount=1", from.Code, to.Code))
```

PR #1416, opened 2023-01-19
refactor(moderation): No mute role error impl

Wolveric:
Updating the implementation when a mute role isn't selected, to use `web.ManageServerURL`, inline with other similar errors.


ashishjh-bst:
If there are other sections where you want to replace it, do it in this PR

Wolveric:
Already did it with the PR #1412 . It's just this was changed before I made that, so it wasn't on master, and wasn't sure what to do about it, so mentioned in that PR, and left it until it was merged for the next release...

PR #1426, opened 2023-01-26
.Channel.OwnerID object (suggestion #2034)

Ranger-4297:
This PR implements the `.Channel.OwnerID` object.
This returns `nil` if the `.Channel` type is `GUILD_TEXT`
And the thread owners snowflake if the `.Channel` type is `PUBLIC_THREAD` or `PRIVATE_THREAD`


Ranger-4297:
Wrong branch. Will sort this now

PR #1431, opened 2023-02-02
refactor(youtube): Refactor parseYtUrl

Wolveric:
Use a closure to minimise code duplication between code branches in `parseYtUtl` function.

### DIFF
--- a/youtube/feed.go
+++ b/youtube/feed.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"regexp"
 
 	"github.com/botlabs-gg/yagpdb/v2/analytics"
 	"github.com/botlabs-gg/yagpdb/v2/common"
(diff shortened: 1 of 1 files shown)


youtube/feed.go
```diff
@@ -277,37 +278,45 @@ var (
 )
 
 func (p *Plugin) parseYtUrl(url string) (t ytUrlType, id string, err error) {
+	validateCaptures := func(expr *regexp.Regexp, i int) (id string, valid bool) {
```
ashishjh-bst, line 281: move this to Plugin struct instead of keeping as a lambda function

`func (p *Plugin) getIDFromRegexURL(expr *regexp.Regex, url string, index int) (id string, valid bool)`

ashishjh-bst (changes requested):
looks okay, but lambda function is unnecessary

Wolveric:
Tbh, my reasoning there was that this logic is pretty specific to the function, but I'll break it out into a dedicated function...

There's also a couple of issues in the logic I should probably address, as if the resulting matches don't align with the index provided, it'll be trying to access an index which is out of range.

Wolveric:
Sorry, Idk why that duplicated import got added...

youtube/feed.go
```diff
@@ -276,39 +277,34 @@ var (
...
+		idType = ytUrlTypeHandle
+	} else {
+		idType = ytUrlTypeInvalid
+		err = errors.New("invalid or incomplete url")
+	}
+
+	return idType, id, err
+}
+
+func (p *Plugin) getIDFromRegexUrl(expr *regexp.Regexp, url string, index int) string {
```
jo3-l, line 301: [Avoid in-band errors](https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors); return `(string, bool)` instead.

youtube/feed.go
```diff
@@ -276,39 +277,34 @@ var (
...
+	} else if id = p.getIDFromRegexUrl(ytUserUrlRegex, url, 5); id != "" {
+		idType = ytUrlTypeUser
+	} else if id = p.getIDFromRegexUrl(ytHandleUrlRegex, url, 5); id != "" {
+		idType = ytUrlTypeHandle
+	} else {
+		idType = ytUrlTypeInvalid
+		err = errors.New("invalid or incomplete url")
+	}
+
+	return idType, id, err
```
jo3-l, line 298: I am unconvinced the naked return pulls its weight here; what does this buy us over the simpler
```go
if id := p.getIDFromRegexUrl(...); id != "" {
	return ytUrlTypeVideo, id, nil
} else if ... {

}
...
return ytUrlTypeInvalid, "", errors.New("invalid or incomplete url")
```
?

youtube/feed.go
Wolveric: Fair enough, I'll change that back...

youtube/feed.go
Wolveric: It minimises the number of elements which are repeated, and as named return values are being used, just a plain `return` statement can be used with this structure, which would implicitly return the values set for the named variables.

jo3-l: I'm aware of what naked returns are. My point is that just because they *can* be used does not mean that they *should* be used. See the [relevant section in the Go wiki](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters), which advises

> Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values.

Wolveric: Ok, I can revert switching to a naked return,

youtube/feed.go
```diff
@@ -276,39 +277,36 @@ var (
...
+}
+
+func (p *Plugin) getIDFromRegexUrl(expr *regexp.Regexp, url string, index int) (id string, valid bool) {
+	capturingGroups := expr.FindAllStringSubmatch(url, -1)
+	if len(capturingGroups) > 0 && len(capturingGroups[0]) > index {
+		id = capturingGroups[0][index]
+		valid = id != ""
+	}
+
+	return
```
jo3-l, line 309: Remove the naked return here as well — it adds no value.
```suggestion
	capturingGroups := expr.FindAllStringSubmatch(url, -1)
	if len(capturingGroups) > 0 && len(capturingGroups[0]) > index {
		return capturingGroups[0][index], true
	}
	return "", false
```

youtube/feed.go
```diff
@@ -276,39 +277,36 @@ var (
...
+	} else {
+		idType = ytUrlTypeInvalid
+		err = errors.New("invalid or incomplete url")
 	}
-	return ytUrlTypeInvalid, "", errors.New("invalid or incomplete url")
+
+	return idType, id, err
+}
+
+func (p *Plugin) getIDFromRegexUrl(expr *regexp.Regexp, url string, index int) (id string, valid bool) {
```
jo3-l, line 302:
```suggestion
func (p *Plugin) getIDFromRegexUrl(expr *regexp.Regexp, url string, index int) (id string, ok bool) {
```
`ok` is the canonical name for return values of this type.

Alternately, you can omit the return value names entirely; they're sufficiently clear in context.

youtube/feed.go
```diff
@@ -276,39 +277,36 @@ var (
...
+		idType = ytUrlTypeUser
+	} else if id, valid = p.getIDFromRegexUrl(ytHandleUrlRegex, url, 5); valid {
+		idType = ytUrlTypeHandle
+	} else {
+		idType = ytUrlTypeInvalid
+		err = errors.New("invalid or incomplete url")
 	}
-	return ytUrlTypeInvalid, "", errors.New("invalid or incomplete url")
+
+	return idType, id, err
```
jo3-l, line 299: Sorry, perhaps I wasn't clear in my previous comment. When I advised removing the naked return, I was suggesting something along the following lines:
```go
if id, ok := p.getIDFromRegexUrl(/* ... */); ok {
	return ytUrlTypeVideo, id, nil
} else if /* ... */ {

}
return ytUrlTypeInvalid, "", errors.New("invalid or incomplete url")
```

jo3-l (commented):
Couple of further nitpicks. I think it might be prudent to wait for @ashishjh-bst to chime in regarding whether these edits are worthwhile; he might be fine with merging as is. (Then again, this PR is purely a refactor and so we might as well spend some time ensuring the code is as idiomatic as possible.)

youtube/feed.go
Wolveric: This actually changes the existing behaviour, as the `if` statement succeeding doesn't result in `valid`/`ok` being implicitly true.

In this case, the named variables make the resulting `!= ""` check more concise.

Wolveric: In the existing implementation, this was expressed as `len(capturingGroups[0][i]) > 0` instead, though these are evidently equivilent.

jo3-l: The check is actually unnecessary as the regex patterns in question ensure that the relevant capturing groups will always be non-empty, but fair point. Deferring to @ashishjh-bst.

ashishjh-bst: The regex handles that, so this check will never be done here.

youtube/feed.go
ashishjh-bst: Do you intend to do that anytime soon @Wolveric ?

Wolveric:
I was thinking of closing this after looking over the system more extensively, pending a more extensive rework of the system...

Wolveric:
Closing in favour of PR #1466.

PR #1450, opened 2023-02-23
Automod: Sanitize content on text based triggers

User9684:
Adds toggles to every text based AMv2 rule for sanitizing text

Also adds "sanitize" function to CCs


User9684:
Oop test failed.

Lemme see what I did wrong rq lol

User9684:
Oh I think it's just because the package doesn't exist on the actual repo

jo3-l:
Note that there are many more obvious "lookalike" characters than those included here, such as `Α` (GREEK CAPITAL LETTER ALPHA). I do recognize that you've selected a subset of "common" characters, but these are unlikely to satisfy all clients. For this reason, I'm not convinced that this feature belongs in core; the criteria for whether a character is "spoofed" is too broad for the trigger to be useful.

User9684:
> Note that there are many more obvious "lookalike" characters than those included here, such as `Α` (GREEK CAPITAL LETTER ALPHA). I do recognize that you've selected a subset of "common" characters, but these are unlikely to satisfy all clients. For this reason, I'm not convinced that this feature belongs in core; the criteria for whether a character is "spoofed" is too broad for the trigger to be useful.

Yes, I am aware that there are a lot more than what I have provided. At this current time, I just nabbed all the characters that https://onlineunicodetools.com/spoof-unicode-text uses, but I do plan on adding more into the map.

User9684:
I believe this would be something such as the nouns, verbs, etc. lists, as they obviously always can add more, but the lists are still present.

common/templates/context.go
```diff
@@ -53,6 +53,7 @@ var (
 		"print":       withOutputLimit(fmt.Sprint, MaxStringLength),
 		"println":     withOutputLimit(fmt.Sprintln, MaxStringLength),
 		"printf":      withOutputLimitF(fmt.Sprintf, MaxStringLength),
+		"deSpoof":     common.DeSpoof,
```
jo3-l, line 56: Out of curiosity, have you tested this? `common.DeSpoof` has two return values (the second of which isn't the `error` type), which I don't believe is supported.

User9684: I sadly have not tested it yet (hold your laughter, please)
But, if it doesn't support returning str & bool then I can probably make a quick fix for that

User9684: Well, I tested the func itself, but I still struggle to get a bare bones selfhost functional lol

User9684: Yeah from what I see at line 312, I think it might not support returning str and bool, I can just move the logic for checking if message was modified into CheckMessage though.

common/spoofdetection.go
```diff
@@ -0,0 +1,66 @@
...
+	// Invisible characters
+	"\u200A", "",
+	"\u200B", "",
+	"\u200C", "",
+	"\u200D", "",
+	"\u2060", "",
+	"\uFEFF", "",
+)
+
+func DeSpoof(input string) (ouput string, err error) {
```
jo3-l, line 62:
```suggestion
func DeSpoof(input string) string {
```
The second return value is now unneeded.

common/spoofdetection.go
```diff
@@ -0,0 +1,66 @@
...
+	output := DeSpoofReplacer.Replace(input)
+
+	return output, nil
```
jo3-l, line 65:
```suggestion
	return DeSpoofReplacer.Replace(input), nil
```

automod/triggers.go
```diff
@@ -1483,3 +1483,34 @@ func (ml *MessageLengthTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dst
...
+}
+
+func (st *SpoofedTextTrigger) UserSettings() []*SettingDef {
+	return []*SettingDef{}
+}
+
+func (st *SpoofedTextTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dstate.ChannelState, m *discordgo.Message, mdStripped string) (bool, error) {
+	str, _ := common.DeSpoof(m.Content)
+
+	return (m.Content != str), nil
```
jo3-l, line 1515:
```suggestion
	str, _ := common.DeSpoof(m.Content)

	return m.Content != str, nil
```
Consider also combining both lines into one.

automod/triggers.go
```diff
@@ -1483,3 +1483,34 @@ func (ml *MessageLengthTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dst
...
+func (st *SpoofedTextTrigger) DataType() interface{} {
+	return nil
+}
+
+func (st *SpoofedTextTrigger) Name() (name string) {
+	return "Message includes spoofed unicode characters"
+}
+
+func (st *SpoofedTextTrigger) Description() (description string) {
+	return "Triggers on messages where common forms of automod bypassing is detected (i.e. a-zero width-joiner)"
```
jo3-l, line 1505: "common forms of automod bypassing" is rather vague, but I don't have a better suggestion at the moment.

common/spoofdetection.go
User9684: Oh I kinda expected it to want a 2nd value

automod/triggers.go
User9684: Yeaaah I was having issues forming an actual good description so I just spat that out, lol

User9684:
Wrong button

User9684:
lmk if you can think of any more confusable characters not included

mrbentarikau:
go has its own normalization tools, why use this manual file for char comparisons?
https://go.dev/blog/normalization

User9684:
> go has its own normalization tools, why use this manual file for char comparisons? https://go.dev/blog/normalization

I was not aware of this, nor have I tested that lib yet. I'll check if it's better in a bit

User9684:
Just tested, and the output of `unicode/norm` does not give adequate normalization. I.e. "Crillic Small Letter A" is not normalized into "Latin Small Letter A"

Tested code:
```go
package main

import (
	"fmt"
	"unicode"

	"golang.org/x/text/transform"
	"golang.org/x/text/unicode/norm"
)

func isMn(r rune) bool {
	return unicode.Is(unicode.Mn, r)
}

func main() {
	x := []rune{1072, 1072, 1072, 1072} // аааа
	z := []rune{97, 97, 97, 97}         // aaaa
	t := transform.Chain(norm.NFD, transform.RemoveFunc(isMn), norm.NFC)
	normStr1, _, _ := transform.String(t, string(x))

	fmt.Println(normStr1 == string(z))
}
```

mrbentarikau:
for cyrillic matching, this is not normalization, but this package should help... "the confusables"  : ) https://github.com/oragono/confusables

User9684:
> for such Cyrillic matchings etc, this is already not normalization, but this library should help... "the confusables" : ) https://github.com/oragono/confusables
> 
> this is similar to your common/normalize.go, just maybe more complete

The library mentioned just looks to be an overcomplicated replacer.

mrbentarikau:
maybe instead of making separate normalize.go file, just place those functions inside common/util.go

common/normalize.go
```diff
@@ -0,0 +1,32 @@
...
+	"golang.org/x/text/transform"
+	"golang.org/x/text/unicode/norm"
+)
+
+func isMn(r rune) bool {
+	return unicode.Is(unicode.Mn, r)
+}
+
+// stupid name, cant think of a better one atm
+func FixText(content string, diacritics, matchConfusables bool) string {
```
l-zeuch, line 16: Given how you're already having a `matchConfusables`, also consider this change:
```suggestion
func FixText(content string, removeDiacritics, matchConfusables bool) string {
```
(Or something similar to that extent)

automod/triggers.go
```diff
@@ -47,6 +48,12 @@ func (r BaseRegexTrigger) UserSettings() []*SettingDef {
 			Min:  1,
 			Max:  250,
 		},
+		{
+			Name:    "FixText the message content and check the result aswell",
```
l-zeuch, line 52:
```suggestion
			Name:    "Also match Zalgo and confusable characters",
```
For all instances where you have added this.

l-zeuch (changes requested):

automod/triggers.go
User9684: I forgot to change those oops

common/normalize.go
User9684: Agreed, more consistent that way

User9684:
> maybe instead of making separate normalize.go file, just place those functions inside common/util.go

honestly, yea.

common/normalize.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var transformer = transform.Chain(norm.NFD, transform.RemoveFunc(isMn), norm.NFC)
+
+func isMn(r rune) bool {
+	return unicode.Is(unicode.Mn, r)
+}
+
+// stupid name, cant think of a better one atm
+func FixText(content string) string {
```
ashishjh-bst, line 18: SanitizeText?

User9684: That was actually the original name and then it was changed like 80 times lmao, ill change back to SanitizeText

User9684:
Any updates & things I need to change?

automod/triggers.go
```diff
@@ -47,6 +48,12 @@ func (r BaseRegexTrigger) UserSettings() []*SettingDef {
 			Min:  1,
 			Max:  250,
 		},
+		{
+			Name:    "Also match Zalgo and confusable characters",
```
ashishjh-bst, line 52: Make it simple for a average user? What does Zalgo and confusable characters here mean?

automod/triggers.go
```diff
@@ -180,6 +188,12 @@ func (wl *WordListTrigger) UserSettings() []*SettingDef {
 			Key:  "ListID",
 			Kind: SettingTypeList,
 		},
+		{
+			Name:    "Also match Zalgo and confusable characters",
```
ashishjh-bst, line 192: Make it simple for a average user? What does Zalgo and confusable characters here mean?

automod/triggers.go
```diff
@@ -448,6 +468,12 @@ func (caps *AllCapsTrigger) UserSettings() []*SettingDef {
 			Min:     1,
 			Max:     100,
 		},
+		{
+			Name:    "Also match Zalgo and confusable characters",
```
ashishjh-bst, line 472: Make it simple for a average user? What does Zalgo and confusable characters here mean?

automod/triggers.go
```diff
@@ -938,7 +970,7 @@ func (r *MessageRegexTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dstat
 	}
 
 	re := item.Value().(*regexp.Regexp)
-	if re.MatchString(m.Content) {
+	if re.MatchString(m.Content) || dataCast.SanitizeText && re.MatchString(common.SanitizeText(m.Content)) {
```
ashishjh-bst, line 973: create a separate variable here to store this `common.SanitizeText(m.Content)`

automod/triggers.go
```diff
@@ -1034,9 +1073,16 @@ func (spam *SpamTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dstate.Cha
 
 		if strings.ToLower(strings.TrimSpace(v.Content)) == mToCheckAgainst {
 			count++
-		} else {
-			break
+			continue
 		}
+
+		// this if statement makes me want to cry
+		if settingsCast.SanitizeText && strings.ToLower(strings.TrimSpace(common.SanitizeText(v.Content))) == mToCheckAgainst {
```
ashishjh-bst, line 1080: If this statement makes you cry then make it simpler? why do you want to write oneliners? Create variables  with computed information and then use them here.

automod/triggers.go
```diff
@@ -1087,7 +1133,7 @@ func (r *NicknameRegexTrigger) CheckNickname(t *TriggerContext) (bool, error) {
 	}
 
 	re := item.Value().(*regexp.Regexp)
-	if re.MatchString(t.MS.Member.Nick) {
+	if re.MatchString(t.MS.Member.Nick) || dataCast.SanitizeText && re.MatchString(common.SanitizeText(t.MS.Member.Nick)) {
```
ashishjh-bst, line 1136: don't make conditions unnecessarily complex, store computed values in a variable and to make it readable, I don't want to guess what you're trying to do.

automod/triggers.go
```diff
@@ -1143,6 +1190,12 @@ func (nwl *NicknameWordlistTrigger) UserSettings() []*SettingDef {
 			Key:  "ListID",
 			Kind: SettingTypeList,
 		},
+		{
+			Name:    "Also match Zalgo and confusable characters",
```
ashishjh-bst, line 1194: Make it simple for a average user? What does Zalgo and confusable characters here mean?

Also, I see it being repeated a lot in this file, maybe create a single variable and use it everywhere?

automod/triggers.go
```diff
@@ -1155,6 +1208,10 @@ func (nwl *NicknameWordlistTrigger) CheckNickname(t *TriggerContext) (bool, erro
 	}
 
 	fields := strings.Fields(PrepareMessageForWordCheck(t.MS.Member.Nick))
+	if dataCast.SanitizeText {
+		messageFieldsFixTextd := strings.Fields(common.SanitizeText(PrepareMessageForWordCheck(t.MS.Member.Nick)))
```
ashishjh-bst, line 1212: `messageFieldsFixTextd` what does the extra `d` at the end signify in the variable name?

automod/triggers.go
```diff
@@ -1220,7 +1277,7 @@ func (r *UsernameRegexTrigger) CheckUsername(t *TriggerContext) (bool, error) {
 	}
 
 	re := item.Value().(*regexp.Regexp)
-	if re.MatchString(t.MS.User.Username) {
+	if re.MatchString(t.MS.User.Username) || dataCast.SanitizeText && re.MatchString(common.SanitizeText(t.MS.User.Username)) {
```
ashishjh-bst, line 1280: reduce cognitive load in your conditions, make them readable and simple.

automod/triggers.go
```diff
@@ -1276,6 +1334,12 @@ func (uwl *UsernameWordlistTrigger) UserSettings() []*SettingDef {
 			Key:  "ListID",
 			Kind: SettingTypeList,
 		},
+		{
+			Name:    "Also match Zalgo and confusable characters",
```
ashishjh-bst, line 1338: I have written about this too many times.

automod/triggers.go
```diff
@@ -1288,6 +1352,10 @@ func (uwl *UsernameWordlistTrigger) CheckUsername(t *TriggerContext) (bool, erro
 	}
 
 	fields := strings.Fields(PrepareMessageForWordCheck(t.MS.User.Username))
+	if dataCast.SanitizeText {
+		messageFieldsFixTextd := strings.Fields(common.SanitizeText(PrepareMessageForWordCheck(t.MS.User.Username)))
```
ashishjh-bst, line 1356: Same copy pasta everywhere.

common/util.go
```diff
@@ -654,3 +658,20 @@ func FormatList(list []string, conjunction string) string {
 func BoolToPointer(b bool) *bool {
 	return &b
 }
+
+// Sanitize text to match confusables and remove diacritics
+var transformer = transform.Chain(norm.NFD, transform.RemoveFunc(isMn), norm.NFC)
+
+func isMn(r rune) bool {
```
ashishjh-bst, line 665: what is this function name, and what is the use of this function?

common/util.go
```diff
@@ -654,3 +658,20 @@ func FormatList(list []string, conjunction string) string {
...
+	return unicode.Is(unicode.Mn, r)
+}
+
+func SanitizeText(content string) string {
```
ashishjh-bst, line 669: Add some documentation above the function on what it does. "SanitizeText" could mean a lot of things.

ashishjh-bst (changes requested):
Too many readability issues and a lot of repetitive code.

automod/triggers.go
User9684: Yeah I just stole this text from Luca

common/util.go
User9684: I just checked and it appears that RemoveFunc is depreciated anyways, but it's so the transformer knows whether or not to replace the char or not.

common/util.go
User9684: Where would I put the documentation? I'm kinda dumb

automod/triggers.go
User9684: Would it be worth to check if SanitizeText is valid and just replace the argument passed to MatchString?

automod/triggers.go
User9684: True

ashishjh-bst: Also, please only add comments that are descriptive of your code and not your emotions. This isn't your personal project.

automod/triggers.go
User9684: That would definitely work to just have it as a variable to easily change

automod/triggers.go
User9684: That, was an accident. Thanks for picking that up lol

automod/triggers.go
User9684: Yeah I mostly put that there so someone would comment specifically on it

ashishjh-bst (approved):

PR #1451, opened 2023-02-25
customcommands: draft API proposal / specification

l-zeuch:
This draft pull request proposes an API to manage custom commands.

It serves as a place to discuss thoughts, ideas, and questions from the community.

For ease of viewing the markdown file, please follow [this link](https://github.com/l-zeuch/yagpdb/blob/feat/cc-api/customcommands/cc-api.md).

If you have direct feedback, feel free to use the "review" feature and leave a comment (or, if you're a maintainer, request changes as you deem necessary).

For general questions, just leave a general comment. For *just* voting, please use reactions, as the "+1"/"-1" comments tend to drown actual conversations, as they are rather spammy.

### DIFF
--- a/customcommands/api.go
+++ b/customcommands/api.go
@@ -1,0 +1,111 @@
+package customcommands
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/botlabs-gg/yagpdb/v2/common/config"
+	"github.com/botlabs-gg/yagpdb/v2/customcommands/models"
+	"github.com/botlabs-gg/yagpdb/v2/premium"
+	"github.com/botlabs-gg/yagpdb/v2/web"
+	"github.com/volatiletech/sqlboiler/queries/qm"
+	"goji.io"
+
+	"goji.io/pat"
+)
+
+var (
+	confCCAPIEnabled        = config.RegisterOption("yagpdb.custom_command_api.enabled", "Enable custom command management API", false)
+	confCCAPIRequirePremium = config.RegisterOption("yagpdb.custom_command_api.require_premium", "Require guild to be premium for API access", true)
+)
+
+func (p *Plugin) initAPI() {
+	if !confCCAPIEnabled.GetBool() {
+		logger.Warn("Custom commands API disabled, not starting it")
+		return
+	}
+
+	logger.Info("Starting custom commands API")
+
+	subMux := goji.SubMux()
+	subMux.Use(canUseAPIMW)
+
+	// TODO: This needs better middleware, during development we don't care right now
+	subMux.Use(web.RequireActiveServer)
+
+	web.ServerPublicAPIMux.Handle(pat.New("/customcommands"), subMux)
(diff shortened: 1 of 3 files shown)


customcommands/api.go
```diff
@@ -0,0 +1,129 @@
...
+// serving as a wrapper around the same.
+type APICustomCommand struct {
+	ID            int64              `json:"id"`
+	Name          string             `json:"name" valid:",0,100"`
+	IsEnabled     bool               `json:"is_enabled"`
+	TriggerType   CommandTriggerType `json:"trigger_type"`
+	Trigger       string             `json:"trigger" valid:",0,1000"`
+	CaseSensitive bool               `json:"case_sensitive"`
+	Responses     []string           `json:"responses" valid:"template,10000"`
+}
```
jo3-l, line 61: `valid` tags aren't necessary here.

customcommands/api.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		// Doesn't have premium, send 403.
+		if !ok {
+			http.Error(w, "Guild does not have premium.", http.StatusForbidden)
+			return
+		}
```
jo3-l, line 86:
```suggestion
		isPrem, err := premium.IsGuildPremium(activeGuild.ID)
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}

		if !isPrem {
			http.Error(w, "Guild does not have premium.", http.StatusForbidden)
			return
		}
```
Renaming `ok` to something a little more descriptive obviates the need for the comment.

jo3-l (commented):
I'm aware this isn't fully done yet — just wanted to leave these comments before I forget :-) Feel free to address whenever ready.

customcommands/api.go
l-zeuch: Honestly, I'm not even sure whether this whole struct is needed. There's a few questions I have to solve, the main one is how to serialise the `models.CustomCommand` struct into the above.

I'll remove the tags in the coming commit(s) and see how to proceed from there.

l-zeuch:
> I'm aware this isn't fully done yet — just wanted to leave these comments before I forget :-) Feel free to address whenever ready.

Thanks for the feedback, I appreciate it! I prefer small commits over a huge chunk either way, that way it is easier for all involved to review and address feedback.

l-zeuch:
I'll be closing this PR for the time being as to not fill the list with
junk. I'll reopen it when I have the time to work on it again and
refined my Go skills a bit more — this is not to say it won't happen,
but as explained in my initial comment we want to get it right the first
time and not rush it.

The proposal itself could also use some refining, so I'll be working on
that with a higher priority than implementing the API itself — there's
still some things I am yet to learn about the framework(s) we use, so I
am going to take my time with this.

If you have any resources on hand that could help me with this, please
do share them with me.

PR #1453, opened 2023-03-05
Custom commands: Forum support

Ranger-4297:
This PR aims to add a *slightly* larger support bracket for forums and forum threads within Custom Commands.

Currently, YAGPDB can access a limited number of forum based data points from `getThread`/`getChannel`.

`getChannel <forum channel ID>` now returns the tag ID, tag name, moderator only status and emoji name/emoji ID (if applicable) of `AvailableTags` in a cslice (if any)

> **Note**
> `.AvailableTags` returns empty slice for all other channels.

`getThread <forum thread ID>` now returns the IDs of `AppliedTags` in a cslice (if any)

> **Note**
> `.AppliedTags` returns empty slice for all other channels.

Introducing the `setForumTags` (open to change) template which allows you to set tags within a forum thread. Examples below;
There is a limit to 5 tags at a time on a post which is reflected within the template. (Duplicate entries of a tag will not count towards this)

`{{setForumTags 1068169527344500767 (cslice)}}` to remove all tags
`{{setForumTags 1068169527344500767 (111 222 333)}}` to set tags


common/templates/structs.go
```diff
@@ -25,6 +26,8 @@ type CtxChannel struct {
 	PermissionOverwrites []*discordgo.PermissionOverwrite `json:"permission_overwrites"`
 	ParentID             int64                            `json:"parent_id"`
 	OwnerID              int64                            `json:"owner_id"`
+	AppliedTags          []string                         `json:"applied_tags"`
```
l-zeuch, line 29: Why is this a `[]string` when the Discord docs [^1] indicate an array
of snowflakes, i.e. `[]int64`?

Any reason not to have this as a `[]ForumTag`?
Obviously we'll be still ignoring the Discord docs for this, but a
`[]string` is really, really odd.

[^1]: https://discord.com/developers/docs/resources/channel#channel-object-channel-structure

lib/discordgo/structs.go
```diff
@@ -237,6 +241,12 @@ type Channel struct {
 
 	// Thread specific fields
 	ThreadMetadata *ThreadMetadata `json:"thread_metadata"`
+
+	// Access to the current tags in a forum thread
+	AppliedTags []string `json:"applied_tags"`
```
l-zeuch, line 246: Dito

lib/dstate/interface.go
```diff
@@ -214,6 +214,8 @@ type ChannelState struct {
 	RateLimitPerUser int                       `json:"rate_limit_per_user"`
 	OwnerID          int64                     `json:"owner_id,string"`
 	ThreadMetadata   *discordgo.ThreadMetadata `json:"thread_metadata"`
+	AppliedTags      []string                  `json:"applied_tags"`
```
l-zeuch, line 217: Same here as well

common/templates/context_funcs.go
```diff
@@ -1632,6 +1632,73 @@ func (c *Context) tmplEditChannelTopic(channel interface{}, newTopic string) (st
 	return "", err
 }
 
+const ChannelTagLimit = 5
```
l-zeuch, line 1635: Is this a Discord limitation, or arbitrarily set?
If it is a Discord limitation, consider returning the error thrown by
Discord over throwing our own and obfuscating the source.

common/templates/structs.go
Ranger-4297: The object is `[]string` to allow compatibility within the `setForumTags` template
Would it make more sense to have it as `[]ForumTag`? How come if so?

l-zeuch: > The object is `[]string` to allow compatibility within the 
> `setForumTags` template

Nowhere in `tmplSetForumTags` are you returning a non-empty string.
Consider just returning the error instead, or having sensible return 
values and/or types.

>  Would it make more sense to have it as `[]ForumTag`? How come if so?

If it is of type `[]ForumTag` users would be able to directly `index`
said slice and have the `ForumTag` object ready from the get-go:

```go
{{/* Grab the first ForumTag and output its information */}}
{{ index .Channel.AppliedTags 0 | json }}
```

Otherwise, they would have to somehow get the `ForumTag` object, which
I do not see a template for in your patchset. Though, maybe also add 
such a function as well.

common/templates/context_funcs.go
Ranger-4297: Was a Discord limit I hardcoded into it. Fixed in latest commit

common/templates/context_funcs.go
```diff
@@ -1632,6 +1632,67 @@ func (c *Context) tmplEditChannelTopic(channel interface{}, newTopic string) (st
 	return "", err
 }
 
+func (c *Context) tmplSetForumTags(channel interface{}, input interface{}) (string, error) {
```
ashishjh-bst, line 1635: instead of repurposing this to remove forumtags, maybe create a separate template to removeForumTags, and removeAllForumTags. 

`set` should ideally just set tags. 

I would've personally preferred something like `addForumTags` and `removeForumTags` instead of set which modified the complete list of tags.

l-zeuch: `setRoles` has the same behaviour, it removes roles
the target has but weren't given.

Changing `setTags` to only add would be a bit of an
inconsistency.

common/templates/structs.go
Ranger-4297: Spent some time messing around with the AppliedTags object.
Figured out that for some unknown reason, despite the discord developer docs stating that it is an array of snowflakes, returning it in `[]int64` would result in a response of 0 for any tag.
Can't tell you why it's doing that so that's why I opted with `[]string`.

As for returning it with `[]ForumTag`, similar issue. We get a response and it returns data however it returns blank or again with 0's.

Will resolve the other conversations pertaining to this aside from this one. Waiting on a response from Des in programming discussion and we'll go from there

common/templates/context_funcs.go
Ranger-4297: As Lzodd said, `setRoles` has the same behaviour and implementing another function to remove it wouldn't be necessary.
However, there are also add/removerole based methods. and I wouldn't be against implementing `addTag`/`removeTag` methods as well if you feel as though it would be useful to also have them

common/templates/structs.go
ashishjh-bst: Is this still pending?

Ranger-4297: Yep

PR #1462, opened 2023-03-08
lib/dcmd: don't accept any number when Min == Max

l-zeuch:
Stop accepting any number when the specified Min and Max are equal to
each other. Instead, return an OutOfRangeError.

Apply this to all number args we have (i.e. IntArg and FloatArg).

Expected breakage should be low to none. One would generally expect that
when specifying a range with the same Min as Max that only that number
is to be accepted.

See also:
https://canary.discord.com/channels/166207328570441728/356486960417734666/1010540772367540284


l-zeuch:
After investigating a bit, this change feels
impossible to achieve, as *not* providing a 
Min and Max will result in a [0, 0] range,
thus always failing a parseArgs call that does
not have a range given.

Closing this.

mrbentarikau:
```
else if (i.Min != 0 || i.Max != 0) && i.Max == i.Min {
		if v != i.Max {
			return nil, &OutOfRangeError{ArgName: def.Name, Got: v, Min: i.Min, Max: i.Max}
		}
```

l-zeuch:
Oh wow 🤦 

This is why I shouldn't write code with just four hours of sleep.

PR #1463, opened 2023-03-10
Fix several issues with list generation

User9684:
- Fixes issue where a space was added at the end of replaced characters
- Fixes issue where characters from extended-latin ranges were considered normal latin
- Fixed issue where unicode IDs with 5 or more charactes weren't being defined correctly


User9684:
the reason why all those commits are there is beyond me

PR #1471, opened 2023-03-18
run prettier

2srqc:


User9684:
Nobody is going to want to review 1,400 file changes. Maybe instead suggest to Des to use this "prettier" formatter, rather than just dropping a bomb

PR #1479, opened 2023-03-26
Adding extra characters to extraConfusables

User9684:
Duck spam.


User9684:
haven't generated yet, this draft is so there doesn't need to be a crap ton of prs when a character is discovered

PR #1482, opened 2023-03-29
customcommands: evalcc command

l-zeuch:
Evaluate small code snippets quickly. Can only be used by beings of
high power, as it requires manage_guild permissions (and/or control panel read access).

This is a combination of several commits:

* customcommands: improve evalcc parsing

* customcommands: improve `evalcc` permissions

  Previously, this command only allowed users with  manage_server
  permissions to run this command. This is arguably bad design --
  server staff with write access to the control panel are just as able to
  create custom commands, thus should be able to run this command, too.

* customcommands: optimise check for AllowedWriteRoles

  Moving the call outside of the `range` reduces overhead (significance
  may be debatable), thus being preferable.

### DIFF
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -61,7 +61,7 @@ var _ bot.BotInitHandler = (*Plugin)(nil)
 var _ commands.CommandProvider = (*Plugin)(nil)
 
 func (p *Plugin) AddCommands() {
-	commands.AddRootCommands(p, cmdListCommands, cmdFixCommands)
+	commands.AddRootCommands(p, cmdListCommands, cmdFixCommands, cmdEvalCommand)
 }
 
 func (p *Plugin) BotInit() {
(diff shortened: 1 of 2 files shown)


customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+	parts := codeblockRegexp.FindStringSubmatch(input)
+
+	// No match found, input was not wrapped in (valid) codeblock markdown
+	// just dump it, don't bother fixing things for the user.
+	if parts == nil {
+		return input
+	}
+
+	logger.Debugf("Found matches: %#v", parts)
+	logger.Debugf("Returning %s", parts[1])
```
jo3-l, line 201: Can probably remove this?

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+		ctx.IsExecedByEvalCC = true
+
+		maxRunes := 500
+		if ctx.IsPremium {
+			maxRunes = 1000
+		}
+
+		code := data.Args[0].Str()
+
+		code = parseCodeblock(code)
```
jo3-l, line 166:
```suggestion
		code := parseCodeblock(data.Args[0].Str())
```

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+
+		// Encourage only small code snippets being tested with this command
+		if utf8.RuneCountInString(code) > maxRunes {
+			return "Code is too long for in-place evaluation. Please use the control panel.", nil
+		}
+
+		if channel == nil {
+			return "Something weird happened... Contact the support server.", nil
+		}
```
jo3-l, line 175: What edge case is this guarding against?

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+
+		out, err := ctx.Execute(code)
+
+		if err != nil {
+			errFormatted := err.Error()
+			return "An error caused the custom command to stop:\n`" + errFormatted + "`", nil
```
jo3-l, line 181:
```suggestion
			return "An error caused the custom command to stop:\n`" + err.Error() + "`", nil
```
Also note that this won't format the error the same way as usual; might well be intended behavior but wanted to point it out anyway.

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+		}
+
+		return out, nil
+	},
+}
+
+var codeblockRegexp = regexp.MustCompile(`(?m)\A(?:\x60{2} ?\x60)(?:go(?:lang)?\n)?([\S\s]+)(?:\x60 ?\x60{2})\z`)
```
jo3-l, line 188: Maybe add a comment explaining why this admits spaces?

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
 	IsExecedByLeaveMessage bool `json:"is_execed_by_leave_message"`
 }
 
+var cmdEvalCommand = &commands.YAGCommand{
+	CmdCategory:  commands.CategoryTool,
+	Name:         "Evalcc",
+	Description:  "executes small custom command code",
```
jo3-l, line 123:
```suggestion
	Description:  "Executes custom command code (up to 1k chars)",
```

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+		}
+
+		channel := data.GuildData.CS
+		ctx := templates.NewContext(data.GuildData.GS, channel, data.GuildData.MS)
+		ctx.IsExecedByEvalCC = true
+
+		maxRunes := 500
+		if ctx.IsPremium {
+			maxRunes = 1000
+		}
```
jo3-l, line 162: IMO there's no real point in having a lower limit for non-premium users -- they can write much longer programs using the control panel anyway; this is just a minor inconvenience for no reason. WDYT?

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+			return "An error caused the custom command to stop:\n`" + errFormatted + "`", nil
+		}
+
+		return out, nil
+	},
+}
+
+var codeblockRegexp = regexp.MustCompile(`(?m)\A(?:\x60{2} ?\x60)(?:go(?:lang)?\n)?([\S\s]+)(?:\x60 ?\x60{2})\z`)
+
+// Parses code wrapped in Discord markdown codeblocks.
```
jo3-l, line 190:
```suggestion
// parseCodeblock returns content wrapped in a Discord markdown codeblock if needed.
```
Feel free to reword, but all doc comments should start with the function name.

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+	RequiredArgs: 1,
+	Arguments: []*dcmd.ArgDef{
+		{Name: "code", Type: dcmd.String},
+	},
+	SlashCommandEnabled: false,
+	DefaultEnabled:      true,
+	RunFunc: func(data *dcmd.Data) (interface{}, error) {
+		var hasCoreWriteRole bool
+
+		writeRoles := (common.GetCoreServerConfCached(data.GuildData.GS.ID)).AllowedWriteRoles
```
jo3-l, line 133:
```suggestion
		writeRoles := common.GetCoreServerConfCached(data.GuildData.GS.ID).AllowedWriteRoles
```

customcommands/bot.go
```diff
@@ -117,6 +117,91 @@ type DelayedRunCCData struct {
...
+	Name:         "Evalcc",
+	Description:  "executes small custom command code",
+	RequiredArgs: 1,
+	Arguments: []*dcmd.ArgDef{
+		{Name: "code", Type: dcmd.String},
+	},
+	SlashCommandEnabled: false,
+	DefaultEnabled:      true,
+	RunFunc: func(data *dcmd.Data) (interface{}, error) {
+		var hasCoreWriteRole bool
```
jo3-l, line 131:
```suggestion
		hasCoreWriteRole := false
```
IMO this is a little better as it makes the initial value (`false`) more obvious, but up to you.

customcommands/bot.go
l-zeuch: Excellent question. All *I* remember was that I copied that specific bit from elsewhere. I don't know where from, though.

customcommands/bot.go
l-zeuch: Yup. This is super old code (from before you introduced nicer-looking errors).
I'll update to use the new formatting.

customcommands/bot.go
l-zeuch: It doesn't really hurt to keep it to be honest, it'll only run if debug logging is enabled on that server anyway.

PR #1497, opened 2023-05-02
Added restrictions: topserver and topcommands should be botadmin only, other debug commands should be owner only. 

ashishjh-bst:


mrbentarikau:
maybe too late but still... as sidenote, if change is from admin to owner only, help description should be changed too, case by case

ashishjh-bst:
> Don't forget topcommands and topservers

@SoggySaussages that is already covered in the PR.

SoggySaussages:
> > Don't forget topcommands and topservers
> 
> 
> 
> @SoggySaussages that is already covered in the PR. 

Kewl, blame GitHub mobile cause not showing up for me :)

PR #1498, opened 2023-05-02
Replace transformer with basic string replacer

User9684:


User9684:
there wasn't a reopen button

User9684:
Don't merge, just realized that characters that have the diacritic built into it aren't changed

User9684:
> Don't merge, just realized that characters that have the diacritic built into it aren't changed

As far as I have seen, this was fixed and meets unicode specification

ashishjh-bst:
You'll have to explain in detail what all changes you're doing here, because your code doesn't make it obvious, I see that you're making a go file, but it doesn't explain for what?

PR #1502, opened 2023-05-16
Additional files for use on CapRover

waschinski:
When I was deploying **yagpdb** on my server running CapRover, I saw that the unofficial docker image mentioned in the docker compose file wasn't up-to-date.

So I came up with this little change which should be useful especially - but not solely - for CapRover users like me.

First change is a new Dockerfile for use with external https like when you are using a reverse proxy doing https for you.

Second change is a captain-definition file which basically just tells CapRover to use the new Dockerfile when building the image.

Both changes combined will allow a CapRover user to simply download the `.tar.gz` file from a new release and quickly decompress it before uploading it in order to deploy (an update) directly via Tarball.

### DIFF
--- a/captain-definition
+++ b/captain-definition
@@ -1,0 +1,4 @@
+ {
+  "schemaVersion" : 2,
+  "dockerfilePath" : "./yagpdb_docker/Dockerfile.exthttps"
+ }
--- a/yagpdb_docker/Dockerfile.exthttps
+++ b/yagpdb_docker/Dockerfile.exthttps
@@ -1,0 +1,26 @@
+FROM golang:stretch as builder
+
+WORKDIR /appbuild/yagpdb
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+WORKDIR /appbuild/yagpdb/cmd/yagpdb
+RUN CGO_ENABLED=0 GOOS=linux go build -v -ldflags "-X github.com/botlabs-gg/yagpdb/v2/common.VERSION=$(git describe --tags)"
+
+
+
+FROM alpine:latest
+
+WORKDIR /app
+VOLUME ["/app/soundboard"]
+EXPOSE 80 443
+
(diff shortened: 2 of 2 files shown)


User9684:
I feel like this should be more of an issue for the caprover quick deploy maintainers, rather than botlabs. 
If they directly supported caprover specific install methods, that means they would need to directly support a bunch of other different specific installation methods, which is just too much of a hassle

waschinski:
While I understand your point, I still kind of disagree.

- Quick deploy on CapRover is just for initial setup and also only useful if you want it to deploy the full stack. Want to use an already running Redis or PostgreSQL? Well, you are out of luck.
- Once `yagpdb` is running with docker there's currently no easy way to update it. There's no official docker image and while there are a couple images on Docker Hub they are either outdated or only for a specific arch and have most likely just used the default Dockerfile to build. Which brings me to the next point...
- The new additional Dockerfile is not CapRover specific but useful for everyone not using internal https. Usually in a Docker setup you will be running a reverse proxy in front of `yagpdb`which will be doing https.
- A captain-definition file must be in the root of a project in order to be useful for CapRover. But all it does is point to a Dockerfile so there really isn't anything to maintain or support here.

Honestly, the better solution would be having official docker images for both internal https and external https. Until these exist though, these small additions would be helpful not only for CapRover users but people installing `yagpdb` with docker.

lemm-e:
I'd highly urge everyone to continue this  conversation over at the [YAGPDB server](https://discord.gg/yagpdb) in the self-hosting channel to keep a better track and also gain further insight from other self-hosters.

ashishjh-bst:
This is highly specific to CapRover, If we add for CapRover, we'll have to accept PR for every PaaS. 
it's better suited to maintain in your fork.

waschinski:
I tried to explain how it's not highly specific to CapRover but anyways, thanks for taking the time to look at my PR.

PR #1504, opened 2023-05-20
disable feed if webhook for feed belongs to a forum channel

ashishjh-bst:


ashishjh-bst:
This additionally requires disabling forum channels from channel selectors of all possible places in UI where the bot needs to write, but that will be a bigger PR which touches a lot of ares on the dashboard, so for now this.

PR #1527, opened 2023-07-07
stdcommands: improve 8ball command

l-zeuch:
Remove the AYLIEN plugin, because it is broken anyway.
The API is either dead or inaccessable, so there is no point in keeping
it around.

Secondly, re-add the 8ball command formerly under `aylien/` to stdcommands.
Improve it by actually having more answers than a yes or no as it formerly
was due to AYLIEN being broken.


mrbentarikau:
it would be an august move to post eightball code as well : )

l-zeuch:
> it would be an august move to post eightball code as well : )

Haha, yeah. Forgot to add these. Pushing now.

PR #1528, opened 2023-07-09
stdcommands/customembed: allow codeblock and YAML

l-zeuch:
Allow codeblocks and YAML to be used in the `customembed` command.

Users will now be able to invoke the command with a codeblock wrapped
around their JSON or YAML, which makes writing and reading said data
quite a lot easier.

This change is fully backwards-compatible.

Here's an example of the new usage:

2023-07-09-130747_473x793_scrot

----
This is a combination of two commits:

* common/util: export ParseCodeblock func
    Export the ParseCodeblock function in common util, moving it out
    from its unexported location in customcommands.

    We'll be using this later for the customembed command and honestly
    it should have been exported from the get-go.

* stdcommands/customembed: allow codeblock and YAML
    Allow Discord codeblocks to be used in the customembed command. For
    bonus points, also allow YAML to be given instead, falling back to
    JSON as necessary.

    This was a suggestion on the support server with quite a lot of
    upvotes (understandably so), which you can find here:
    https://canary.discord.com/channels/166207328570441728/356486960417734666/868190676347658260

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -654,3 +654,22 @@ func FormatList(list []string, conjunction string) string {
 func BoolToPointer(b bool) *bool {
 	return &b
 }
+
+// Also accepts spaces due to how dcmd reconstructs arguments wrapped in triple backticks.
+var codeblockRegexp = regexp.MustCompile(`(?m)\A(?:\x60{2} ?\x60)(?:.*\n)?([\S\s]+)(?:\x60 ?\x60{2})\z`)
+
+// parseCodeblock returns the content wrapped in a Discord markdown block.
+// If no (valid) codeblock was found, the given input is returned back.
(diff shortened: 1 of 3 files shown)


User9684:
Another idea is you could also accept files. Just an idea though.

l-zeuch:
> Another idea is you could also accept files. Just an idea though.

That would require YAGPDB to download said file, which we aren't going
to do. Multimedia handling is expensive at this scale, no matter how
small a single file might be.

User9684:
> > Another idea is you could also accept files. Just an idea though.
> 
> That would require YAGPDB to download said file, which we aren't going to do. Multimedia handling is expensive at this scale, no matter how small a single file might be.

Fair enough

PR #1537, opened 2023-07-16
moderation: add mute and unmute cc functions

l-zeuch:
Add muteUser and unmuteUser functions as convenience functions for
moderation to the custom command engine.

### DIFF
--- a/moderation/tmplextensions.go
+++ b/moderation/tmplextensions.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


l-zeuch:
This is a draft for now, as I'd like to refactor out
the common code in both of these functions.

l-zeuch:
After some discussion in the dev channel, I've decided to close this as
**not planned**. Muting and suchlike are really only a oneliner of code.
(Even though it hurts me physically to waste an exec call for this)

If there are cases where more than just a single line of code would be
necessary, I'm going to PR such a convenience function.

PR #1539, opened 2023-07-17
Replacing owlbot with dictionaryapi.dev, as owlbot is discontinuing service from 24th July

ashishjh-bst:
Screenshot 2023-07-17 at 7 19 35 PM


mrbentarikau:
definitely a better dictionary than Owl, because it actually knows what "cookie" is : )

PR #1545, opened 2023-07-23
reminders: Add privacy and cleaner empty reminder response

SoggySaussages:
Based on unofficial suggestion here https://discord.com/channels/166207328570441728/525535451839463424/1132563547864440922 This pr changes some reminder commands to give ephemeral responses when used with the slash command. `reminders`, `creminders`, and `delreminder` are now ephemeral to respect user privacy. The `remindme` command was intentionally left out, hopefully to remind users that their reminders can not, and will not be ephemeral. I considered adding an additional notice to the `reminders` command illustrating this, would love to hear thoughts on the matter.

Additionally, it adds a response to `reminders` and `creminders` for when no reminders are found, as opposed to the current version which sends the prefixed and affixed messages with empty space in between.

Draft because I cannot test w/ my self host for a few days, want to double check for any unintended consequences.

### DIFF
--- a/reminders/plugin_bot.go
+++ b/reminders/plugin_bot.go
@@ -107,7 +107,7 @@ var cmds = []*commands.YAGCommand{
 			}
 
 			out := "You have no reminders. Create reminders with the `Remindme` command."
-			if currentReminders != nil {
+			if len(currentReminders) > 0 {
 				out = "Your reminders:\n"
 				out += stringReminders(currentReminders, false)
 				out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above.\nTo clear all reminders, use `delreminder` with the `-a` switch."
--- a/reminders/plugin_bot.go
+++ b/reminders/plugin_bot.go
@@ -90,7 +90,7 @@ var cmds = []*commands.YAGCommand{
 				return nil, err
 			}
 
-			return "Set a reminder in " + durString + " from now (<t:" + tUnix + ":f>)\nView reminders with the `Reminders` command", nil
+			return "Set a reminder in " + durString + " from now (<t:" + tUnix + ":f>)\nView reminders with the `reminders` command", nil
 		},
 	},
 	{
@@ -106,7 +106,7 @@ var cmds = []*commands.YAGCommand{
 				return nil, err
 			}
 
-			out := "You have no reminders. Create reminders with the `Remindme` command."
+			out := "You have no reminders. Create reminders with the `remindme` command."
 			if len(currentReminders) > 0 {
 				out = "Your reminders:\n"
 				out += stringReminders(currentReminders, false)
--- a/reminders/plugin_bot.go
+++ b/reminders/plugin_bot.go
@@ -137,7 +137,7 @@ var cmds = []*commands.YAGCommand{
 				return nil, err
 			}
 			out := "There are no reminders in this channel."
-			if currentReminders != nil {
+			if len(currentReminders) > 0 {
 				out = "Reminders in this channel:\n"
 				out += stringReminders(currentReminders, true)
 				out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above"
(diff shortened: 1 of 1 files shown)


reminders/plugin_bot.go
```diff
@@ -131,10 +136,12 @@ var cmds = []*commands.YAGCommand{
 			if err != nil {
 				return nil, err
 			}
-
-			out := "Reminders in this channel:\n"
-			out += stringReminders(currentReminders, true)
-			out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above"
+			out := "There are no reminders in this channel."
+			if currentReminders != nil {
```
l-zeuch, line 140:
```suggestion
			if len(currentReminders) > 0 {
```

reminders/plugin_bot.go
```diff
@@ -99,15 +99,19 @@ var cmds = []*commands.YAGCommand{
...
 			currentReminders, err := GetUserReminders(parsed.Author.ID)
 			if err != nil {
 				return nil, err
 			}
 
-			out := "Your reminders:\n"
-			out += stringReminders(currentReminders, false)
-			out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above.\nTo clear all reminders, use `delreminder` with the `-a` switch."
+			out := "You have no reminders. Create reminders with the `Remindme` command."
+			if currentReminders != nil {
```
l-zeuch, line 110:
```suggestion
			if len(currentReminders) > 0 {
```

l-zeuch (changes requested):
Output of reminders is currently not working as intended. The reason
being, that the `currentReminders` slice is empty, not nil, thus your
`if currentReminders != nil` check will always be true.

The difference between a nil slice and an empty slice is that the former
has no underlying data -- it points to *nothing*.
It usually is declared as `var s1 []string`.

An empty slice is declared as `s2 := []string{}` or `s3 := make([]string, 0)`.
These empty slices point to *something*: an empty, non-nil array.

SoggySaussages:
Okay cool, thank you for the thorough explanation, that makes sense.

reminders/plugin_bot.go
```diff
@@ -99,15 +99,19 @@ var cmds = []*commands.YAGCommand{
...
-			out := "Your reminders:\n"
-			out += stringReminders(currentReminders, false)
-			out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above.\nTo clear all reminders, use `delreminder` with the `-a` switch."
+			out := "You have no reminders. Create reminders with the `Remindme` command."
+			if len(currentReminders) > 0 {
+				out = "Your reminders:\n"
+				out += stringReminders(currentReminders, false)
+				out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above.\nTo clear all reminders, use `delreminder` with the `-a` switch."
+			}
 			return out, nil
```
jo3-l, line 115: Very minor suggestion, but I would personally write
```go
if len(currentReminders) == 0 {
	return "You have no reminders. Create reminders with the `Remindme` command.", nil
}

out := "Your reminders\n"
// ...
return out, nil
```
This avoids the reassignment to `out` on line 111,
```go
out = "Your reminders:\n"
```
which confused me on first read — I wasn't sure if it was intended or a typo. (Though it's clear after reading the text, the suggested structure avoids this problem entirely.)

Up to you, though. As I said, this is most minor.

reminders/plugin_bot.go
```diff
@@ -99,15 +99,19 @@ var cmds = []*commands.YAGCommand{
...
 		RunFunc: func(parsed *dcmd.Data) (interface{}, error) {
 			currentReminders, err := GetUserReminders(parsed.Author.ID)
 			if err != nil {
 				return nil, err
 			}
 
-			out := "Your reminders:\n"
-			out += stringReminders(currentReminders, false)
-			out += "\nRemove a reminder with `delreminder/rmreminder (id)` where id is the first number for each reminder above.\nTo clear all reminders, use `delreminder` with the `-a` switch."
+			out := "You have no reminders. Create reminders with the `Remindme` command."
```
jo3-l, line 109: You use capitals here (`Remindme`) but the existing text below uses lowercase (`delreminder`.) Choose one for consistency.

reminders/plugin_bot.go
SoggySaussages: Oh yes that does seem cleaner, sounds good

PR #1552, opened 2023-08-15
custom commands: Trim whitespace on response before character count validation

SoggySaussages:
This pr simply trims leading and trailing whitespace from custom command response before validating the character count. Since Discord would do this anyways, leading and trailing whitespace does not count towards it's 2000 character limit, and therefore imo should not throw an error for response length.

Of course it is still good practice for users to trim space within the code itself, especially when printing to the response within a `range` function, or even simply in multiple lines along the code.

I understand that this pr may encourage some users to learn bad habits and not trim their own whitespace when using `range`, however I believe it is already difficult for users to understand the concept of the `range` function generating whitespaces since, as aforementioned, Discord trims leading and trailing whitespace in the message, and the users are never actually able to see it. I feel like it is confusing that the following code runs fine, and produces a message with 6 characters:
Screenshot 2023-08-14 at 22 46 35

meanwhile this code, seemingly out of nowhere has a response length of >2k:
Screenshot 2023-08-14 at 22 46 46

But now with this pr:
Screenshot 2023-08-14 at 22 47 21

For these reasons, I believe that this pr will increase consistency and accuracy of the response length error, and these qualities are worth pursuing, even if it means handholding the user a little more.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -418,6 +418,9 @@ func (c *Context) newContextFrame(cs *dstate.ChannelState) *ContextFrame {
 func (c *Context) ExecuteAndSendWithErrors(source string, channelID int64) error {
 	out, err := c.Execute(source)
 
+	// trim whitespace for accurate character count
+	out = strings.TrimSpace(out)
+
 	if utf8.RuneCountInString(out) > 2000 {
 		out = "Template output for " + c.Name + " was longer than 2k (contact an admin on the server...)"
 	}
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -859,6 +859,9 @@ func ExecuteCustomCommand(cmd *models.CustomCommand, tmplCtx *templates.Context)
 	chanMsg := cmd.Responses[rand.Intn(len(cmd.Responses))]
 	out, err := tmplCtx.Execute(chanMsg)
 
+	// trim whitespace for accurate character count
+	out = strings.TrimSpace(out)
+
 	if utf8.RuneCountInString(out) > 2000 {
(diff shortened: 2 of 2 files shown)


common/templates/context.go
```diff
@@ -418,6 +418,9 @@ func (c *Context) newContextFrame(cs *dstate.ChannelState) *ContextFrame {
 func (c *Context) ExecuteAndSendWithErrors(source string, channelID int64) error {
 	out, err := c.Execute(source)
 
+	// trim whitespace for accurate character count
+	out = strings.TrimSpace(out)
```
ashishjh-bst, line 422: This would need changes on the UI with char counter.

ashishjh-bst (changes requested):

common/templates/context.go
SoggySaussages: Which changes? It doesn't impact the custom command editor at all, or the length of the code, it's only the output of executed commands before it's sent

ashishjh-bst: Oh alright, I assumed it was removing the whitespace before saving a command. My bad.

mrbentarikau:
following this code
```
{{ range seq 0 10000 }}
{{ $x := . }}
{{ end }}
HELLO!
```
would return error `Failed executing template: response grew too big (>25k)` so this is just a band aid for one problem and does not solve the issue with whitespaces.

SoggySaussages:
Oh I see, do you think a better solution would be not to write the leading or trailing whitespaces to the buffer in the first place? (I'm certainly not going to try it, I'll leave that to smarter people than me, just interested in your thoughts) Or do you think that it should just be left to the users to trim their own whitespace?

mrbentarikau:
I don't know, I was fine with the way it was... this change currently does not make anything better, just elevates the situation to another error.

PR #1553, opened 2023-08-15
Publish messages functions

SoggySaussages:
This pr adds message publishing functionality to custom commands. It adds two new functions, `publishMessage`, and `publishResponse`.

`publishMessage` channel messageID
Publishes a message by its ID in given announcement channel. channel can be either its ID, name or nil for triggering channel. Can be called once. This function will not work for leave or join messages.

`publishResponse`
Publishes the response when executed in an announcement channel. This function will not work for leave or join messages.

Both functions will validate the channel is an announcement channel, and check if the message is not already published (in the case of publishMessage) before attempting to publish. If either of these events are false, the function fails silently. This is intentional, but of course up for debate.

Discord's current rate limit for published messages is 10 per channel per hour, so it is disabled for fast feeds such as join and leave. Current behaviour is that the functions will stay defined in context, but throw an error on use, based on suggestion from Luca.

https://github.com/botlabs-gg/yagpdb/assets/110698921/58c11ce4-6299-4b40-a190-6e0e425a147f

https://github.com/botlabs-gg/yagpdb/assets/110698921/75ebe4c5-8265-4db2-a567-61de8aa56e7f

https://github.com/botlabs-gg/yagpdb/assets/110698921/d877afe0-d4e4-46bd-b311-c5da62ea748c


common/templates/context_funcs.go
```diff
@@ -508,6 +508,52 @@ func (c *Context) tmplPinMessage(unpin bool) func(channel, msgID interface{}) (s
 	}
 }
 
+func (c *Context) tmplPublishMessage(channel, msgID interface{}) (string, error) {
+	// Too heavily ratelimited by Discord to allow rapid feeds to publish
+	if c.IsExecedByLeaveMessage || c.IsExecedByJoinMessage {
+		return "", errors.New("you cannot publish messages with a feed")
```
jo3-l, line 514:
```suggestion
		return "", errors.New("cannot publish messages from a join/leave feed")
```

common/templates/context_funcs.go
```diff
@@ -508,6 +508,52 @@ func (c *Context) tmplPinMessage(unpin bool) func(channel, msgID interface{}) (s
...
+	}
+
+	_, err = common.BotSession.ChannelMessageCrosspost(cID, mID)
+	return "", err
+}
+
+func (c *Context) tmplPublishResponse() (string, error) {
+	// Too heavily ratelimited by Discord to allow rapid feeds to publish
+	if c.IsExecedByLeaveMessage || c.IsExecedByJoinMessage {
+		return "", errors.New("you cannot publish messages with a feed")
```
jo3-l, line 548: ditto

common/templates/context.go
```diff
@@ -174,6 +174,7 @@ type Context struct {
 	CurrentFrame *ContextFrame
 
 	IsExecedByLeaveMessage bool
+	IsExecedByJoinMessage  bool
```
jo3-l, line 177: This should probably be a single `ExecutedFrom` field at this point. Maybe leave this to a followup cleanup PR to address though.

notifications/plugin_bot.go
```diff
@@ -125,6 +125,7 @@ func sendTemplate(gs *dstate.GuildSet, cs *dstate.ChannelState, tmpl string, ms
 	ctx := templates.NewContext(gs, cs, ms)
 	ctx.CurrentFrame.SendResponseInDM = cs.Type == discordgo.ChannelTypeDM
 	ctx.IsExecedByLeaveMessage = !enableSendDM
+	ctx.IsExecedByJoinMessage = enableSendDM
```
jo3-l, line 128: Consider renaming `enableSendDM` to `isLeaveMsg`.

common/templates/context.go
SoggySaussages: I totally agree with all of that, I was actually thinking the same thing while I was working on this. I agree that it would be better addressed in a follow up PR.

ashishjh-bst:
Please rebase and resolve conflicts.

SoggySaussages:
Should be done now

PR #1555, opened 2023-08-18
lib/{dstate,discordgo}: add Banner to Guild struct

l-zeuch:
Add the Banner field to discordgo.Guild and dstate.GuildSet structs,
thus exposing them to custom commands.
Furthermore add a BannerURL method on dstate.GuildSet, returning the
appropriate URL (animated/static) to the Banner.

A relevant suggestion can be found here:
https://canary.discord.com/channels/166207328570441728/356486960417734666/992995802093981726

### DIFF
--- a/lib/discordgo/endpoints.go
+++ b/lib/discordgo/endpoints.go
@@ -39,6 +39,7 @@ var (
 	EndpointCDNIcons        string
 	EndpointCDNSplashes     string
 	EndpointCDNChannelIcons string
+	EndpointCDNBanners      string
 
 	EndpointAuth           string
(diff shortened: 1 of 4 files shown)


mrbentarikau:
dstate/helpers.go > `func GuildStateFromDgo(...` needs also updating

mrbentarikau:
another thing to consider with this field is - if guild does not have a banner, on Discord the bot just keeps writing without any return.

l-zeuch:
This behaviour is already present in custom commands that do not output anything either.

I'm not sure how this is suddenly something that needs consideration.

PR #1559, opened 2023-08-24
discordgo: correct default avatar URLs

jo3-l:
In introducing the new username system, Discord also switched to a different indexing scheme for default avatar URLs. In particular, per https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints:

> For users on the new username system, `index` will be `(user_id >> 22) % 6`.

This PR thus adjusts `User.AvatarURL` to account for this, aligning the output of `User.AvatarURL` with what is displayed on Discord.

### DIFF
--- a/lib/discordgo/user.go
+++ b/lib/discordgo/user.go
@@ -2,6 +2,7 @@ package discordgo
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/botlabs-gg/yagpdb/v2/lib/gojay"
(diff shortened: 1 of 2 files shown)


lib/discordgo/user.go
```diff
@@ -72,13 +73,25 @@ func (u *User) NKeys() int {
...
+		//
+		// "For users on the new username system, `index` will be `(user_id >> 22) % 6`.
+		// For users on the legacy username system, `index` will be `discriminator % 5`."
+		var index int
+		if u.Discriminator == "0" {
+			index = int((u.ID >> 22) % 6)
+		} else {
+			discrim, _ := strconv.Atoi(u.Discriminator)
+			index = discrim % 5
+		}
```
jo3-l, line 93: I considered factoring this bit out into a separate `DefaultAvatarIndex` method — notably, this is what upstream d.go does — but decided against it because the logic's fairly simple and there's no use that I can see for it elsewhere. Happy to change that if the separate method seems worthwhile.

PR #1563, opened 2023-08-27
web: display deleted channels in dropdowns

jo3-l:
Show deleted channels in dropdowns, and while at it, refactor the code generating the dropdown HTML for improved clarity as it was rather difficult to follow previously. (No visual changes other than the inclusion of deleted channels — see `outputDeletedChannels`, which has no counterpart in the existing code — have been made.) The refactor consists primarily of introducing a helper structure `channelOptsHTMLGenState` and then moving parts of the existing implementation into methods of this new structure. This provides much clearer organization of the various steps of the HTML generation process compared to before.

Not showing deleted channels (the current behavior) results in a difficult-to-diagnose visual bug, which can be reproduced as follows:

1. Create a new channel, say `#foo`.
2. Add a new custom command group and set it to only run in `#foo`.
3. Add a custom command to this group. Observe that it can be run in `#foo`, as expected.
4. Delete the channel `#foo`.
5. Observe that the custom command group settings misleadingly show that there are no channel restrictions (suggesting that custom commands within this group can be run anywhere), while in reality the custom command cannot be executed at all because it is restricted to operating in a deleted channel that is not shown in the UI.

### DIFF
--- a/web/template.go
+++ b/web/template.go
@@ -215,72 +215,99 @@ func tmplChannelOpts(channelTypes []discordgo.ChannelType) interface{} {
 	}
 }
 
-func tmplChannelOptsMulti(channelTypes []discordgo.ChannelType) func(channels []dstate.ChannelState, selections []int64) template.HTML {
+func tmplChannelOptsMulti(allowedChannelTypes []discordgo.ChannelType) func(channels []dstate.ChannelState, selections []int64) template.HTML {
 	return func(channels []dstate.ChannelState, selections []int64) template.HTML {
-
-		var builder strings.Builder
-
-		channelOpt := func(id int64, name string, channelType discordgo.ChannelType) {
-			builder.WriteString(`<option value="` + discordgo.StrID(id) + "\"")
-			for _, selected := range selections {
-				if selected == id {
-					builder.WriteString(" selected")
-				}
-			}
-			var prefix string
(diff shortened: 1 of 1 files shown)


web/template.go
```diff
@@ -215,74 +215,101 @@ func tmplChannelOpts(channelTypes []discordgo.ChannelType) interface{} {
...
-func tmplChannelOptsMulti(channelTypes []discordgo.ChannelType) func(channels []dstate.ChannelState, selections []int64) template.HTML {
+func tmplChannelOptsMulti(allowedChannelTypes []discordgo.ChannelType) func(channels []dstate.ChannelState, selections []int64) template.HTML {
 	return func(channels []dstate.ChannelState, selections []int64) template.HTML {
+		gen := &channelOptsHTMLGenState{allowedChannelTypes: allowedChannelTypes, channels: channels, selections: selections}
+		return gen.HTML()
+	}
+}
 
-		var builder strings.Builder
+type channelOptsHTMLGenState struct {
```
jo3-l, line 225: Open to suggestions for better names here — channelOptsHTMLGenState is quite a mouthful 😄

mrbentarikau:
already merged, bu `Deleted: %[1]` conserves screens space and because this is a channel dropdown, no need to tell it again.

PR #1566, opened 2023-09-04
lib/template: allow range over number

phenpessoa:
This PR adds the ability to range over a number (i.e. ints and uints) in template.

```
{{range $k, $v := 2}}{{$k}}{{$v}}{{end}}
```
Should print `0011`

```
{{range 2}}{{.}}{{end}}
```
Should print `01`

```
{{range 0}}{{else}}empty{{end}}
```
Should print `empty`

```
{{range -1}}{{else}}empty{{end}}
```
Should throw an error

### DIFF
--- a/lib/template/exec.go
+++ b/lib/template/exec.go
@@ -508,8 +508,12 @@ func (s *state) walkRange(dot reflect.Value, r *parse.RangeNode) controlFlowSign
 		return controlFlowNone
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		intVal := int(val.Int())
-		if intVal <= 0 {
+		if intVal == 0 {
 			break
+		}
+		if intVal < 0 {
+			s.errorf("range can't iterate over a negative number")
+			return controlFlowNone
 		}
 		for i := 0; i < intVal; i++ {
 			val := reflect.ValueOf(i)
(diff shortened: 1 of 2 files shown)


lib/template/exec.go
```diff
@@ -506,6 +506,36 @@ func (s *state) walkRange(dot reflect.Value, r *parse.RangeNode) controlFlowSign
 			break
 		}
 		return controlFlowNone
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		intVal := int(val.Int())
+		if intVal <= 0 {
+			break
+		}
```
jo3-l, line 513: I would rather raise a runtime error in this case. WDYT?

lib/template/exec.go
```diff
@@ -506,6 +506,36 @@ func (s *state) walkRange(dot reflect.Value, r *parse.RangeNode) controlFlowSign
...
+			case controlFlowReturnValue:
+				return controlFlowReturnValue
+			}
+		}
+		return controlFlowNone
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		intVal := int(val.Uint())
+		if intVal <= 0 {
+			break
+		}
```
jo3-l, line 528: ditto

lib/template/exec.go
phenpessoa: Sounds fine with me, but I think it might not be so user friendly if the value of `n` is a variable that the user does not know the value, like a count from DB
Perhaps `< 0` could be a runtime error but `= 0` could remain a simple break, WDYT?

jo3-l: I think `range 0` should be permitted, but negative is nonsensical.

phenpessoa: Updated

SoggySaussages:
Why is this necessary when we already have `seq`? You can get the same result in each of your examples by prefixing `seq 0` to each number. This method saves 5 characters for each call, is that the only difference? Would this confuse users and devalue the existing function?

phenpessoa:
It provides, imo, a better syntax. Also, it does not require allocating a slice, thus using less memory. 
But yes, this PR is basically syntactic sugar.

> Would this confuse users and devalue the existing function?

I don't think so. Seq still allows you to start from a value other than 0.

PR #1567, opened 2023-09-05
Increase the maximum time for a temporary role duration to 400 days

daira:
Motivation: We have a server for which we would like members to agree to a code of conduct in order to access most channels, and have them recommit to that agreement every 60 days. The role assignment and channel access control is working fine as per [this tutorial](https://docs.yagpdb.xyz/tools-and-utilities/self-assignable-roles). The simplest way to implement the 60-day requirement would be to use the "Temporary role duration" setting, however, that is limited to 1440 minutes (24 hours).

This PR changes the limit to 576000 minutes (400 days).

TemporaryRoleDuration is an `int` (at least 32 bits), and is used as `time.Duration(temporaryDuration)*time.Minute`. The underlying type of [`time.Duration`](https://pkg.go.dev/time#Duration) is an `int64` measured in nanoseconds, and [`time.Time`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/time/time.go;l=129) has a 33-bit seconds field, so this cannot introduce any possibility of overflow. The scheduled event is already persistent and should survive a server restart.

### DIFF
--- a/rolecommands/assets/rolecommands.html
+++ b/rolecommands/assets/rolecommands.html
@@ -272,7 +272,7 @@
                     <div class="row">
                         <div class="form-group col-lg-4">
                             <label for="group-temporary-role">Temporary roles (minutes)</label>
-                            <input type="number" min="0" max="1440" class="form-control" id="group-temporary-role"
+                            <input type="number" min="0" max="576000" class="form-control" id="group-temporary-role"
                                 name="TemporaryRoleDuration" value="{{.Group.TemporaryRoleDuration}}">
                             <p class="help-block">Remove roles in this group after a certain duration after assignment
                                 (0 to disable)</p>
--- a/rolecommands/web.go
+++ b/rolecommands/web.go
@@ -62,7 +62,7 @@ type FormGroup struct {
 
 	SingleAutoToggleOff   bool
 	SingleRequireOne      bool
-	TemporaryRoleDuration int `valid:"0,1440"`
+	TemporaryRoleDuration int `valid:"0,576000"`
 }
 
 func (p *Plugin) InitWeb() {


mrbentarikau:
If you have such big plans, it's way easier to selfhost the bot-code, also don't PR to master

ashishjh-bst:
Feel free to mod the source code and self-host, but this isn't necessary for every other server or self-host that uses the bot.

daira:
Fine, but:
* There is no documentation in your README saying what branch to target PRs to. Sure, I could have looked at other PRs, but it's an easy mistake to make for a new contributor.
* You could have just asked me to change the target branch.
* It's a two-line change, not "big plans". Obviously there is still a great deal of convenience in using the hosted service for us.
* It has zero impact on anyone else's use of the bot (and I put effort into checking absence of overflow in order to ensure that was the case).
* If we find this useful then others might. That's the whole point of contributing to an open source project.
* This kind of dismissive response discourages people from contributing to your project.

ashishjh-bst:
Hi Diara, 



> * It's a two-line change, not "big plans". Obviously there is still a great deal of convenience in using the hosted service for us.

- It is indeed a 2 line change, but it increases storage of the bot for scheduled events ( hence the recommendation to self-host).  Except mod events we restrict scheduled events to as low duration as possible, to prevent abuse and to reduce costs of storage. 

> *it has zero impact on anyone else's use of the bot (and I put effort into checking absence of overflow in order to ensure that was the case).

Our point exactly, a feature that only a single server will use doesn't belong in upstream, feel free to self-host and keep it in your personal repository, we have a suggestion mechanism in our support server where we evaluate feedback from the community on features and if something is really required we try to add it.   "I find it useful, others may too" is not a good benchmark for acceptance of pull requests. 

> * There's no apparent reason for the maximum to be restricted to 24 hours.

- There is, read point 1 above. 

> *  If we find this useful then others might. That's the whole point of contributing to an open source project.
-  Read point 2 above. 

>* This kind of dismissive response discourages people from contributing to your project.
I agree that sometimes people may come across as dismissive, we'll try to work on that, and make our guidelines better on what is acceptable as a feature and what is not. 

Thank you. 
I'll be locking this thread so that this doesn't turn into a typical internet argument.

PR #1568, opened 2023-09-08
add publish options for feeds

SoggySaussages:


SoggySaussages:
oops lol I leaned on the enter button

PR #1572, opened 2023-09-30
Publish message feed support

SoggySaussages:
This PR adds full support for YouTube and Streaming feeds to be published, either with a toggle or with `publishResponse`. This is an improvement over the current workaround where a user needs to `sendMessageRetID` and subsequently `publishMessage`. Using the toggle should work regardless of using custom or builtin message, and the toggle makes it clearer to users that the function is supported.

In addition, this pr merges three `IsExecedBy...` bool fields from template context into a single `ExecutedFrom` field.

### DIFF
--- a/common/mqueue/discordprocessor.go
+++ b/common/mqueue/discordprocessor.go
@@ -116,7 +116,6 @@ func trySendNormal(l *logrus.Entry, elem *QueuedElement) (err error) {
 		return
 	}
 
-	// Publish the announcement
 	if elem.PublishAnnouncement {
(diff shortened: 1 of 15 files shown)


ashishjh-bst:
Can you  add a screenshot of the UI please?

SoggySaussages:
Please see attached:
Screenshot 2023-10-04 at 23 04 39
Screenshot 2023-10-04 at 23 04 26
Screenshot 2023-10-04 at 23 04 14

ashishjh-bst:
I don't think this needs a dedicated toggle, and should be done by custom notification template using publishMessage or publishResponse.

SoggySaussages:
Cool, I've removed the toggle now

PR #1577, opened 2023-10-11
Money - Error Handling

Shadow21AR:
As Des said, handling all error like other commands do. (Just changed error to not give the response code and all things, just says unable to fetch data from API)


stdcommands/forex/forex.go
```diff
@@ -36,112 +38,108 @@ var Command = &commands.YAGCommand{
...
-	if resp.StatusCode != 200 {
-		return nil, commands.NewPublicError("HTTP err: ", resp.StatusCode)
+		return err
 	}
 	defer resp.Body.Close()
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, err
+
+	if resp.StatusCode != 200 {
```
l-zeuch, line 104: Should use `http.StatusOK` to avoid magic numbers.

Is the API guaranteed to always return `200 OK` on success, or are there
other `2xx` status codes it may return if the request is successful.

stdcommands/forex/forex.go
```diff
@@ -36,112 +38,108 @@ var Command = &commands.YAGCommand{
...
-			Footer:      &discordgo.MessageEmbedFooter{Text: fmt.Sprintf("Based on currency rate 1 : %f", output.Info.Rate)},
 			Timestamp:   time.Now().UTC().Format(time.RFC3339),
 		}
 		return embed, nil
 	},
 }
 
-func requestAPI(query string) (*ExchangeAPIResult, error) {
+func requestAPI(query string, result interface{}) error {
 	req, err := http.NewRequest("GET", query, nil)
```
l-zeuch, line 92: While we're at nitpicking some (minor) details:
```suggestion
	req, err := http.NewRequest(http.MethodGet, query, nil)
```

stdcommands/forex/forex.go
```diff
@@ -36,112 +38,108 @@ var Command = &commands.YAGCommand{
...
-		return nil, commands.NewPublicError("HTTP err: ", resp.StatusCode)
+		return err
 	}
 	defer resp.Body.Close()
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		return nil, err
+
+	if resp.StatusCode != 200 {
+		return commands.NewPublicError("Unable to fetch data from API")
```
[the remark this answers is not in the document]
Shadow21AR: Done

ashishjh-bst (approved):

PR #1580, opened 2023-10-17
frontend: move moderation pages to their own tab

l-zeuch:
Move subpages relevant to moderation to their own tab in order to
declutter the Tools & Utilities category on the control panel.

Seeing as Tools & Utilities is one of the largest cagetory on the
control panel, it can be viewed as a "catch-all", which is undesirable.

See also suggestion 2148.
https://discord.com/channels/166207328570441728/356486960417734666/1163826271231418380

### DIFF
--- a/automod/automod_web.go
+++ b/automod/automod_web.go
@@ -56,7 +56,7 @@ var (
 
 func (p *Plugin) InitWeb() {
 	web.AddHTMLTemplate("automod/assets/automod.html", PageHTML)
-	web.AddSidebarItem(web.SidebarCategoryTools, &web.SidebarItem{
+	web.AddSidebarItem(web.SidebarCategoryModeration, &web.SidebarItem{
 		Name: "Advanced Automoderator",
 		URL:  "automod",
 		Icon: "fas fa-robot",
--- a/automod_legacy/web.go
+++ b/automod_legacy/web.go
@@ -27,7 +27,7 @@ var panelLogKeyUpdatedSettings = cplogs.RegisterActionFormat(&cplogs.ActionForma
 func (p *Plugin) InitWeb() {
 	web.AddHTMLTemplate("automod_legacy/assets/automod_legacy.html", PageHTML)
 
-	web.AddSidebarItem(web.SidebarCategoryTools, &web.SidebarItem{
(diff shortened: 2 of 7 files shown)


l-zeuch:
I went with `fas fa-user-shield`, but obviously this is up for a change.
Just didn't want to spend too much time on deciding on an icon.

Attached is also a comparison before/after.

| Before | After |
|--------|--------|
| image | image |

ashishjh-bst:
Should verification go to Moderation too? ( I am not suggesting it should, I am unsure) 
Logging also only creates logs when moderation events happen, so should that move to moderation too?

l-zeuch:
> Should verification go to Moderation too? ( I am not suggesting it
> should, I am unsure)
> Logging also only creates logs when moderation events happen, so
> should that move to moderation too?

Not sure either. I went with the most obvious things first (i.e. the
ones with moderation in their name).
Verification makes probably sense, I think it is under Moderation in
Discord's server settings as well.

Logs might also make sense — even though they can also be created
on-demand, it is usually to initiate some sort of moderative action.

I'll wait for others to chime in on this as well and then make the
change if there are no founded objections.

l-zeuch:
Just pushed the changes, seeing as the functional equivalents of Logging
and Verification are listed under the Moderation tab in Discord's server
settings as well.

mrbentarikau:
you could split it same way to Automation section, Automoderators + Autorole. Question is, is it really this necessary to add yet another section?

l-zeuch:
> you could split it same way to Automation section, Automoderators + Autorole.

Would *also* make sense, yeah.

> Question is, is it really this necessary to add yet another section?

Pushing this a bit further: why not just get outright rid of this
catch-all category and sort things properly?

mrbentarikau:
Maybe keeping automated processes separate is even better, it perhaps prevents users thinking these can be manually controlled and embellishes set-it-forget-it approach more.

l-zeuch:
I'd like to wait for @ashishjh-bst to voice his opinion on this, too.
Personally I do agree with this sentiment; not to be a broken record,
but I'm not a huge fan of this sort-of catch-all category, so I'll
probably prepare some other patches that properly sort things on the
control panel.

I've got to do some Uni work now, but I'm watching this thread somewhat
closely.

PR #1583, opened 2023-10-29
Context menu commands + Delete DM command

User9684:
Implements context menu commands, and also adds the "Delete DM" context menu command.

### DIFF
--- a/bot/discordevents.go
+++ b/bot/discordevents.go
@@ -351,6 +351,10 @@ func HandleInteractionCreate(evt *eventsystem.EventData) {
 	if ic.User.ID == common.BotUser.ID {
 		return
 	}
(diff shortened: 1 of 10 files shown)


User9684:
There is a 90% chance that some changes are needed, so let me know if you spot anything (:

ashishjh-bst:
I think these changes are an overkill. 
Delete DM shouldn't be given to the end user, it can lead to the bot being ratelimited as all dm events have a stricter rate limi and all DM events happen on shard 0. 

I'll be closing this PR, but will be borrowing some things that have been modified here to improve command overwrites and to possibly apply them on slash commands too.

User9684:
I personally believe it would be better to expose it to the end user, but if all DM events do indeed run on shard 0 as you said, I do understand how that would lead to ratelimit issues.

Would a bot owner command that accepts a userID and an *optional* target message be acceptable?

PR #1587, opened 2023-11-13
feat(logs): Add login prompt for members only logs

Wolveric:
Users have continually expressed confusion regarding the error used
when attempting to open a message log for a server,
where logs are restricted to members only.
The current message implies the user isn't a member of the server,
even if they are.

This pull request attempts to rectify this, by adding a prompt
suggesting the user should use the provided login link, if they are a member.

### DIFF
--- a/logs/web.go
+++ b/logs/web.go
@@ -232,7 +232,11 @@ func CheckCanAccessLogs(w http.ResponseWriter, r *http.Request, config *models.G
 
 	member := web.ContextMember(ctx)
 	if member == nil {
-		tmpl.AddAlerts(web.ErrorAlert("This server has restricted log access to members only."))
+		goTo := url.QueryEscape(r.RequestURI)
+		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s>here</a>`, web.BaseURL(), goTo)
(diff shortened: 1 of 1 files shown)


ashishjh-bst (approved):

logs/web.go
```diff
@@ -230,7 +232,11 @@ func CheckCanAccessLogs(w http.ResponseWriter, r *http.Request, config *models.G
 
 	member := web.ContextMember(ctx)
 	if member == nil {
-		tmpl.AddAlerts(web.ErrorAlert("This server has restricted log access to members only."))
+		goTo := url.QueryEscape(r.RequestURI)
+		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s>here</a>`, web.BaseURL(), goTo)
+		alertMsg := fmt.Sprintf("This server has restricted log access to members only.\nIf you are a member, click %s to login.", alertLink)
```
l-zeuch, line 237: Consider increasing the length of the click-able area to be more
mobile-friendly.

```suggestion
		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s>click here</a>`, web.BaseURL(), goTo)
		alertMsg := fmt.Sprintf("This server has restricted log access to members only.\nIf you are a member, %s to login.", alertLink)
```

PR #1588, opened 2023-11-14
Implement custom status type

clari7744:
Implementing custom status type so that `-setstatus` will use custom status instead of Playing... now that it's available to bots
Flags have been added to permit the user to choose the presence type and online status of the bot
`GameType` has been renamed to `ActivityType`
`ActivityTypeCustom` and `ActivityTypeCompeting` have been added to the ActivityType struct
`UpdateStatus` now uses `ActivityTypeCustom` by default
`UpdatePlayingStatus`, `UpdateWatchingStatus`, and `UpdateCompetingStatus` have been added
`Update*Status` have had their parameters adjusted to allow the user to set online/idle/dnd/offline
Two new fields have been added to the Redis DB to store chosen presence type and online status.
Also thanks @l-zeuch for helping test and then helping me set up selfhost so I could test myself

### DIFF
--- a/bot/util.go
+++ b/bot/util.go
@@ -117,15 +117,15 @@ func SnowflakeToTime(i int64) time.Time {
 	return t
 }
 
-func SetStatus(gameType, statusText, statusType, streamingUrl string) {
+func SetStatus(activityType, statusType, statusText, streamingUrl string) {
 	if statusText == "" {
-		statusText = "v" + common.VERSION + " :)"
+		statusText = common.VERSION + " :)"
 	}
+	err1 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_activity_type", activityType))
+	err2 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_type", statusType))
+	err3 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_text", statusText))
+	err4 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_streaming_url", streamingUrl))
 
-	err1 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_streaming_url", streamingUrl))
-	err2 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_name", statusText))
-	err3 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_type", statusType))
-	err4 := common.RedisPool.Do(radix.Cmd(nil, "SET", "status_game_type", gameType))
 	if err1 != nil {
 		logger.WithError(err1).Error("failed setting bot status in redis")
 	}
@@ -275,29 +275,25 @@ func NodeID() string {
 
 // RefreshStatus updates the provided sessions status according to the current status set
 func RefreshStatus(session *discordgo.Session) {
-	var streamingUrl string
-	var statusText string
-	var gameType string
-	var statusTypeStr string
+	var activityType, statusTypeStr, statusText, streamingUrl string
 	var statusType discordgo.Status
-	err1 := common.RedisPool.Do(radix.Cmd(&streamingUrl, "GET", "status_streaming_url"))
-	err2 := common.RedisPool.Do(radix.Cmd(&statusText, "GET", "status_name"))
-	err3 := common.RedisPool.Do(radix.Cmd(&gameType, "GET", "status_game_type"))
-	err4 := common.RedisPool.Do(radix.Cmd(&statusTypeStr, "GET", "status_type"))
+	err1 := common.RedisPool.Do(radix.Cmd(&activityType, "GET", "status_activity_type"))
+	err2 := common.RedisPool.Do(radix.Cmd(&statusTypeStr, "GET", "status_type"))
+	err3 := common.RedisPool.Do(radix.Cmd(&statusText, "GET", "status_text"))
+	err4 := common.RedisPool.Do(radix.Cmd(&streamingUrl, "GET", "status_streaming_url"))
 
 	if err1 != nil {
-		logger.WithError(err1).Error("failed retrieving bot streaming status")
+		logger.WithError(err1).Error("failed retrieving bot activity type")
 	}
 	if err2 != nil {
-		logger.WithError(err2).Error("failed retrieving bot status")
+		logger.WithError(err2).Error("failed retrieving bot status type")
 	}
 	if err3 != nil {
-		logger.WithError(err3).Error("failed retrieving bot game type")
+		logger.WithError(err3).Error("failed retrieving bot status text")
 	}
 	if err4 != nil {
-		logger.WithError(err4).Error("failed retrieving bot status type")
+		logger.WithError(err4).Error("failed retrieving bot streaming url")
 	}
-
 	switch statusTypeStr {
 	case "online":
 		statusType = discordgo.StatusOnline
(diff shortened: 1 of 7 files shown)


stdcommands/setstatus/setstatus.go
l-zeuch, line 1: Perhaps explain the behaviour of the newly introduced switches,
outlining what they default to if no matching value was given.

You could use the `Description` field for that.

bot/util.go
```diff
@@ -265,26 +275,60 @@ func NodeID() string {
...
+			session.UpdatePlayingStatus(statusText, statusType)
+		case "listening":
+			session.UpdateListeningStatus(statusText, statusType)
+		case "watching":
+			session.UpdateWatchingStatus(statusText, statusType)
+		case "competing":
+			session.UpdateCompetingStatus(statusText, statusType)
+		default: // covers "custom"
+			session.UpdateStatus(statusText, statusType)
+		}
```
l-zeuch, line 327: Wouldn't it be easier to change the function signature of
`session.UpdateStatus` to also accept the `gameType`? That way we can
avoid this massive switch-case statement that doesn't do much else.

Though, we'd probably still have to have the variants we're calling here
elsewhere, probably unexported. In my opinion the user of this library
(even if it's just us) should not have to worry about calling the proper
function, we can handle that for them.

bot/util.go
```diff
@@ -265,26 +275,60 @@ func NodeID() string {
...
+		statusType = discordgo.StatusOnline
+	case "idle":
+		statusType = discordgo.StatusIdle
+	case "dnd":
+		statusType = discordgo.StatusDoNotDisturb
+	case "offline":
+		statusType = discordgo.StatusOffline
+	default:
+		statusType = discordgo.StatusOnline
+	}
```
l-zeuch, line 312: I feel... weird about this `statusTypeStr` versus `statusType`, however
I cannot quite pinpoint *why* I feel weird about this part.

It probably is fine:tm:, though.

l-zeuch (commented):
Explicitly neither approving nor requesting changes, as my comments are
more general feedback and thoughts than an actual review.

bot/util.go
clari7744: Are you saying move the switch-case to `UpdateStatus`? Could do that, was just following the format that was already there (which, admittedly, was only streaming or not streaming)

l-zeuch: > Are you saying move the switch-case to `UpdateStatus`? Could do that,
> was just following the format that was already there (which,
> admittedly, was only streaming or not streaming)

Yeah, I think that'd be easier to follow.

bot/util.go
clari7744: You probably feel weird about it because it means there are two variables that are basically the same except for one is a string and the other is an enum value. I'm not really sure what the best way to convert is, or even if it's necessary to convert, I did it here mostly for precision later when it gets passed around, but it certainly could do with just a single check and then pass the string around instead of the enum value 🤷‍♀

stdcommands/setstatus/setstatus.go
clari7744: I added descriptions to the switch `Help` fields, but you're right, the actual command description is a bit out of date.
Should I put the switch defaults in switch help or command help?

l-zeuch: I think switch help is fine and a tad clearer.

l-zeuch:
Looks OK to me now.

stdcommands/setstatus/setstatus.go
```diff
@@ -12,17 +12,40 @@ var Command = &commands.YAGCommand{
...
+		streamingUrl := data.Switch("url").Str()
+		switch activityType {
+		case "playing":
+		case "streaming":
+		case "listening":
+		case "watching":
+		case "custom":
+		case "competing":
+		default:
+			activityType = "custom"
```
jo3-l, line 38: Why not issue an error for unknown activities to catch typos such as `wtching` instead of silently defaulting to `custom`? Users can still set custom status via passing `custom` directly.

stdcommands/setstatus/setstatus.go
```diff
@@ -12,17 +12,40 @@ var Command = &commands.YAGCommand{
...
+		statusType := data.Switch("status").Str()
+		statusText := data.Args[0].Str()
+		streamingUrl := data.Switch("url").Str()
+		switch activityType {
+		case "playing":
+		case "streaming":
+		case "listening":
+		case "watching":
+		case "custom":
+		case "competing":
```
jo3-l, line 36: Fairly sure this works:
```suggestion
		case "playing", "streaming", "listening", "watching", "custom", "competing":
			// do nothing
```

stdcommands/setstatus/setstatus.go
```diff
@@ -12,17 +12,40 @@ var Command = &commands.YAGCommand{
...
+		default:
+			activityType = "custom"
+		}
+		switch statusType {
+		case "online":
+		case "idle":
+		case "dnd":
+		case "offline":
+		default:
+			statusType = "online"
```
jo3-l, line 46: Ditto

lib/discordgo/gateway.go
```diff
@@ -797,52 +797,93 @@ func (g *GatewayConnection) Close() error {
...
 
 // UpdateStatus is used to update the user's status.
-// If idle>0 then set status to idle.
-// If game!="" then set game.
-// if otherwise, set status to active, and no game.
-func (s *Session) UpdateStatus(idle int, game string) (err error) {
-	return s.UpdateStatusComplex(*newUpdateStatusData(idle, GameTypeGame, game, ""))
+// Set the custom status to statusText.
+// Set the online status to statusType.
+func (s *Session) UpdateStatus(activityType string, statusType Status, statusText, streamingUrl string) (err error) {
```
jo3-l, line 823: The presence of `streamingUrl` is a sign of bad API design IMO -- if I want to set playing status, I still need to pass an empty string. So I'm not convinced this function is appropriate. WDYT?

jo3-l: ...that said perhaps this is a necessary evil (and it's not too bad anyway) if we want to avoid code duplication. That said, if we're keeping this function I'd suggest splitting `string -> ActivityType` into one function `ParseActivityType` and then change `UpdateStatus` to accept an `ActivityType` instead of a `string`.

stdcommands/setstatus/setstatus.go
clari7744: Uh...not sure of the best way to error 😅 can reference other commands though

stdcommands/setstatus/setstatus.go
clari7744: Oh cool, don't know much about Go so that's good to know lol.

lib/discordgo/gateway.go
clari7744: > The presence of `streamingUrl` is a sign of bad API design IMO -- if I want to set playing status, I still need to pass an empty string. So I'm not convinced this function signature is appropriate. WDYT?

In this case, since it is the top-level function that's supposed to handle deciding for the user, IMO this is the simplest way

> ...that said perhaps this is a necessary evil (and it's not too bad anyway) if we want to avoid code duplication.
yeah this

> That said, if we're keeping this function I'd suggest splitting `string -> ActivityType` into one function `ParseActivityType` and then change `UpdateStatus` to accept an `ActivityType` instead of a `string`.
will do :+1:

clari7744: Wait why? Right now it's not being converted because it decides the function right away, wouldn't it just be extra steps to convert to ActivityType first? Or are you recommending to remove `s.Update{Type}Status` entirely, and have `UpdateStatus` make the call to `UpdateStatusComplex` right away?

vincor-qc: Sure, either would work.

clari7744: Alright, I separated out that conversion into another function, did leave the `s.Update{Type}Status` in place as well though.

ashishjh-bst:
Please attach screenshots of command usage and how it looks.

clari7744:
image
image
image

PR #1591, opened 2023-12-18
customcommands: add Size field to LightDBEntry

l-zeuch:
Add (approximate) Size to the LightDBEntry struct, set to the length of
the raw byte value.

It should be noted that, although this reading may be inaccurate,
internally we already limit the raw byte value to 100_000 in length;
thus, this change (and by extension this reading) is consistent with how
entry size is already determined.

As an added bonus, and because it was requested in
botlabs-gg/yagpdb#1590, I've also added the size reading to the therein
introduced view page.

I believe to recall that such a thing was either suggested in
`#suggestions`, or that it was requested in a drive-by comment in one of
the help channels; but I cannot find either.

### DIFF
--- a/customcommands/assets/customcommands_database.html
+++ b/customcommands/assets/customcommands_database.html
@@ -115,7 +115,7 @@
                         <td>{{.UserID}}</td>
                         <td>{{.Key}}</td>
                         <td>{{.Value}}</td>
-                        <td>{{.Size}}</td>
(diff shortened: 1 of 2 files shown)


mrbentarikau:
.ValueSize would be more in the context of what size it is

l-zeuch:
Fair enough. I've pushed a fixup just now.

PR #1599, opened 2024-01-04
Add Motivation command

savage4618:
I got a container of motivational quotes for Christmas, working in automation, I decided I should find a way to randomly access one. That's when I remembered I had a self-hosted yag bot and I could add a command to give me a random motivational message. This is that command.


mrbentarikau:
Do such things achievable with custom commands really need a stdcommand? For slash?

savage4618:
> Do such things achievable with custom commands really need a stdcommand? For slash?

I don't have a defense for this, tbh. Yes, I could have just done a custom command, but could the same be argued for roasts? throw? catfact? dogfact?
I guess I just wanted to share.

savage4618:
gonna change this up.

PR #1600, opened 2024-01-04
Added Trivia

ashishjh-bst:
Sends a random question from opentdb.com, and records all answers sent within 10 seconds, and shows the list of winners. 

Screenshot 2024-01-05 at 12 56 32 AM


mrbentarikau:
10 seconds is too short, sometimes it could take more than 10 secs to even read the trivia.

ashishjh-bst:
> 10 seconds is too short, sometimes it could take more than 10 secs to even read the trivia.
> 
> also I'd add Cooldown to trivia command - there's already channel session cooldown, but general cooldown to prevent spamming it all around the channels.

Thanks, I have increased duration to 30 seconds, will also play around later with CAH like multiplayer sessions and a scoreboard.

PR #1604, opened 2024-01-21
Add Basic Thread Management

KTStephano:
This update does 2 things:
* Updates the discordgo restapi to include support for creation and management Threads
* Adds `createThread`, `deleteThread`, `addThreadMember`, `removeThreadMember` which work in custom command templates

Other info:
* After discussion in comments, forum posts aren't supported (later PR with separate set of functions)
* deleteThread on a regular channel is rejected (only active threads considered valid)
* createThread is set to 1/1 (free/premium) and deleteThread also set to 1/1 (free/premium) per CC. Both are separate counters. I do not know what good numbers would be for these limits.

API:

Examples (Create):

    {{$channel := ...}}
    {{$thread := createThread $channel nil "YAG Public Thread"}}
    {{$thread = createThread $channel nil "YAG Private Thread" true}}

    {{$messageID := ...}}
    {{$msgThread := createThread $channel $messageID "Message Thread"}}

Example (Delete):

    {{deleteThread $thread.ID}}

Example (Add):

    {{addThreadMember (index .CmdArgs 0) (index .CmdArgs 1)}}

Example (Remove):

    {{removeThreadMember (index .CmdArgs 0) (index .CmdArgs 1)}}


jo3-l:
WDYT about using the names `addThreadMember` and `removeThreadMember` instead?

KTStephano:
> WDYT about using the names `addThreadMember` and `removeThreadMember` instead?

That sounds good to me. Should I go ahead and update this PR with the change?

KTStephano:
`createThread` has been set to initial max 1 normal, 2 premium per CC
`deleteThread` has been set to initial max 1 normal, 2 premium per CC
(both separate limits)

No idea what values these could or should be

mrbentarikau:
change PR's destination from master branch to dev branch.

common/templates/context.go
```diff
@@ -657,6 +657,11 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("getPinCount", c.tmplGetChannelPins(true))
 	c.addContextFunc("getRole", c.tmplGetRole)
 	c.addContextFunc("getThread", c.tmplGetThread)
+	c.addContextFunc("createThread", c.tmplCreateThread)
+	c.addContextFunc("createMessageThread", c.tmplCreateMessageThread)
```
ashishjh-bst, line 661: maybe createThread can take a msgID and if it is nil, it will create a non-message thread, the way they both work otherwise is same.

Ranger-4297: Speaking on forumTags. I made a PR that ended up being closed due to activity a while bit ago. The PRs aim was to implement setting tags via the same premise that `setRoles` did as well as retrieve the current forums tags. It was albeit not a perfect way of going about it due to my little experience with Go but if you do need a starting point it's there

KTStephano: Thanks I'll find that PR and look through it. Maybe it'll give me some ideas

ashishjh-bst: Here: https://github.com/botlabs-gg/yagpdb/pull/1453

mrbentarikau:
complexThread argument `is_private` is not really necessary, simple `private` is enough

KTStephano:
Good idea, I changed it to just `private`

I couldn't find a way to send the content to non-Forum threads with discordgo. To get around it I added a check for content != nil for non-Forum threads. If that's the case the bot creates the thread and then immediately sends a followup message into it

mrbentarikau:
maybe in more general sense it would be better to keep forum stuff separate from basic threads? 
what do others think...

also, just side note, it is possible to
```
...
switch val := val.(type) {
case *discordgo.MessageSend:
    thread.Content = val
...
```
this does type assertion on the fly and no need for `thread.Content = val.(*discordgo.MessageSend)`

common/templates/context.go
KTStephano: Pushed changes that bring in some of the ideas/feedback from #1453 . `complexThread` now supports a tags input. When a forum thread with tags is created, the returned templates.CtxChannel.AppliedTags is populated with a list of discordgo.ForumTag.

KTStephano:
> maybe in more general sense it would be better to keep forum stuff separate from basic threads? what do others think...

Are you thinking createThread would be set to just ignore `content` for basic threads? Or are you thinking more along the lines of having a separate function to create forum stuff

mrbentarikau:
My opinion is yeah to keep thread and forum functions separate, or at least the complex types, because tags, invitable, private mix between threads/forum posts could get overwhelming and confusing to users.

lib/dstate/interface.go
```diff
@@ -263,6 +263,12 @@ type ChannelState struct {
 	ThreadMetadata   *discordgo.ThreadMetadata `json:"thread_metadata"`
 
 	PermissionOverwrites []discordgo.PermissionOverwrite `json:"permission_overwrites"`
+
+	// The set of tags that can be used in a forum channel.
+	AvailableTags []discordgo.ForumTag `json:"available_tags"`
+
+	// The IDs of the set of tags that have been applied to a thread in a forum channel.
+	AppliedTags []discordgo.ForumTag `json:"applied_tags"`
```
mrbentarikau, line 271: This part is little confusing, what is meant to be used here, slice of IDs or a slice of full ForumTag. Discord documentation states AppliedTags is an array of tag ID snowflakes not full objects

KTStephano: Yeah discordgo is supposed to return []string with I think the snowflakes. I have the code manually upgrading them to full ForumTags by walking the AvailableTags slice. We could downgrade it to []int64 if that seems better

mrbentarikau: I'd use IDs as int64 all around, even in AppliedTags slices - so here I'd use a helper function to convert from names users add to `complexThread "tags" (cslice ... )` as names to IDs.

But developer input needed, just my personal preference.

lib/discordgo/structs.go
```diff
@@ -314,6 +320,24 @@ type AddedThreadMember struct {
...
+	EmojiName string `json:"emoji_name,omitempty"`
+}
+
+// ForumTag represents a tag that is able to be applied to a thread in a forum channel.
+type ForumTag struct {
+	ID        string `json:"id,omitempty"`
+	Name      string `json:"name"`
+	Moderated bool   `json:"moderated"`
+	EmojiID   string `json:"emoji_id,omitempty"`
+	EmojiName string `json:"emoji_name,omitempty"`
```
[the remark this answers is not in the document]
KTStephano: Latest push moves ForumTag to use int64 ids. I tried moving the tag id slice in Channel from []string to []int64 and ran into json marshal/unmarshal errors so I skipped that one for now.

KTStephano:
> My opinion is yeah to keep thread and forum functions separate, or at least the complex types, because tags, invitable, private mix between threads/forum posts could get overwhelming and confusing to users.

Trying to think of a good way to split it up. Any feedback would be great. Does this direction seem any better?

`createThread channel thread` -> returns *templates.CtxChannel
> thread type requirement = *templates.CtxThreadStart

`thread name private` -> returns *templates.CtxThreadStart
`messageThread name messageID` -> returns *templates.CtxThreadStart
`forumThread name content tags...` -> returns *templates.CtxThreadStart
> tags = optional list of names

Usage:
{{createThread channel (thread ...)}}
{{createThread channel (messageThread ...)}}
{{createThread channel (forumThread ...)}}

ashishjh-bst:
> My opinion is yeah to keep thread and forum functions separate, or at least the complex types, because tags, invitable, private mix between threads/forum posts could get overwhelming and confusing to users.

I agree with this, `createForumPost` and `removeForumPost` would make it easier for the end user, and also allow us to remove forum related validations from `createThread`

ashishjh-bst:
> > My opinion is yeah to keep thread and forum functions separate, or at least the complex types, because tags, invitable, private mix between threads/forum posts could get overwhelming and confusing to users.
> 
> Trying to think of a good way to split it up. Any feedback would be great. Does this direction seem any better?
> 
> `createThread channel thread` -> returns *templates.CtxChannel
> 
> > `thread` type requirement = *templates.CtxThreadStart
> 
> `basicThread name private` -> returns *templates.CtxThreadStart `messageThread name messageID` -> returns *templates.CtxThreadStart `forumThread name content tags...` -> returns *templates.CtxThreadStart
> 
> > `tags` = optional slice of tag names
> 
> Usage: {{createThread channel (basicThread ...)}} {{createThread channel (messageThread ...)}} {{createThread channel (forumThread ...)}}

I think createThread should just work on regular channels, and forums should have their own set of templates instead of further complicating this. Forums also have extra properties unlike threads in regular channels.

KTStephano:
> I think createThread should just work on regular channels, and forums should have their own set of templates instead of further complicating this. Forums also have extra properties unlike threads in regular channels.

Ok, this makes sense to me. The latest changes remove `complexThread` and simplify createThread so that it can only handle threads in regular channels (basic thread and message threads).

I'll open a separate PR sometime after this one with templates dedicated to forum posts.

lib/discordgo/structs.go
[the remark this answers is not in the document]
mrbentarikau: for Channel.AppliedTags []int64 and others gojay unmarshaling is probably necessary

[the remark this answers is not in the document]
KTStephano: That's awesome, I'll use int64 and IDSlice for the forums PR (unless someone else PRs it first)

PR #1605, opened 2024-01-21
logs: don't create logs in threads of blacklisted channels

jo3-l:
Previously, if channel A was blacklisted, the logs command could still be used in subthreads of channel A (though not channel A itself.)

Thanks to wroclaw on Discord for reporting this bug.

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -23,12 +23,10 @@ import (
 	"golang.org/x/net/context"
 )
 
-
-
 var (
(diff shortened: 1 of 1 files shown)


mrbentarikau:
fix "seperator" in comments as well : )

mrbentarikau:
in logs/plugin_bot.go... return message "This channel is blacklisted from creating message logs..." needs maybe some fine-tuning, because thread needs to be clarified.... "This channel/thread's parent..." or something like that.

ashishjh-bst (approved):

PR #1609, opened 2024-01-25
Add Forum Post Support

KTStephano:
Continuation of #1604 and #1453 

This PR does two things:
* Updates discordgo, dstate with new fields related to forums (available tags, applied tags, default thread rate limits, etc.)
* Adds `createForumPost` and `deleteForumPost` templates

Other Info:
* Limits for create and remove are currently shared with thread's create and remove limits
* Optional args for `createForumPost` are set to use key-value style so that order doesn't matter and we can also expose other optional args later

API (templates):

Field | Description
--- | ---
.Channel.IsForum | true if channel is forum type
.Channel.AvailableTags | Slice of discordgo.ForumTag. Only present if channel is forum type.
.Channel.AppliedTags | Slice of int64 representing tag ids. Only present if channel represents a specific forum thread/post.

API (functions):

Examples:

(Create)

    {{$forumChannel := ...}}
    {{$postTitle := "Forum Post Title"}}
    {{$post := createForumPost $forumChannel $postTitle "post content as string"}}

    {{$tags = cslice "discussion" "bot" "info"}}
    {{$post = createForumPost $forumChannel $postTitle (cembed ...) "tags" $tags}}

    {{$oneMinute := 60}}
    {{$post = createForumPost $forumChannel $postTitle (cembed ...) "slowmode" $oneMinute}}

    {{$post = createForumPost $forumChannel $postTitle (cembed ...) "tags" $tags "slowmode" $oneMinute}}

(Delete)

    {{deleteForumPost $post.ID}}

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -657,10 +657,16 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("getPinCount", c.tmplGetChannelPins(true))
 	c.addContextFunc("getRole", c.tmplGetRole)
 	c.addContextFunc("getThread", c.tmplGetThread)
+
+	// thread functions
 	c.addContextFunc("createThread", c.tmplCreateThread)
 	c.addContextFunc("deleteThread", c.tmplDeleteThread)
 	c.addContextFunc("addThreadMember", c.tmplThreadMemberAdd)
 	c.addContextFunc("removeThreadMember", c.tmplThreadMemberRemove)
+
+	// forum functions
+	c.addContextFunc("createForumPost", c.tmplCreateForumPost)
+	c.addContextFunc("removeForumPost", c.tmplRemoveForumPost)
 
 	c.addContextFunc("currentUserAgeHuman", c.tmplCurrentUserAgeHuman)
 	c.addContextFunc("currentUserAgeMinutes", c.tmplCurrentUserAgeMinutes)
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -666,7 +666,7 @@ func baseContextFuncs(c *Context) {
 
 	// forum functions
 	c.addContextFunc("createForumPost", c.tmplCreateForumPost)
-	c.addContextFunc("removeForumPost", c.tmplRemoveForumPost)
+	c.addContextFunc("deleteForumPost", c.tmplDeleteThread)
 
 	c.addContextFunc("currentUserAgeHuman", c.tmplCurrentUserAgeHuman)
 	c.addContextFunc("currentUserAgeMinutes", c.tmplCurrentUserAgeMinutes)
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -1424,6 +1424,7 @@ func (c *Context) tmplCreateThread(channel, msgID interface{}, name string, priv
 	return CtxChannelFromCS(&tstate), nil
 }
 
+// This function can delete both basic threads and forum threads
 func (c *Context) tmplDeleteThread(thread interface{}) (string, error) {
 
 	if c.IncreaseCheckCallCounterPremium("delete_thread", 1, 1) {
@@ -1646,10 +1647,6 @@ func (c *Context) tmplCreateForumPost(channel interface{}, name string, content 
 	c.AddThreadToGuildSet(&tstate)
 
 	return CtxChannelFromCS(&tstate), nil
-}
-
-func (c *Context) tmplRemoveForumPost(thread interface{}) (string, error) {
-	return c.tmplDeleteThread(thread)
 }
 
 func (c *Context) tmplGetChannelOrThread(channel interface{}) (*CtxChannel, error) {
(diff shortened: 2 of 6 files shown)


common/templates/context.go
```diff
@@ -657,11 +657,17 @@ func baseContextFuncs(c *Context) {
...
+
+	// thread functions
 	c.addContextFunc("createThread", c.tmplCreateThread)
 	c.addContextFunc("deleteThread", c.tmplDeleteThread)
 	c.addContextFunc("addThreadMember", c.tmplThreadMemberAdd)
 	c.addContextFunc("removeThreadMember", c.tmplThreadMemberRemove)
 
+	// forum functions
+	c.addContextFunc("createForumPost", c.tmplCreateForumPost)
+	c.addContextFunc("removeForumPost", c.tmplRemoveForumPost)
```
mrbentarikau, line 669: `deleteForumPost` feels better (create > delete, add > remove, give > take logic), also no need for additional `c.tmplRemoveForumPost` why not just `c.tmplDeleteThread`

common/templates/context_funcs.go
```diff
@@ -1491,6 +1494,164 @@ func (c *Context) tmplThreadMemberRemove(threadID, memberID interface{}) string
...
+			}
+		default:
+			return 0, nil, errors.New(`invalid key "` + key + `"`)
+		}
+	}
+
+	return rateLimit, tags, nil
+}
+
+func (c *Context) tmplCreateForumPost(channel, content interface{}, name string, optional ...interface{}) (*CtxChannel, error) {
```
mrbentarikau, line 1586: To me workflow > channel, name, content, (optional...) feels better. Post title is sort of first than post content. I know it needs interface{} twice then : )

KTStephano: Latest commit switches the order to be channel, name, content, (optional...)

common/templates/context.go
KTStephano: That makes sense, I renamed it "deleteForumPost". I also removed tmplRemoveForumPost in favor of reusing tmplDeleteThread (added a comment about being it able to delete both types of threads like you suggested).

common/templates/context_funcs.go
KTStephano: I'm also wondering if `createThread` from the last PR should be changed to follow the same format. For example 
createThread channel name (optional: "message_id" arg "private" arg)

mrbentarikau: looks reasonable, to have thread's name first rest is already all content and how content is shown

lib/dstate/interface.go
```diff
@@ -263,6 +263,14 @@ type ChannelState struct {
...
 
 	PermissionOverwrites []discordgo.PermissionOverwrite `json:"permission_overwrites"`
+
+	AvailableTags []discordgo.ForumTag `json:"available_tags"`
+	AppliedTags   []int64              `json:"applied_tags"`
+
+	DefaultReactionEmoji          discordgo.ForumDefaultReaction `json:"default_reaction_emoji"`
+	DefaultThreadRateLimitPerUser int                            `json:"default_thread_rate_limit_per_user"`
+	DefaultSortOrder              *discordgo.ForumSortOrderType  `json:"default_sort_order"`
+	DefaultForumLayout            discordgo.ForumLayout          `json:"default_forum_layout"`
```
mrbentarikau, line 273: I think, for IsPrivate() method few lines down, `|| c.Type == discordgo.ChannelTypeGuildPrivateThread` needs to be added to `if`

common/templates/context_funcs.go
mrbentarikau: it think that parameter `name` could also be part of type interface{} and you later ToString() it, because in some cases it is easier to use pure numbers as thread name : )

KTStephano: Pushed a change that switches both create functions to use interface{} for name. It calls ToString() on it before passing it to discordgo.

lib/dstate/interface.go
KTStephano: Does GuildPrivateThread sit on the same level as DM privacy?

mrbentarikau: looking at how this is handled in context functions, well it is a badly named field. Channel Type 12 says yeah it's private as well. so better not change this.

lib/discordgo/structs.go
```diff
@@ -237,6 +263,27 @@ type Channel struct {
...
+	// This field is copied to the thread at creation time and does not live update.
+	DefaultThreadRateLimitPerUser int `json:"default_thread_rate_limit_per_user"`
+
+	// The default sort order type used to order posts in forum channels.
+	// Defaults to null, which indicates a preferred sort order hasn't been set by a channel admin.
+	DefaultSortOrder *ForumSortOrderType `json:"default_sort_order"`
+
+	// The default forum layout view used to display posts in forum channels.
+	// Defaults to ForumLayoutNotSet, which indicates a layout view has not been set by a channel admin.
+	DefaultForumLayout ForumLayout `json:"default_forum_layout"`
```
mrbentarikau, line 286: Perhaps these fields should be in templates/structs.go > CtxChannel as well

KTStephano: Yeah could add them. When I first went through it I just wasn't sure if people would need those fields in CCs

PR #1613, opened 2024-02-07
fix: thanks detection logic for blacklist/whitelist

zeing:
## Problem

When you add channels for 
**Whitelisted channels for giving rep using thanks** 
or 
**Blacklisted channels for giving rep using thanks**
in reputation in `/manage/server_id/reputation` page

Bot still allows in threads in blacklisted channels
Bot doesn't work in threads in whitelisted channels

### DIFF
--- a/reputation/plugin_bot.go
+++ b/reputation/plugin_bot.go
@@ -69,7 +69,7 @@ func handleMessageCreate(evt *eventsystem.EventData) {
 	}
 
 	// Check if thanks detection is allowed in the parent channel
-	if cState.Type.IsForum() || cState.Type.IsThread() {
+	if cState.Type.IsThread() {
 		if !isThanksDetectionAllowedInChannel(conf, cState.ParentID) {
 			return
 		}
--- a/reputation/plugin_bot.go
+++ b/reputation/plugin_bot.go
@@ -63,8 +63,20 @@ func handleMessageCreate(evt *eventsystem.EventData) {
 		return
 	}
 
-	if !isThanksDetectionAllowedInChannel(conf, msg.ChannelID) {
-		return
+	cState := evt.CSOrThread()
+	if cState == nil {
+		return // No channel state, ignore
+	}
+
+	// Check if thanks detection is allowed in the parent channel
+	if cState.Type.IsForum() || cState.Type.IsThread() {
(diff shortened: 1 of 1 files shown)


reputation/plugin_bot.go
```diff
@@ -63,8 +63,20 @@ func handleMessageCreate(evt *eventsystem.EventData) {
...
 
-	if !isThanksDetectionAllowedInChannel(conf, msg.ChannelID) {
-		return
+	cState := evt.CSOrThread()
+	if cState == nil {
+		return // No channel state, ignore
+	}
+
+	// Check if thanks detection is allowed in the parent channel
+	if cState.Type.IsForum() || cState.Type.IsThread() {
```
SoggySaussages, line 72: `cState.Type.IsForum()` returns true if the channel is a forum channel, not a forum post. Forum posts are public threads, so `IsThread` is all you need here I believe. Have you tested this?

zeing: No, I haven't tested with `cState.Type. IsThread()` only

image

Let me check

zeing: Yes, you're right, `IsThread` is all I need

reputation/plugin_bot.go
```diff
@@ -63,8 +63,20 @@ func handleMessageCreate(evt *eventsystem.EventData) {
...
-	if !isThanksDetectionAllowedInChannel(conf, msg.ChannelID) {
-		return
+	cState := evt.CSOrThread()
+	if cState == nil {
+		return // No channel state, ignore
+	}
+
+	// Check if thanks detection is allowed in the parent channel
+	if cState.Type.IsThread() {
+		if !isThanksDetectionAllowedInChannel(conf, cState.ParentID) {
```
ashishjh-bst, line 73: Why not just write it like this?  What you've written looks like a conditional mess.  
```
channelID := cState.ChannelID
if cState.Type.IsThread(){
   channelID := cState.ParentID
}
if !isThanksDetectionAllowedInChannel(conf, channelID) {
			return
}
```

ashishjh-bst (changes requested):
Refactor to make the code readable.

ashishjh-bst (approved):

PR #1616, opened 2024-02-19
Autocomplete

clari7744:
This pull requests implements autocomplete in YagCommand for slash command arguments.
It also implements autocomplete in `rolemenu create` as an example.

Disclaimer: I have no idea what I'm doing, this works because it was added to the middleware that hooks into every command. There's probably a proper way to do this with another function in `Cmd`, but I could not wrap my head around doing that properly without setting up another middleware thing. 
Very open to feedback, but this is at least a working prototype

image
image


commands/slashcommands.go
```diff
@@ -81,10 +81,10 @@ func (p *Plugin) updateGlobalCommands() {
 
 	logger.Info("Slash commands changed, updating....")
 
-	ret, err := common.BotSession.BulkOverwriteGlobalApplicationCommands(common.BotApplication.ID, result)
-	// ret, err := common.BotSession.BulkOverwriteGuildApplicationCommands(common.BotApplication.ID, 614909558585819162, result)
+	// ret, err := common.BotSession.BulkOverwriteGlobalApplicationCommands(common.BotApplication.ID, result)
+	ret, err := common.BotSession.BulkOverwriteGuildApplicationCommands(common.BotApplication.ID, 778487633205788712, result)
```
mrbentarikau, line 85: why is this value 778487633205788712 for guildID parameter there?

commands/plugin_bot.go
```diff
@@ -186,6 +186,23 @@ func YAGCommandMiddleware(inner dcmd.RunFunc) dcmd.RunFunc {
...
 
+		// Check if the user can execute the command
+		canExecute, resp, settings, err := yc.checkCanExecuteCommand(data)
+		if err != nil {
+			yc.Logger(data).WithError(err).Error("An error occured while checking if we could run command")
+		}
+
+		parseErr := dcmd.ParseCmdArgs(data)
+		// Stop here if it's autocomplete; we don't need to block it
+		if data.SlashCommandTriggerData.Interaction.Type == discordgo.InteractionApplicationCommandAutocomplete {
```
mrbentarikau, line 197: this will panic if data.SlasCommandTriggerData is nil... same thing for method Run in yagcommand.go

commands/slashcommands.go
clari7744: That was my test guild, whoops 😅 will remove

cah/commands.go
```diff
@@ -41,6 +42,17 @@ func (p *Plugin) AddCommands() {
 
 			return nil, err
 		},
+		AutocompleteFunc: func(data *dcmd.Data) ([]*discordgo.ApplicationCommandOptionChoice, error) {
```
mrbentarikau, line 45: This thing here needs a limiter, official doc says `Choices for STRING, INTEGER, and NUMBER types for the user to pick from, max 25` this is also same for ApplicaitonCommand.Options field....

I run into it because I have 27 CAH packs : )
but problem could be yeah even more common for rolemenu groups

mrbentarikau (approved):
Looks nice, I personally sorted CAH-pack choices alphabetically - matter of taste, also these debug fmt.Prints rolemenu's Autocomplete function has are not necessary : )

commands/plugin_bot.go
```diff
@@ -172,7 +172,7 @@ func YAGCommandMiddleware(inner dcmd.RunFunc) dcmd.RunFunc {
 			guildID = data.GuildData.GS.ID
 		}
 
-		if data.TriggerType == dcmd.TriggerTypeSlashCommands {
+		if data.TriggerType == dcmd.TriggerTypeSlashCommands && data.SlashCommandTriggerData.Interaction.Type != discordgo.InteractionApplicationCommandAutocomplete {
```
mrbentarikau, line 175: why not  >
`if data.TriggerType == dcmd.TriggerTypeSlashCommands && data.SlashCommandTriggerData != nil && data.SlashCommandTriggerData.Interaction.Type != discordgo.InteractionApplicationCommandAutocomplete`

clari7744:
> Looks nice, I personally sorted CAH-pack choices alphabetically - matter of taste, also these debug fmt.Prints rolemenu's Autocomplete function has are not necessary : )

I did as well, don't think I've pushed those changes yet though - been debugging because I realized I didn't actually put any comparisons on CAH autocomplete, so it's just showing all the results, all the time.
I did notice the debugs when I was looking back through and removed those as well lol, that change will be soon™️ as well

I'm trying to decide on the best way to set up multi-pack since autocomplete is limited to 100 characters, I'll push what I have soonish for help.

rolecommands/menu.go
```diff
@@ -30,6 +30,31 @@ import (
 
 var recentMenusTracker = NewRecentMenusTracker(time.Minute * 10)
 
+func acFuncRoleMenuCreate(parsed *dcmd.Data) ([]*discordgo.ApplicationCommandOptionChoice, error) {
+	//return []*discordgo.ApplicationCommandOptionChoice{{Name: "Test Choice", Value: "Test Choice"}}, nil
+	fmt.Printf("RoleMenuCreate args: %v\n", parsed.Args)
+	fmt.Printf("RoleMenuCreate Options: %v\n", parsed.SlashCommandTriggerData.Options)
+	name := parsed.Args[0].Str()
+	choices := []*discordgo.ApplicationCommandOptionChoice{}
+	groups, err := models.RoleGroups(qm.Where("guild_id=?", parsed.GuildData.GS.ID), qm.Where("name ILIKE ?", "%"+name+"%"), qm.Load("RoleCommands"), qm.Limit(25), qm.OrderBy("id asc")).AllG(parsed.Context())
```
jo3-l, line 39: Do we really want to hit the database every time users use `rolemenu create`?

rolecommands/menu.go
```diff
@@ -30,6 +30,31 @@ import (
...
+	if err != nil {
+		if errors.Cause(err) == sql.ErrNoRows {
+			return choices, nil
+		}
```
jo3-l, line 43: Try `errors.Is(err, sql.ErrNoRows)`?

rolecommands/menu.go
```diff
@@ -30,6 +30,31 @@ import (
...
+
+		return nil, err
+	}
+	fmt.Print("Got groups")
+	fmt.Printf("RoleMenuCreate groups: %v\n", groups)
```
jo3-l, line 48: Leftover from debugging, presumably?

rolecommands/menu.go
```diff
@@ -30,6 +30,31 @@ import (
 
 var recentMenusTracker = NewRecentMenusTracker(time.Minute * 10)
 
+func acFuncRoleMenuCreate(parsed *dcmd.Data) ([]*discordgo.ApplicationCommandOptionChoice, error) {
+	//return []*discordgo.ApplicationCommandOptionChoice{{Name: "Test Choice", Value: "Test Choice"}}, nil
+	fmt.Printf("RoleMenuCreate args: %v\n", parsed.Args)
+	fmt.Printf("RoleMenuCreate Options: %v\n", parsed.SlashCommandTriggerData.Options)
```
jo3-l, line 36: Leftover from debugging, presumably?

commands/plugin_bot.go
```diff
@@ -186,6 +186,23 @@ func YAGCommandMiddleware(inner dcmd.RunFunc) dcmd.RunFunc {
 			}
 		}
 
+		// Check if the user can execute the command
+		canExecute, resp, settings, err := yc.checkCanExecuteCommand(data)
+		if err != nil {
+			yc.Logger(data).WithError(err).Error("An error occured while checking if we could run command")
+		}
+
+		parseErr := dcmd.ParseCmdArgs(data)
```
jo3-l, line 195: It's idiomatic to use the name `err` for errors throughout, so perhaps reassign `err = dcmd.ParseCmdArgs(data)` instead if possible?

cah/commands.go
```diff
@@ -19,7 +20,7 @@ func (p *Plugin) AddCommands() {
 		Aliases:     []string{"c"},
 		Description: "Creates a Cards Against Humanity game in this channel, add packs after commands, or * for all packs. (-v for vote mode without a card czar).",
 		Arguments: []*dcmd.ArgDef{
-			{Name: "packs", Type: dcmd.String, Default: "main", Help: "Packs separated by space, or * for all of them."},
+			{Name: "packs", Type: dcmd.String, Default: "main", Help: "Packs separated by space, or * for all of them.", Autocomplete: true},
```
jo3-l, line 23: Does the current implementation handle multiple autocomplete arguments correctly?

jo3-l (commented):
It might be worth filtering autocomplete options by string similarity to the input (using `lib/jarowinkler`), WDYT? Some other comments below.

clari7744:
> It might be worth filtering autocomplete options by string similarity to the input (using `lib/jarowinkler`), WDYT? Some other comments below.

I've been starting on some manual comparison, and I'm trying to set it up so it returns something like `pack1 pack2 (packs matching current search)` to support multiple packs, but if a lib exists probably is better to use that

clari7744:
At this point, would it be better to drop the CAH changes, merge this PR with just the middleware autocomplete implementation and rolecommand changes, and create a new PR for CAH autocomplete?

mrbentarikau:
I don't even understand what it is auto-completing...  

This time I am using YAGPDB code and your PR, does not "complete" or search for redbox cah-pack. 
image

clari7744:
> I don't even understand what it is auto-completing...
> 
> This time I am using YAGPDB code and your PR, does not "complete" or search for redbox cah-pack.

@mrbentarikau yeah, I initially uh...forgot to actually implement any kind of pattern matching in CAH, try pulling? Newest push should include that, it just gets a little janky after 100 characters

As for rolemenu, that works because I stuck the search in the DBQuery, and I agree that it shouldn't be requesting on every keypress, just not sure how best to cache the data.

mrbentarikau:
yup, tried the code directly from your repo... at least now I see how it's supposed to work...

clari7744:
The use case for searching is more for when you have more than 25 and you need to narrow it down more.
I think I'm going to make CAH its own PR though, since I'd like to get at least the basis merged so others can start implementing it in more commands if they want.

rolecommands/bot.go
```diff
@@ -51,15 +51,16 @@ func (p *Plugin) AddCommands() {
...
 		},
 		ArgSwitches: []*dcmd.ArgDef{
 			{Name: "m", Help: "Message ID", Type: dcmd.BigInt},
 			{Name: "nodm", Help: "Disable DM"},
 			{Name: "rr", Help: "Remove role on reaction removed"},
 			{Name: "skip", Help: "Number of roles to skip", Default: 0, Type: dcmd.Int},
 		},
-		RunFunc: cmdFuncRoleMenuCreate,
+		RunFunc:          cmdFuncRoleMenuCreate,
+		AutocompleteFunc: acFuncRoleMenuCreate,
```
ashishjh-bst, line 63: I think this should be part of ApplicationCommandOption instead of ApplicationCommand, because if there are more than one Autocomplete arguments it becomes a pain to extend with this.

PR #1617, opened 2024-02-25
customcommands: add support for time.Duration multiplication and division

SoggySaussages:
Multiplying or dividing a `time.Duration` value with an integer currently returns an integer value, not a `time.Duration` value. This greatly limits functionality, and is against go convention which encourages multiplication of `time.Duration` constants.

This pr adds special functionality for `time.Duration` values to properly multiply or divide them with the `mult` and `div` functions and return a `time.Duration` value.

Screenshot 2024-02-25 at 00 33 06

This is a breaking change, if anyone is currently multiplying time constants and expecting an int64 return of the nanosecond count this will break functionality.

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -634,9 +634,9 @@ func tmplMult(args ...interface{}) interface{} {
 		return 0
 	}
 
-	switch args[0].(type) {
+	switch a := args[0].(type) {
 	case float32, float64:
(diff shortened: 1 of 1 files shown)


SoggySaussages:
I never knew `toDuration` could just take an integer value, my mistake. This is the better solution.

PR #1619, opened 2024-03-06
cc: add functions and triggers for components and modals

SoggySaussages:
This PR updates YAGPDB custom commands to allow use of newer Discord features. Specifically, the use of Message Components and Modals. The ability to receive these interactions also opens up the possibility of interaction responses, which this PR also adds.

Changelog:
- added an interaction field to context, which takes a pointer to a `CustomCommandInteraction` type value.
- added a `CustomCommandInteraction` type value which is a pointer to a discordgo.Interaction with a RespondedTo field. This variable is set and pointer added to context upon interaction creation, allowing for the `RespondedTo` field to accurately reflect whether the bot has responded to an interaction yet across different cc executions from the same response. This dictates logic that decides whether the bot should send a "response" or a "followup" when asked to send a message re: an interaction.
- Restructured the `context.SendResponse` function to use a new `sendMessageType` value instead of `isDM`. Since a response may now need to be sent as either a guild channel message, a dm, an interaction response, or an interaction followup, this new type covers our bases.
-  `context.SendResponse` can now send a response as an interaction response or a followup if the triggering CC is triggered by an interaction. The bot will decide whether to send an original response or a followup based on if the interaction has been responded to already or not. It will respond ephemerally if `ephemeralResponse` function is used.
- added templating functions `editResponse,` and `editResponseNoEscape`. `editResponse` works similarly to `editMessage`, but is used to edit interaction responses. While most interaction responses and followup messages can be edited with `editMessage` (given the bot has the right permissions in the channel), ephemeral responses must be edited using a different endpoint, thus necessitating `editResponse`. It uses the following syntax: `editResponse interactionToken msgID msg`. `interactionToken` can be either a token or `nil` to use the current interaction token in context (the one which triggered the cc). `msgID` can either be a message ID, required for editing a followup message, or `nil`, to edit the original interaction response. `msg` functions the same as `editMessage`.
- added function `ephemeralResponse`, which causes the response message to be sent ephemerally if the command was triggered by an interaction.
- added function `sendModal`, which sends the provided modal as an interaction response. syntax is `sendModal modal` where `modal` takes a modal value which can be created with `cmodal`. Since `sendModal` may only be run as an interaction response, it takes no `interactionToken` argument, and will only respond to the interaction triggering the cc.
- added functions `sendResponse`, `sendResponseNoEscape`, `sendResponseNoEscapeRetID`, and `sendResponseRetID`. These function similarly to their `sendMessage` counterparts. They will either send an original interaction response or a followup based on whether the interaction has already been responded to. They use the syntax `sendMessage interactionToken msg`, where `interactionToken` can either be a token or `nil`, to use the current token in context (the interaction which triggered the command). `msg` functions the same as `sendMessage`. You cannot send an original interaction response to an interaction not triggered by the current command. i.e: `if interactionToken != currentFrame.Interaction.Token { send a followup }`
- added functions `updateMessage` and `updateMessageNoEscape`. These function similarly to `editMessage`. it edits whichever message the triggering component (such as a button) was used on. `updateMessage` sends an original interaction response, which differentiates it from `editResponse` which may only send a followup. This distinction is necessary since if an admin would like the only response to a button to be the editing of the message, using `editResponse` will not send an interaction response, and the triggering user will get an error saying the interaction failed regardless of whether it truly has. it follows the syntax `updateMessage msg`. `msg` functions similarly to `editMessage`. Since `updateMessage` may only be run as an interaction response, it takes no `interactionToken` argument, and will only respond to the interaction triggering the cc.
- added function `cbutton` to create buttons which can then be added to `complexMessage`s or `complexMessageEdit`s. It works similarly to `cembed`, directly encoding and unmarshalling values into a button structure WITH THESE THREE EXCEPTIONS: 
    1. the `style` field parses the string names of the different styles such as `primary`, `green`, `secondary`, `link` as an alternative to integer values of the button types. 
    2. If a user passes a `link` argument to the builder, it will automatically correct this to the `url` field discord requires for link buttons. essentially adding an alias for the `url` field since the difference between `link` and `url` may be confusing to users. 
    3. The function will prepend `templates-` to any custom id to limit them to use within templates only. 
This function additionally has validation, and checks for the following errors: 
    - link button specified but no link provided
    - non-link button specified but no custom id provided, assigns a default id if so
    - no label AND no emoji provided
- added function `cmenu` to create select menus which can then be added to `complexMessage`s or `complexMessageEdit`s. It works similarly to `cembed`, directly encoding and unmarshalling values into a button structure WITH THESE TWO EXCEPTIONS: 
    1. the `type` field parses the string names of the different types of select menu such as `user`, `role`, or `text` instead of an integer value representing the type. 
    2. The function will prepend `templates-` to any custom id to limit them to use within templates only. 
This function additionally has validation, and checks for the following errors: 
    - no custom id provided, assigns a default id if so 
    - invalid number of options provided
    - invalid `minValues` or `maxValues` argument provided
    - duplicate `value` fields within the provided options
- added function `cmodal` to create modals which can then be sent with `sendModal`. Parsed manually because the developer in charge of designing modals at discord wants to see the world burn. Available fields are `title`, `custom_id`, and `fields`, which takes a `slice` of `sdict`s. The sdicts are encoded and unmarshalled into `discordgo.TextInput` objects. Assigned a default custom id if not overridden. Also prepends `templates-` to a custom id if it is provided.
- edited `complexMessage` to add `components` and `ephemeral` fields. `ephemeral` adds the ephemeral flag to the message if the value is not falsey. `components` takes either a single message component, a slice of message components, or a slice of slices of message components. If given a single component, it appends a new action row with that single component. If given a slice of components, it automatically distributes the provided components into action rows, trimming off any components which don't fit. It also errors if duplicate custom ids for the components are used. Providing a slice of slices of components allows custom organization of the components into rows. excess is trimmed off. Also assigns default custom ids to components if one is not provided.
- edited `complexMessageEdit` to take a `components` value working exactly the same as in `complexMessage`.
- added Message Component and Modal Submission triggers, which match the provided text trigger as a regex against any component/modal submission received which is prefixed by `templates-`. On the website, switching to these triggers disables the message edit toggle.
- altered the customcommands command to properly display the new types of custom command triggers
- added `.` context data for components and modal submission-triggered CCs:
    - Interaction - full interaction object as provided by discord. Should be used if you need to get `.Interaction.Token` to be saved to database, allowing for subsequent CCs to send followup responses.
    - InteractionData - data specific to the component used, such as custom id and values submitted.
    - CustomID - custom id of the component used
    - Cmd - alias for custom id
    - StrippedID - custom ID with the matched trigger stripped off
    - StrippedMsg - alias for stripped id
    - IsButton - returns true if the used component was a button
    - IsMenu - returns true if the used component was a select menu
    - IsModal - returns true if the used component was a select menu
    - MenuType - returns a string name of the type of select menu if applicable
    - CmdArgs - returns a slice of submitted values to a modal or select menu
- upstream changes to the discordgo library were pulled to enhance compatibility.

This is a PR which has been in the works for about 15 months, ever since I first got a functional yag host and began my journey in golang. Special thanks to the team of incredible testers that helped me make this code more user friendly, widely applicable, and bug-free* over the past few weeks:
Sam
Black Wolf
OneBigLotus
EyesEyes
TOW.
Ranger
sado
Shizuka
Borbot
Henri
lzodd
(and also to all the people on the christopaganism server who have been live-testing my prototypes for a few months)

Please see some use cases we’ve come up with:

Sado’s UNO (uses ephemeral responses, select menus, buttons)

uno1
uno2
uno3

Question of the day (uses buttons, modals, ephemeral responses)

qotd1
qotd2

Verification (uses buttons, ephemeral responses)

verification


SoggySaussages:
@Ranger-4297 pointed out a really smart potential change re: adding components to messages. We've come up with a potential alternative to using `"components" ( cslice ( cbutton ... ) ( cmenu ... ) )` after noting the following concerns about the current implementation:
1. Two functions you will never use outside of `complexMessage` or `complexMessageEdit`.
2. What the hell is a `component`? This may confuse new users, who may not understand why a menu and a button are the same thing.

The solution we have come up with is this:
```go
{{ $msg := complexMessage "buttons" ( cslice ( sdict … ) ( sdict … ) ) "menus" ( cslice ( sdict … ) ( sdict … ) ) }}
```
**Pros**
1. Clearly denotes which components are which, may make more sense to more users. 
2. Eliminates need for a new function
4. Could still order your rows by doing `"buttons" … "menus" … "buttons" … "buttons" … "menus" …
5. Now that I’m thinking this through, I like this option better
**Cons**
1. May not be immediately apparent that you can change the order
2. Encourages users to double up on button or menu declarations which may become confusing, and also isn’t very conventionally accepted despite it making sense in this case
4. More difficult for conditional branching, especially for users who don’t know how to `.Set` an sdict.

My proposal is thus: Keep the current functionality with buttons and menus in `"components"`, but ADD the ability to use `"buttons"` and `"menus"` as the recommended, simpler option. Keep `"components"` for more advanced users. Please share thoughts as to how to move forward in this regard.

lib/discordgo/components.go
```diff
@@ -121,7 +126,7 @@ const (
 // ComponentEmoji represents button emoji, if it does have one.
 type ComponentEmoji struct {
 	Name     string `json:"name,omitempty"`
-	ID       int64  `json:"id,string,omitempty"`
+	ID       string `json:"id,omitempty"`
```
mrbentarikau, line 129: All YAG's snowflake IDs are int64, why revert back to string?

lib/discordgo/message.go
```diff
@@ -115,7 +115,7 @@ type Message struct {
 	Attachments []*MessageAttachment `json:"attachments"`
 
 	// A list of components attached to the message.
-	Components []MessageComponent `json:"-"`
+	Components []unmarshalableMessageComponent `json:"components"`
```
mrbentarikau, line 118: why not (m *Message)UnmarshalJSON the whole message? adding lowercase type in-between makes no logical sense

to get a button object from type .unmarshalableMessageComponent
`{{ index (index (getMessage nil <mID>).Components 0).MessageComponent.Components 0 | printf "%#v" }}`

 versus from .MessageComponent
` {{ index (index (g etMessage nil <mID>).Components 0).Components 0 | printf "%#v" }}`

Borbot33: I wanted to do it this way, but I don't know how. I thought it was the same thing

mrbentarikau: well, of all the code you copied from bwmarrin, UnmarshalJSON is also there. : )

lib/discordgo/components.go
SoggySaussages: I'm not sure, this was an upstream bwmarrin/discordgo update. I copy-pasted the entire file from upstream, I should have looked more carefully at each change.

mrbentarikau: Bwmarrin's are all strings, YAG uses int64

lib/discordgo/message.go
SoggySaussages: Would you please elaborate on the better solution? The issue we were encountering was that when the json was `-`, of course no data from the get request was being unmarshalled via `(s *Session) ChannelMessage`, and when the json was `components`, it failed to unmarshal because `MessageComponent` is an interface, not a struct.

mrbentarikau: exactlly like bwmarrin has its UnmarshalJSON helpers for message and message events in events.go, and yes, exactly Components []MessageComponent `json:"-"` for type Message.... it will not panic out

SoggySaussages: got it, I believe I have fixed this with the recent commit. `getMessage` and `.Message` now seem to return the components neatly. everything relevant should now match upstream, I agree that that is the most important thing, so thank you for helping me out with it. does it look alright to you?
thanks for your patience.

lib/discordgo/events.go
```diff
@@ -561,4 +578,4 @@ type GuildAuditLogEntryCreate struct {
 type GuildJoinRequestUpdate struct{}
 type GuildJoinRequestDelete struct{}
 type VoiceChannelStatusUpdate struct{}
-type ChannelTopicUpdate struct{}
\ No newline at end of file
+type ChannelTopicUpdate struct{}
```
mrbentarikau, line 581: here you add new struct to the fold, so it would be safer to execute that eventhandler script, comment in the beginning of the file recommends - or copy stuff from eventhandlers.go, manual - error prone.

otherwise looks good, if something was wrong, bot would crash at once due to SEGFAULT

SoggySaussages: Screenshot 2024-03-07 at 18 43 26
I thought that was just a whitespace removal by `go fmt`, but I'm not totally sure. How could I tell that I added a function by accident?

mrbentarikau: oh, cool, already there then... no problem, my bad : )

customcommands/assets/customcommands-editcmd.html
```diff
@@ -501,21 +515,45 @@ <h2 class="card-title">Custom Command Information</h2>
 
             $("#cc-time-trigger-details").addClass("hidden");
             $("#cc-text-trigger-details").addClass("hidden");
+            $("#cc-message-trigger-details").addClass("hidden");
```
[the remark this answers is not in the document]
SoggySaussages: I agree, this was a hassle to figure out

customcommands/bot.go
```diff
@@ -812,6 +899,66 @@ func findReactionTriggerCustomCommands(ctx context.Context, cs *dstate.ChannelSt
...
+		if !CmdRunsForUser(v.CC, ms) {
+			continue
+		}
+
+		filtered = append(filtered, v)
+	}
+
+	sortTriggeredCCs(filtered)
+
+	return filtered, nil
```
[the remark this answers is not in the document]
SoggySaussages: Ah nice catch, thank you so much

lib/discordgo/endpoints.go
```diff
@@ -134,6 +134,9 @@ var (
 	EndpointChannelWebhooks = func(cID int64) string { return "" }
 	EndpointWebhook         = func(wID int64) string { return "" }
 	EndpointWebhookToken    = func(wID int64, token string) string { return "" }
+	EndpointWebhookMessage  = func(wID int64, token string, messageID int64) string {
```
mrbentarikau, line 137: This here does not follow the logic YAG's endpoints.go is built up
this var section should be `EndpointWebhookMessage  = func(wID int64, token, messageID string) string { return "" }`

and you give the full inside CreateEndpoints function down below...

notice I use string for messageID type, why? it has `@original` possible > https://discord.com/developers/docs/interactions/receiving-and-responding#get-original-interaction-response

SoggySaussages: There is currently a separate endpoint for original interaction response, if I add capability for @original to be used in WebhookMessage endpoint, shall I also remove the original interaction response endpoint? That would match upstream discordgo more closely iirc.

mrbentarikau: closer to the upstream the better, less changes in the future

SoggySaussages: Should there be a constant for `"@original"` to be passed to (s *Session) WebhookMessage? I can't find one on discordgo upstream, but I don't love needing to pass the same string literal for each use of the function.

SoggySaussages: Ah, I think I see what you're saying now, that makes sense to me. Thanks so much for your wisdom and patience as always :)

mrbentarikau (approved):

mrbentarikau:
This is just a suggestion, but what I would do is place all new interaction context functions to separate file, `context_interactions.go` for example. There's basically 1000 lines new code and this bloats general and context_funcs

SoggySaussages:
> This is just a suggestion, but what I would do is place all new interaction context functions to separate file, `context_interactions.go` for example. There's basically 1000 lines new code and this bloats general and context_funcs

I don't personally love the idea of putting the general functions in an "interactions" file because they're not really related to interactions, they are just object functions which dictate features of a message. You could use disabled buttons purely for aesthetics and never call a CC interaction.

I feel that something such as `context_interactions.go` implies that the included functions require interaction data in context to function. While that is true for 3 of them, the others can function as intended without interaction data in context, and some could use either interaction context or not. I think it wouldn't be worthwhile to include three functions and not the others. For these reasons I am personally against having them in a separate file.

jo3-l:
For what it is worth, I don't think the naming is an issue and would prefer to see a separate file.

mrbentarikau:
functions: CreateComponent, CreateButton, CreateSelectMenu, CreateModal, distributeComponents, validateCustomID, validateActionRowsCustomIDs

methods: tmplEditInteractionResponse, tmplEphemeralResponse, tmplGetResponse, tmplSendModal, tmplSendInteractionResponse, tmplUpdateMessage, tokenArg

are all tied to interactions and sum up around 800 lines of code... seeing these separate makes more sense to me

also in assets/customcommands-edit.html > "Custom ID Regex" does not really say much to regular user and it should be at least "Component (Custom) ID Regex"

customcommands/assets/customcommands-editcmd.html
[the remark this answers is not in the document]
jo3-l: Heads up: I finally got around to sending #1646 to clean this up. If that PR is merged, there will be some merge conflicts here, but those should be easy to resolve -- refer to my PR description for changes.

ashishjh-bst:
Is anything specific pending in this @SoggySaussages ?

mrbentarikau:
all interaction related stuff should still be in a separate file like context_interactions.go, like I said on Mar 18th

ashishjh-bst:
> all interaction related stuff should still be in a separate file like context_interactions.go, like I said on Mar 18th

@SoggySaussages I agree with this it would make the code cleaner keeping all interaction related stuff in a different file.

SoggySaussages:
> > all interaction related stuff should still be in a separate file like context_interactions.go, like I said on Mar 18th
> 
> @SoggySaussages I agree with this it would make the code cleaner keeping all interaction related stuff in a different file. 

Oh sorry, I could've sworn I pushed that change, I did change that locally and will push when I get home.

SoggySaussages:
Rebased and updated with separate file for new functions and js fixes in accordance with 1646 (hopefully everything fit together smoothly)

customcommands/assets/customcommands-editcmd.html
```diff
@@ -541,17 +555,21 @@ <h2 class="card-title">Custom Command Information</h2>
 
     const intervalTriggerEls = ['#interval-cc-run-now', '#cc-time-trigger-details'];
     const textTriggerEls = ['#cc-text-trigger-details', '#cc-extra-settings'];
+    const messageTriggerEls = [...textTriggerEls, '#cc-message-trigger-label', '#cc-message-trigger-details'];
+    const customIDTriggerEls = [...textTriggerEls, '#cc-customid-trigger-label'];
```
jo3-l, line 559: nit, but I would rather call these `interactionTriggerEls`. WDYT?

SoggySaussages: totes

PR #1620, opened 2024-03-12
Added a Bot Owner Only,  Clear DM Command. 

ashishjh-bst:
This deletes upto 100 DM messages sent by the bot.

### DIFF
--- a/stdcommands/cleardm/cleardm.go
+++ b/stdcommands/cleardm/cleardm.go
@@ -1,0 +1,53 @@
+package cleardm
+
+import (
+	"fmt"
+
+	"github.com/botlabs-gg/yagpdb/v2/commands"
+	"github.com/botlabs-gg/yagpdb/v2/common"
+	"github.com/botlabs-gg/yagpdb/v2/lib/dcmd"
+	"github.com/botlabs-gg/yagpdb/v2/lib/discordgo"
+	"github.com/botlabs-gg/yagpdb/v2/stdcommands/util"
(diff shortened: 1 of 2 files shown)


stdcommands/cleardm/cleardm.go
```diff
@@ -0,0 +1,53 @@
...
+			logrus.WithError(err).Errorf("Failed to get DM messages for user %d", target.ID)
+			return nil, err
+		}
+		if len(messages) == 0 {
+			return "No messages found", nil
+		}
+		count := 0
+		for _, message := range messages {
+			if message.Author.ID == common.BotUser.ID {
+				err = common.BotSession.ChannelMessageDelete(dm.ID, message.ID)
```
mrbentarikau, line 43: have not tested, but wouldn't this section be better using goroutines and channels?
something like 
```go
go func(cID, mID int64) {
    time.Sleep(500 * time.Millisecond)
    err := common.BotSession.ChannelMessageDelete(cID, mID)
    if err != nil {
       errors <- err
    }
 }(dm.ID, message.ID)```

ashishjh-bst: I agree go routine would be faster for this, I didn't think of this, thanks!

ashishjh-bst: On a second thought, I don't see a clear benefit of this. If it is to prevent rate limits, the bucket on the Ratelimit bucket on the ChannelMessageDelete endpoint already handles it.

PR #1621, opened 2024-03-12
customcommands: populate ctx.Msg in evalcc

l-zeuch:
Populate templates.Context's Msg field with a copy of the triggering
message that launched evalcc.

This bug was caught via my own testing; I was looking to launch replog
via exec and used evalcc for that — it did not work. The reason is, that
the exec (and by extension the execAdmin) logic make a copy of
templates.Context.Msg, which was never populated.

By populating it, we ensure that the exec logic has a message to work
with and thusly cooporate as expected with restricted inbuilt commands
like replog.

This is defintely a bug, given how evalcc is supposed to behave *almost
like* a normal custom command.


mrbentarikau:
without message context it yeah should default to botUser as target

l-zeuch:
Yep, it does default to the bot user:
image

PR #1624, opened 2024-03-24
customcommands: link to ccs in cc command

SoggySaussages:
Link to the CC's corresponding edit page on the dashboard when examining a single CC with the CustomCommands command. The link is masked by the CC ID so the user-facing response doesn't get any more bulky.

Also suppresses embeds so said links do not become embeds.

Screenshot 2024-03-23 at 23 10 25

### DIFF
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -243,7 +243,7 @@ var cmdListCommands = &commands.YAGCommand{
 		}
 
 		var ccFile *discordgo.File
-		var msg *discordgo.MessageSend
+		msg := &discordgo.MessageSend{Flags: discordgo.MessageFlagsSuppressEmbeds}
 
 		responses := fmt.Sprintf("```\n%s\n```", strings.Join(cc.Responses, "```\n```"))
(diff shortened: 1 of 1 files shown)


mrbentarikau:
if control panel is not publicly, membersOnly or rolesOnly readable, then it should not add that link there.

SoggySaussages:
> if control panel is not ~~publicly,~~ membersOnly or rolesOnly readable, then it should not add that link there.
> 
> 
> 
> publicly does not really matter, because non-members can't see command's return : )

I thought about only returning the links if the executing user has read permission. The reason I didn't was the logs command doesn't do that by default either. It doesn't check if the executing user has log read permission before sending the link, it just shows an error message on the dash if you try to click it without the right permission. 

I could see implementing said check for the executing user, but currently I am against only returning links if the guild members or roles readable. I think it should match the behaviour of "if this link isn't for you, ignore it" exhibited by the logs command. 

I am interested to hear your thoughts.

ashishjh-bst (approved):

PR #1625, opened 2024-03-26
Implement user-installed integration

User9684:
In order for this to actually be used on a bot, the setting must be enabled in the installation panel.

image

### DIFF
--- a/bot/paginatedmessages/paginatedinteractions.go
+++ b/bot/paginatedmessages/paginatedinteractions.go
@@ -95,15 +95,15 @@ func createNavigationButtons(prevDisabled bool, nextDisabled bool) []discordgo.M
 	}
 }
 
-func CreatePaginatedMessage(data *dcmd.Data, guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc) (*PaginatedMessage, error) {
+func CreatePaginatedMessage(guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc, data ...*dcmd.Data) (*PaginatedMessage, error) {
 	if initPage < 1 {
 		initPage = 1
 	}
 
 	var interaction = &discordgo.Interaction{ID: 0}
 
-	if !reflect.ValueOf(data.SlashCommandTriggerData).IsNil() {
-		interaction = data.SlashCommandTriggerData.Interaction
+	if len(data) >= 1 && !reflect.ValueOf(data[0].SlashCommandTriggerData).IsNil() {
+		interaction = data[0].SlashCommandTriggerData.Interaction
 	}
 
 	pm := &PaginatedMessage{
--- a/bot/paginatedmessages/paginatedinteractions.go
+++ b/bot/paginatedmessages/paginatedinteractions.go
@@ -1,7 +1,6 @@
 package paginatedmessages
 
 import (
-	"reflect"
 	"strconv"
 	"sync"
 	"time"
@@ -102,7 +101,7 @@ func CreatePaginatedMessage(guildID, channelID int64, initPage, maxPages int, pa
 
 	var interaction = &discordgo.Interaction{ID: 0}
 
-	if len(data) >= 1 && !reflect.ValueOf(data[0].SlashCommandTriggerData).IsNil() {
+	if len(data) >= 1 && data[0].SlashCommandTriggerData != nil {
 		interaction = data[0].SlashCommandTriggerData.Interaction
 	}
 
--- a/bot/paginatedmessages/paginatedinteractions.go
+++ b/bot/paginatedmessages/paginatedinteractions.go
@@ -1,12 +1,14 @@
 package paginatedmessages
 
 import (
+	"reflect"
 	"strconv"
 	"sync"
 	"time"
 
 	"github.com/botlabs-gg/yagpdb/v2/bot/eventsystem"
 	"github.com/botlabs-gg/yagpdb/v2/common"
+	"github.com/botlabs-gg/yagpdb/v2/lib/dcmd"
 	"github.com/botlabs-gg/yagpdb/v2/lib/discordgo"
 )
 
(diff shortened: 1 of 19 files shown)


lib/discordgo/structs.go
```diff
@@ -1421,6 +1421,8 @@ type CreateApplicationCommandRequest struct {
 	Options           []*ApplicationCommandOption `json:"options"`                      // the parameters for the command
 	DefaultPermission *bool                       `json:"default_permission,omitempty"` // (default true)	whether the command is enabled by default when the app is added to a guild
 	NSFW              bool                        `json:"nsfw,omitempty"`               // marks a command as age-restricted
+	Integration_Types []int                       `json:"integration_types"`            // Defines the integration types a command supports https://discord.com/developers/docs/resources/application#application-object-application-integration-types
```
mrbentarikau, line 1424: Go linters have this saying: "don't use underscores in go names"
https://go.dev/doc/effective_go#mixed-caps

same for for variable `integration_types` you have in slashcommands.go

User9684: Will remove when home

mrbentarikau:
One thing I'd add is if interaction context is of [type](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-context-types) 2 `PRIVATE_CHANNEL`, the response should be ephemeral - no need to spam user DMs with bot interactions.

User9684:
> One thing I'd add is if interaction context is of [type](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-context-types) 2 `PRIVATE_CHANNEL`, the response should be ephemeral - no need to spam user DMs with bot interactions.

While I do agree that there's no need to flood dms with interactions, I also think it would ruin the fun of users who want to use the bot in a GDM, as the type is `PRIVATE_CHANNEL` in there aswell.

User9684:
There might be some way to check if the channel is a GDM, but I'd have to look further into the API, and I feel as though that would be out of scope for this PR anyways.

mrbentarikau:
I think Context field for type Interaction is in the scope of this PR.
Is running application commands in random people's DMs necessary or fun, I guess is for the maintainer to decide.

User9684:
Pagination will probably need some degree of changes if a command with RunInDM enabled uses it anywhere (i.e. forex)

mrbentarikau:
What I did...
I created a separate pagination function for interactions, this one sends CreateFollowupMessage and HandlePageButtonClick method handles non-guild related messages separately via EditOriginalInteractionResponse

it works, ok buttons don't get deleted... yet

bot/paginatedmessages/paginatedinteractions.go
```diff
@@ -93,11 +95,17 @@ func createNavigationButtons(prevDisabled bool, nextDisabled bool) []discordgo.M
 	}
 }
 
-func CreatePaginatedMessage(guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc) (*PaginatedMessage, error) {
+func CreatePaginatedMessage(data *dcmd.Data, guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc) (*PaginatedMessage, error) {
```
mrbentarikau, line 98: you could make data parameter variadic, then you don't have to edit all function calls all over the place

User9684: Would making it variadic allow 0 of it to be passed and still work? I've never really used variadic parameters as I typically see no reason for them/

mrbentarikau: https://go.dev/play/p/AU_Xu0_QKLW

User9684:
Discord fixed https://github.com/discord/discord-api-docs/issues/6730

bot/paginatedmessages/paginatedinteractions.go
```diff
@@ -93,11 +95,17 @@ func createNavigationButtons(prevDisabled bool, nextDisabled bool) []discordgo.M
...
 
-func CreatePaginatedMessage(guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc) (*PaginatedMessage, error) {
+func CreatePaginatedMessage(guildID, channelID int64, initPage, maxPages int, pagerFunc PagerFunc, data ...*dcmd.Data) (*PaginatedMessage, error) {
 	if initPage < 1 {
 		initPage = 1
 	}
 
+	var interaction = &discordgo.Interaction{ID: 0}
+
+	if len(data) >= 1 && !reflect.ValueOf(data[0].SlashCommandTriggerData).IsNil() {
```
jo3-l, line 105: Why is reflection used here? If it is truly necessary, a comment might be nice -- this pattern does not seem to be used anywhere else in the codebase -- but I suspect it is gratuitous and a direct test against `nil` would suffice.

commands/slashcommands.go
```diff
@@ -164,13 +164,23 @@ func (p *Plugin) yagCommandToSlashCommand(cmd *dcmd.RegisteredCommand) *discordg
 	}
 	t := true
 
+	integrationtypes := []int{0}
+	contexts := []int{0}
+
+	if cast.RunInDM {
+		integrationtypes = append(integrationtypes, 1) // USER_INSTALL
+		contexts = append(contexts, 1, 2)              // BOT_DM, PRIVATE_CHANNEL
```
jo3-l, line 172: Can we have newtypes `IntegrationType`, `InteractionContextType` or similar for these?

bot/paginatedmessages/paginatedinteractions.go
User9684: I forgot to remove that, it's not supposed to be anymore. It was because Discord had a bug with the response.

commands/slashcommands.go
User9684: Like this?
image

jo3-l: No, I meant constants replacing the magic numbers `1`, `2`, etc. See, e.g. `ForumLayout`:
https://github.com/botlabs-gg/yagpdb/blob/447a52732f5b901314397fe9079290e59d9f8d45/lib/discordgo/structs.go#L172-L181

User9684: Ah that makes more sense, do you think I should still have the IntegrationTypes and IntegrationContexts types though?

jo3-l: No, `[]IntegrationType` and `[]IntegrationContext` is clear enough IMO.

User9684: Agreed

User9684: Have a const for authorizing_integration_owners too, correct?

User9684: Actually, I'm not entirely sure how I could do that for the "1" key because the value is a userID.

jo3-l: Sorry, I'm not quite sure what the issue is. Could you elaborate?

User9684: Nevermind, I was overthinking it lol

PR #1629, opened 2024-04-07
customcommands: public ccs

SoggySaussages:
This PR grants the ability to share individual CCs via a public link. To share a CC you must enable sharing via the Publicity toggle on the panel, and share the public link with another user. This is facilitated by the Share Command modal on the cc edit page, allowing you to copy the public link.

When visiting the public link, you will get err=noaccess if the cc is not public, and sent to homepage if ID is invalid. If you have view perms for the server, it redirects you to regular edit page.

If you reach the public page you will get a stripped version of the command edit page with all inputs disabled, guild-specific info like command group, run count, and restrictions omitted. It is also syntax highlighted responses and a Copy button to copy the code of each response. This copy text button is available on the regular edit page as well.

The CustomCommands command will return the public link masked in the CC ID if the CC is public, or the link to its edit page if the user has read access. It functions as normal if neither of these are true.

public_end
share_custom_command
publicity_toggle

### DIFF
--- a/customcommands/web.go
+++ b/customcommands/web.go
@@ -129,6 +129,9 @@ func (p *Plugin) InitWeb() {
 	subMux.Handle(pat.Post("/creategroup"), web.ControllerPostHandler(handleNewGroup, getHandler, GroupForm{}))
 	subMux.Handle(pat.Post("/groups/:group/update"), web.ControllerPostHandler(handleUpdateGroup, getGroupHandler, GroupForm{}))
 	subMux.Handle(pat.Post("/groups/:group/delete"), web.ControllerPostHandler(handleDeleteGroup, getHandler, nil))
+
+	web.ServerPublicMux.Handle(pat.Get("/customcommands/commands/:cmd"), PublicCommandMW(getCmdHandler))
+	web.ServerPublicMux.Handle(pat.Get("/customcommands/commands/:cmd/"), PublicCommandMW(getCmdHandler))
 }
 
 func handleGetDatabase(w http.ResponseWriter, r *http.Request) (web.TemplateData, error) {
@@ -235,6 +238,7 @@ func handleGetCommand(w http.ResponseWriter, r *http.Request) (web.TemplateData,
 	templateData["CC"] = cc
 	templateData["Commands"] = true
 	templateData["IsGuildPremium"] = premium.ContextPremium(r.Context())
+	templateData["PublicLink"] = fmt.Sprintf("%s/public/%d/customcommands/commands/%d", web.BaseURL(), activeGuild.ID, cc.LocalID)
 
 	return serveGroupSelected(r, templateData, cc.GroupID.Int64, activeGuild.ID)
 }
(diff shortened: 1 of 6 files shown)


mrbentarikau:
this feels a bit weird, first the control panel gets closed off to public and now we are going to open it section by section again with a lot of new not necessary code.

ashishjh-bst:
What I would have preferred was make a short public link, https://yagpdb.xyz/cc/id_here

Where Id here would be a sharing id, and the link can be used to share CCs, and give an import button on which shows a user the list of their servers, and let them import it to a specific server. 

The server ID shouldn't be exposed here, and if importing should be a 1 click thing. 

Some stats can be shown on the original CC like "total imports". 

And CC can imports can be disabled anytime by the original server.

The link shouldn't be discoverable unless shared by the someone from the original server, otherwise it opens this to scraping.

customcommands/web.go
```diff
@@ -725,3 +885,33 @@ func updateTemplateWithCountData(count int, templateData web.TemplateData, ctx c
...
+		cc, err := models.CustomCommands(
+			models.CustomCommandWhere.PublicID.EQ(publicID)).OneG(r.Context())
+		if err != nil {
+			logger.Error(errors.WithStackIf(err))
+			http.Redirect(w, r, "/", http.StatusTemporaryRedirect)
+			return
+		}
+
+		if read, _ := web.IsAdminRequest(ctx, r); read {
+			http.Redirect(w, r, fmt.Sprintf("/manage/%d/customcommands/commands/%d/", cc.GuildID, cc.LocalID), http.StatusSeeOther)
```
ashishjh-bst, line 903: If someone comes from a public link to the CC, let them be on the public link whoever they are, the redirection here shouldn't happen if they admin or have read access to the CC.

customcommands/web.go
```diff
@@ -539,6 +588,117 @@ func handleUpdateAndRunNow(w http.ResponseWriter, r *http.Request) (web.Template
...
+	pubsub.EvictCacheSet(cachedCommandsMessage, activeGuild.ID)
+
+	return templateData, err
+}
+
+// checks the ID truncated at length characters, and returns this truncated
+// string if it doesn't already exist. If it does exist, the function runs
+// recursively, adding 1 to length and returns the result
+func addCharIfCollides(ctx context.Context, full string, length int) string {
+	truncated := full[:length]
```
ashishjh-bst, line 693: Possible YAGNI, the chances of collision here are negligible, if you think this can happen maybe the code can be 10 chars long by default. git uses 7 char long truncated hashes, so 10 should be sufficient.

ashishjh-bst (changes requested):

phenpessoa:
Is there any reason why we'd want to use a base64 encoded sha1 hash instead of a UUID?

SoggySaussages:
> Is there any reason why we'd want to use a base64 encoded sha1 hash instead of a UUID?

Only because I understand sha1 and base64 more than UUID. Would UUID work similarly well if we are truncating to the first 10 characters? 

It would probably then look like https://yagpdb.xyz/cc/xxxxxxxx-x, which purely from an aesthetic point of view I would prefer https://yagpdb.xyz/cc/xxxxxxxxxx, and if we take out the dashes when truncating it's still a two step process. 

I also think the thought was to pass as little decodable data in the short link as possible, hence having a guild ID + local ID salted with current time, making it extremely difficult to decode any data from it at all, and even harder to iterate through possibilities to hunt for public CCs one hasn't been directly given access to.

phenpessoa:
> Only because I understand sha1 and base64 more than UUID. Would UUID work similarly well if we are truncating to the first 10 characters?
> It would probably then look like https://yagpdb.xyz/cc/xxxxxxxx-x, which purely from an aesthetic point of view I would prefer https://yagpdb.xyz/cc/xxxxxxxxxx, and if we take out the dashes when truncating it's still a two step process.

Is there a reason why we are trucating it to 10 digits only? 

> I also think the thought was to pass as little decodable data in the short link as possible, hence having a guild ID + local ID salted with current time, making it extremely difficult to decode any data from it at all, and even harder to iterate through possibilities to hunt for public CCs one hasn't been directly given access to.

UUID are random, and are pretty hard to guess, considering a secure random source. But, just like any other hashing algorithm, one could theoretically guess it. If the link is public, we can assume it is not secure, regardless of how hard we make it to guess the URL. 
That being said, I asked about UUIDs because this looks like exactly what they were created for. And they solve the point brought up by @ashishjh-bst.

ashishjh-bst:
@phenpessoa  we want to implement short links for easy sharing, kind of like discord invite codes and bit.ly urls. 

The sharable urls would look like `yagpdb.xyz/cc/shortlinkhere`

phenpessoa:
Fair enough, I don't think UUIDs would work then. Please disregard @SoggySaussages.

PR #1632, opened 2024-04-12
fix: Clean-up usage of X-Audit-Log-Reason

Wolveric:
It was recently noted that the YAGPDB was not setting an audit log reason, when kicking users via command, or when unbanning users. This first of these, was due to the fact the relevant method was based on an old API design, and hadn't been updated, and the second because the relevant method didn't exist.

This PR fixes both of these issues, by adding the relevant method, as well as updating the existing one to use the `X-Audit-Log-Reason` header, rather than the old query string method.

### DIFF
--- a/lib/discordgo/restapi.go
+++ b/lib/discordgo/restapi.go
@@ -758,6 +758,22 @@ func (s *Session) GuildBanDelete(guildID, userID int64) (err error) {
 	return
 }
 
+// GuildBanDeleteWithReason removes the given user from the guild bans,
+// including sending an audit log reason.
+// guildID    : The ID of a Guild
+// userID     : The ID of a User
+// reason     : The reason for removing the ban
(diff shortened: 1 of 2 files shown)


mrbentarikau:
maybe prepend author also to RemoteTimeout function

mrbentarikau:
now the same thing is in three places in the code, why not function it?

Wolveric:
Where would you suggest putting it in that case?

mrbentarikau:
into a function?

func fullReason(...
or whatever that name would be

Wolveric:
I mean in the file, or file system...

mrbentarikau:
to the bottom of punishemts.go for example : )

mrbentarikau:
for example, I use > 
```go
// prepends author to action reason
func authorToReason(reason string, author *discordgo.User) (fullReason string) {
	fullReason = reason
	if author.ID != common.BotUser.ID {
		fullReason = author.String() + ": " + reason
	}
	return
}

PR #1638, opened 2024-04-17
Refactored the Public CC page, and did some UI fixes for old issues. 

ashishjh-bst:
- Separated public CC into a separate template and controller. 
- Made UI less clunky 
- Fixed legacy issues like lengths for trigger, name, and cc not showing on page load. 
- Fixed broken header on tablet ui.

### DIFF
--- a/customcommands/web.go
+++ b/customcommands/web.go
@@ -44,6 +44,9 @@ var PageHTMLMain string
 //go:embed assets/customcommands-database.html
 var PageHTMLDatabase string
 
+//go:embed assets/customcommands-public.html
+var PageHTMLPublicCmd string
+
 // GroupForm is the form bindings used when creating or updating groups
 type GroupForm struct {
 	ID                int64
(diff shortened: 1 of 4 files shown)


customcommands/web.go
```diff
@@ -285,6 +281,27 @@ func handleGetCommand(w http.ResponseWriter, r *http.Request) (web.TemplateData,
...
+	// accessed via shortlink, interpret as public ID instead
+	cc, err = models.CustomCommands(
+		models.CustomCommandWhere.PublicID.EQ(cmdParam)).OneG(r.Context())
+	if err != nil {
+		templateData.AddAlerts(web.ErrorAlert("Command couldn't be found via shortlink"))
+		return templateData, errors.WithStackIf(err)
+	}
+	templateData["CC"] = cc
+	templateData["Commands"] = true
+	templateData["PublicID"] = cc.PublicID
```
SoggySaussages, line 299: I'm having trouble identifying where this is used. Would you please share what it's used for?

ashishjh-bst: I was going to use this to replace .LocalID everywhere on the Public page, but then that wasn't needed later.

PR #1641, opened 2024-04-19
Made custom commands length 20k for premium servers

ashishjh-bst:


mrbentarikau:
public asset should have MaxCCLength as well and func handleGetPublicCommand in web.go. 
Also what happens if premium length gets imported to non-premium

mrbentarikau:
Valid points.
Another, maybe smaller issue, if user has max amounts of CC already in use for that specific guild where they want to import, what will happen then? This import should fail as well, no need for additional database write.

ashishjh-bst:
> Valid points. Another, maybe smaller issue, if user has max amounts of CC already in use for that specific guild where they want to import, what will happen then? This import should fail as well, no need for additional database write.

I think that case is already handled by https://github.com/botlabs-gg/yagpdb/blob/c5ca0542aed688df0bc179d0155cd30482377c38/customcommands/web.go#L389, as the import just has a middleware before newCommandHandler. 

I have also added the checks and appropriate error handling as discussed above.

PR #1642, opened 2024-04-19
patch to show only write guilds in Imports

ashishjh-bst:


SoggySaussages:
if this pr is already here, can we add only showing premium guilds for command lengths that need premium?

ashishjh-bst:
> if this pr is already here, can we add only showing premium guilds for command lengths that need premium?

I think that would be unnecessary, for most selfhosts this wouldn't matter and for yag it would leave people clueless on why all servers aren't showing. So just an error is better.

PR #1649, opened 2024-05-20
all: futureproof viewperms and template Permissions map by autogenerating perm name to bit map

jo3-l:
This PR refactors some code surrounding the following permission-related functionality:
- the viewperms command (which lists the permissions of a given user), and
- the `Permissions` context data in custom commands (which is a map from permission name to bit value, so that `.Permissions.Administrator = 8` and so on.)

Best reviewed commit-by-commit.

## Motivation and context for change

Ideally, when Discord adds a new permission bit (such as, for instance, `MODERATE_MEMBERS`), to support this permission in YAGPDB it should be sufficient to just add a new entry in `lib/discordgo/permission.go`. Unfortunately this is not the case currently: besides updating `lib/discordgo/permission.go`) to export a new `discordgo.Permission___`, we should also support the permission in the `viewperms` command by adding a new case to `util.HumanizePermissions` and additionally add a new entry in the `Permissions` map used for custom command templating.

The latter two steps are prone to be forgotten when adding new permissions. `discordgo.PermissionPrioritySpeaker` (among many other permissions), for instance, has been present for many years but was only supported in `viewperms` in #1227. The more recent `discordgo.ModerateMembers` (timeout permission) was also added to `viewperms` in that PR, but was never added to the context `Permissions` map in custom commands. In fact, out of the 41 permissions exported from `discordgo` at present, the context `Permissions` map is missing _13_ of them.

The problem is thus as follows: **when implementing support for new permissions, there are multiple places _spread across the codebase_ that need to be updated, and often we only remember to update `discordgo` and change the other code only when (and if) it is noticed that the permissions are missing there.**

## The refactor

The change proposed here attempts to address the problem by:
- exporting a list of all permission bits from `discordgo`, and
- autogenerating a function `discordgo.PermissionName` that goes from bit value to human-readable name. So `discordgo.PermissionName(8) = discordgo.PermissionName(discordgo.PermissionAdministrator) = "Administrator"`.

These two primitives allow us to simplify the implementation of `util.HumanizePermissions` (which `viewperms` is dependent on) so that it no longer requires manual changes when new permissions are added.

<table>
<tr>
<th>Before</th>
<th>After*</th>
</tr>
<tr>
<td>

```go
func HumanizePermissions(perms int64) (res []string) {
	if perms&discordgo.PermissionAdministrator == discordgo.PermissionAdministrator {
		res = append(res, "Administrator")
	}
	if perms&discordgo.PermissionManageServer == discordgo.PermissionManageServer {
		res = append(res, "ManageServer")
	}
	if perms&discordgo.PermissionViewGuildInsights == discordgo.PermissionViewGuildInsights {
		res = append(res, "ViewGuildInsights")
	}

	if perms&discordgo.PermissionReadMessages == discordgo.PermissionReadMessages {
		res = append(res, "ReadMessages")
	}
	if perms&discordgo.PermissionSendMessages == discordgo.PermissionSendMessages {
		res = append(res, "SendMessages")
	}
	if perms&discordgo.PermissionSendTTSMessages == discordgo.PermissionSendTTSMessages {
		res = append(res, "SendTTSMessages")
	}
	if perms&discordgo.PermissionManageMessages == discordgo.PermissionManageMessages {
		res = append(res, "ManageMessages")
	}
	if perms&discordgo.PermissionEmbedLinks == discordgo.PermissionEmbedLinks {
		res = append(res, "EmbedLinks")
	}
	if perms&discordgo.PermissionAttachFiles == discordgo.PermissionAttachFiles {
		res = append(res, "AttachFiles")
	}
	if perms&discordgo.PermissionReadMessageHistory == discordgo.PermissionReadMessageHistory {
		res = append(res, "ReadMessageHistory")
	}
	if perms&discordgo.PermissionMentionEveryone == discordgo.PermissionMentionEveryone {
		res = append(res, "MentionEveryone")
	}
	if perms&discordgo.PermissionUseExternalEmojis == discordgo.PermissionUseExternalEmojis {
		res = append(res, "UseExternalEmojis")
	}
	if perms&discordgo.PermissionUseExternalStickers == discordgo.PermissionUseExternalStickers {
		res = append(res, "UseExternalStickers")
	}
	if perms&discordgo.PermissionUseApplicationCommands == discordgo.PermissionUseApplicationCommands {
		res = append(res, "UseApplicationCommands")
	}
	if perms&discordgo.PermissionUseEmbeddedActivities == discordgo.PermissionUseEmbeddedActivities {
		res = append(res, "UseEmbeddedActivities")
	}
	// ...26 more if checks...
}
```
</td>
<td>

```go
var legacyPermNames = map[int64]string{
	discordgo.PermissionManageGuild:  "ManageServer",
	discordgo.PermissionViewChannel:  "ReadMessages",
	discordgo.PermissionViewAuditLog: "ViewAuditLogs",
}

func HumanizePermissions(perms int64) (res []string) {
	for _, p := range discordgo.AllPermissions {
		if perms&p == p {
			if legacyName, ok := legacyPermNames[p]; ok {
				res = append(res, legacyName)
			} else {
				res = append(res, discordgo.PermissionName(p))
			}
		}
	}

	return
}
```
</td>
</tr>
</table>

and a similar simplification can be made for building the `Permissions` map in custom command context.

*The new implementation is actually longer than it really should be due to backward compatibility reasons. If these can be ignored the updated code can be <5 lines long and self-documenting to boot.


common/util.go
```diff
@@ -295,7 +295,7 @@ func RemoveRoleDS(ms *dstate.MemberState, role int64) error {
 }
 
 var StringPerms = map[int64]string{
-	discordgo.PermissionReadMessages:       "Read Messages",
+	discordgo.PermissionViewChannel:        "View Channel",
```
jo3-l, line 298: I didn't bother keeping the old names for backward compatibility because `StringPerms` is only used in the web UI and there's nothing bad that can happen if that changes a little.

lib/discordgo/tools/cmd/permnames/main.go
```diff
@@ -0,0 +1,129 @@
+package main
```
jo3-l, line 1: This code is admittedly complicated but the current discordgo code already relies on some codegen (`eventhandlers`) and we need not touch the code ever again after it is committed.

lib/discordgo/permission.go
```diff
@@ -1,53 +1,99 @@
...
-	PermissionUsePrivateThreads       int64 = 0x01000000000                      // (1 << 36)	Allows for creating and participating in private threads	T
-	PermissionUseExternalStickers     int64 = 0x02000000000                      // (1 << 37)	Allows the usage of custom stickers from other servers	T
-	PermissionSendMessagesInThreads   int64 = 0x04000000000                      // (1 << 38)   Allows for sending messages in threads	T
-	PermissionUseEmbeddedActivities   int64 = 0x08000000000                      // (1 << 39)   Allows for using Activities (applications with the EMBEDDED flag) in a voice channel	V
-	PermissionModerateMembers         int64 = 0x10000000000                      // (1 << 40)   Allows for timing out users to prevent them from sending or reacting to messages in chat and threads, and from speaking in voice and stage channels
-
-	// all bits set except the leftmost to avoid using negative numbers in case discord dosen't handle it
-	PermissionAll int64 = int64(^uint64(0) >> 1)
+	PermissionCreateInstantInvite     int64 = 0x00000000001 // (1 << 0)		Allows creation of instant invites	T, V, S
```
mrbentarikau, line 4: just interested, why not use 1 << 0 instead, and not 0x....001 and so on

jo3-l: I don't know why the existing code uses hexadecimal either. My best guess is that it is copied from the [permission flags table in the Discord API documentation](https://discord.com/developers/docs/topics/permissions), but that also provides the left-shift expressions so ¯\\\_(ツ)\_/¯. I personally would have written this using right shifts, but seeing that the diff for this PR is large enough already I figured it'd be nice to not switch too many things up.

mrbentarikau: maybe add the ones after 1 << 40 as well from official docs, for completeness sake

jo3-l: Good point, thanks. I originally thought that the only ones missing were the new poll-related permissions so was planning on adding those in another PR for poll support in general, but seeing we're missing quite a few I will go ahead and add those tomorrow.

lib/dstate/permissions.go
```diff
@@ -8,9 +8,9 @@ const AllPermissions int64 = discordgo.PermissionAll
 // discord performs no server side validation so this is needed
 // as to not run into some really weird situations
 const ChannelPermsMask = ^(discordgo.PermissionAdministrator |
-	discordgo.PermissionManageServer |
+	discordgo.PermissionManageGuild |
 	discordgo.PermissionChangeNickname |
-	discordgo.PermissionManageServer |
+	discordgo.PermissionManageGuild |
```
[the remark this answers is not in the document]
jo3-l: Thanks. `PermissionManageServer` was duplicated in the original code, so this PR did not introduce the issue, but now that you point it out I have fixed it in https://github.com/botlabs-gg/yagpdb/pull/1649/commits/861f5fc75da32df385cc41b8b16fd0fc9f450d85.

lib/discordgo/permission.go
jo3-l: I added the new permissions in https://github.com/botlabs-gg/yagpdb/pull/1649/commits/d1a828d8d370ff401cc75d151a21a2c733a2d1fb, which is also a nice example of the benefits of this refactor (per commit description.)

While adding these new permissions, I thought about the issue re: hexadecimal you brought up earlier some more. Originally, like I said, I planned on keeping the code format untouched, but adding the new values in hexadecimal was rather inconvenient since Discord API's doc quotes them to 15 digits: `0x0002000000000000`, for instance. This messed up the alignment and made lines stretch to ridiculous levels after `gofmt`'ing, so I decided to switch everything to use left-shifts in https://github.com/botlabs-gg/yagpdb/pull/1649/commits/4b38bf1cc6cc849af59d3dc13c3a6033e25b85df.

Related: I did some digging in the commit history of `bwmarrin/discordgo` to figure out why they are using hexadecimal values, and found https://github.com/bwmarrin/discordgo/commit/843c765ae3eeecaf8e9b808ea8e3da10ba600f91. They say
> We define our permissions constants based on Discord's values,
and thus the use of `iota` is fragile, makes docs harder to read,
and makes changes touching these constants harder to review.
Here, we now use the full 64-bit representation of each constant,
which both removes the implicit ordering dependencies and makes
the values easier to immediately see.

I agree with the point re: no `iota` -- it makes the code very fragile and is prone to cause issues when changed -- for instance, `PermissionSendPolls = 1 << 49` is right after `PermissionSendVoiceMessages = 1 << 46` and using `iota` carelessly would erroneously set it to `1 << 47`. However, I think that using the explicit left-shift representation still provides the benefits described (easy to review and to immediately see values.)

jo3-l: I checked that I did not make any copy-paste errors while switching to the left-shift representation using the following script: https://go.dev/play/p/zlzrZOZquF7. The permission constants are copied from the new file and the series of Println checks are generated by taking the old file and substituting `s/(\w+)\s+int64 = (0x\d+)/fmt.Println("$1", $1 == $2)/g`.

mrbentarikau: looks good and works on my testing, only minor issue I have are file names > permissions.go (like dstate has) and permission_names.go, plural just feel better to me, but this is non-issue

ashishjh-bst (approved):

PR #1651, opened 2024-05-22
soundboard: remove stray references to gorm

jo3-l:
The soundboard module formerly used gorm, but was migrated to sqlboiler some years ago in https://github.com/botlabs-gg/yagpdb/commit/628ea9a228ab11dcc327ad0d017c5654312025af. There are still two (erroneous) references to gorm remaining which were not caught since this section of code ignores errors. This PR changes these to use sqlboiler as well.

This PR is marked as a DRAFT since I do not yet have soundboard set up locally, so have not tested this change. It is put up so that 1) I don't forget about it and 2) in case any folks who do have soundboard working have time to test and report their findings. If not, I will set up soundboard eventually and verify that this PR works.


jo3-l:
I ran soundboard on my selfhost just now and confirmed this PR works as expected.

PR #1652, opened 2024-05-22
reminders: migrate away from gorm to sqlboiler

jo3-l:
This is 1/n PRs switching the codebase away from gorm to sqlboiler.

Best reviewed commit-by-commit. The database schema code (first commit) should be reviewed especially carefully, since the relevant queries will be run automatically on startup.

## On `NOT NULL` constraints

By default, gorm does not add `NOT NULL` constraints to columns not explicitly marked as such. However, a number of columns in our gorm models across the codebase are *effectively* non-nullable -- for instance, the `reminders.user_id` column is never empty in valid data, but is nonetheless nullable in the table definition. This complicates migration to sqlboiler, since if the current table definition is kept, sqlboiler will faithfully reflect the state of the database and generate models with nullable field types. But as our current codebase implicitly assumes many fields to be non-null, working with sqlboiler models thus becomes quite convoluted.

To address this issue, we have decided to retroactively add `NOT NULL` constraints when applicable to existing tables. This will be done automatically at startup once these changes are applied using the existing database migration machinery. **Self-hosters upgrading to a release containing the changes here should therefore be ready for a delayed first restart, as adding new `NOT NULL` constraints require Postgres to acquire an exclusive lock on and fully scan tables to check that no entries contain null columns.**


reminders/plugin_bot.go
```diff
@@ -220,70 +218,33 @@ var cmds = []*commands.YAGCommand{
...
-			delMsg := fmt.Sprintf("Deleted reminder **#%d**: '%s'", reminder.ID, limitString(reminder.Message))
-
-			// If there is another reminder with the same timestamp, do not remove the scheduled event
-			for _, v := range currentReminders {
-				if v.When == reminder.When {
-					return delMsg, nil
-				}
-			}
-
-			return delMsg, nil
```
jo3-l, line 244: The comment here implies that the scheduled event to send the reminder will be removed if there exists no reminder with the same timestamp, but this is not the case. The current code simplifies fully to `return delMsg, nil`, so there is no functional change despite appearances.

Per the new comment added above, it is perfectly fine to not delete the scheduled event and just remove the database entry.

reminders/plugin_bot.go
```diff
@@ -220,70 +218,33 @@ var cmds = []*commands.YAGCommand{
...
+			return fmt.Sprintf("Deleted reminder **#%d**: '%s'", reminder.ID, CutReminderShort(reminder.Message)), nil
 		},
 	},
 }
 
-func stringReminders(reminders []*Reminder, displayUsernames bool) string {
```
jo3-l, line 249: This was moved to `reminders.go` and renamed to `DisplayReminders`.

reminders/plugin_bot.go
```diff
@@ -220,70 +218,33 @@ var cmds = []*commands.YAGCommand{
...
 	if err != nil {
 		return true, err
 	}
 
-	now := time.Now()
-	nowUnix := now.Unix()
-	for _, v := range reminders {
-		if v.When <= nowUnix {
-			err := v.Trigger()
+	// TODO: can we move this filtering step into the database query?
```
jo3-l, line 243: I think we can, but it is probably preferable to minimize non-database refactors (that are not obviously correct) to prevent introducing bugs. Perhaps we can try this refactor in a future version, once the gorm -> sqlboiler transition is complete.

reminders/reminders.go
```diff
@@ -1,27 +1,30 @@
...
 	"github.com/botlabs-gg/yagpdb/v2/common/mqueue"
 	"github.com/botlabs-gg/yagpdb/v2/common/scheduledevents2"
 	"github.com/botlabs-gg/yagpdb/v2/lib/discordgo"
-	"github.com/jinzhu/gorm"
+	"github.com/botlabs-gg/yagpdb/v2/reminders/models"
 	"github.com/sirupsen/logrus"
+	"github.com/volatiletech/sqlboiler/v4/boil"
 )
 
+//go:generate sqlboiler --no-hooks --add-soft-deletes psql
```
jo3-l, line 19: I was surprised to see that reminders currently soft-delete instead of hard-delete. That is, once triggered, they remain in the database with the `deleted_at` field set. I am unsure whether this is an oversight -- `gorm` does not make it particularly obvious that soft-delete is the default behavior -- or intended.

This PR currently continues to soft-delete, but we could easily switch to hard deletions.

reminders/schema.go
```diff
@@ -0,0 +1,58 @@
...
+
+-- Delete invalid data (should not exist, but just to be sure.)
+DELETE FROM reminders
+WHERE created_at IS NULL
+	OR updated_at IS NULL
+	OR user_id IS NULL
+	OR channel_id IS NULL
+	OR guild_id IS NULL
+	OR message IS NULL
+	OR "when" IS NULL;
```
jo3-l, line 57: This comment needs to resolved before this PR is merged.

lib/discordgo/discord.go
```diff
@@ -153,3 +154,7 @@ func CheckRetry(_ context.Context, resp *http.Response, err error) (bool, error)
 func StrID(id int64) string {
 	return strconv.FormatInt(id, 10)
 }
+
+func ParseID(s string) (int64, error) {
```
jo3-l, line 158: See description of third commit for why this alias was added.

reminders/schema.go
jo3-l: Resolved: some old entries in production prior to November 2018 have `guild_id = NULL`, since the reminders table did not contain a guild_id column until https://github.com/botlabs-gg/yagpdb/commit/9f5ef28d6fbab191775dbc85e6e83d41c53248b5 was deployed. Luckily, due to the timeline, all such 'legacy' entries correspond to expired reminders that can be deleted. So the latest commit runs a query that -- if the `NOT NULL` constraint has not yet been added -- deletes all expired entries with `guild_id = NULL` and only then adds the constraint. There is no need for additional effort (e.g., manual queries) from self-hosters.

jo3-l:
I think this PR is ready to review and merge, when appropriate.

PR #1658, opened 2024-05-30
moderation: migrate away from gorm to sqlboiler

jo3-l:


moderation/commands.go
```diff
@@ -878,19 +880,19 @@ var ModerationCommands = []*commands.YAGCommand{
 			}
 
 			if parsed.Switches["id"].Value != nil {
-				var warn []*WarningModel
-				err = common.GORM.Where("guild_id = ? AND id = ?", parsed.GuildData.GS.ID, parsed.Switches["id"].Int()).First(&warn).Error
-				if err != nil && err != gorm.ErrRecordNotFound {
+				id := parsed.Switches["id"].Int()
+				warning, err := models.FindModerationWarningG(parsed.Context(), id)
```
jo3-l, line 884: The previous query filtered by both `guild_id` and `id`, but this turns out to be unnecessary as the `id` column is the sole PK of the table and so uniquely identifies the warning -- it is not per-guild.

jo3-l:
I am tentatively marking this PR as ready to review. The obvious mistakes that I found have been fixed and the features I have tested so far do work. But as the previous (incorrect) comment demonstrated, there are some subtle aspects here that need to be reviewed and tested very carefully.

PR #1663, opened 2024-06-10
Adds ability for automod to detect duplicate messages across channels

KTStephano:
# Changes

* **lib/dstate/inmemorytracker** now maintains 1 extra "channel" worth of message data per guild. This is an ordered snapshot of most recent messages from any channel. Its memory footprint is limited by `shard.conf.ChannelMessageLen`.
* **lib/dstate/interface's** `GetMessages` returns messages matching query for a specific channel as before. However, if channel id is passed in as 0, it will check against the guild message snapshot.
* **automod/triggers's** SpamTrigger was updated to have a "Match across channels" option. The default value is `false` which causes the trigger to work the same way it does now (each channel's message events isolated from all others).

Screenshot 2024-06-10 at 4 35 13 PM

# Use Case

Someone copy+pastes the same spam message very quickly, but across more than channel. This allows automod to detect this and take action.

This change will also make it possible for other automod rules to match across channels in the future.

### DIFF
--- a/automod/triggers.go
+++ b/automod/triggers.go
@@ -996,9 +996,10 @@ func (r *MessageRegexTrigger) CheckMessage(triggerCtx *TriggerContext, cs *dstat
 /////////////////////////////////////////////////////////////
 
 type SpamTriggerData struct {
-	Treshold     int
-	TimeLimit    int
-	SanitizeText bool
+	Treshold          int
+	TimeLimit         int
+	SanitizeText      bool
+	CrossChannelMatch bool
 }
 
 var _ MessageTrigger = (*SpamTrigger)(nil)
@@ -1045,6 +1046,12 @@ func (spam *SpamTrigger) UserSettings() []*SettingDef {
 			Kind:    SettingTypeBool,
 			Default: false,
 		},
+		{
+			Name:    "Match duplicates across channels",
+			Key:     "CrossChannelMatch",
+			Kind:    SettingTypeBool,
+			Default: false,
+		},
 	}
 }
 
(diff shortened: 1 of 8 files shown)


jo3-l:
At a glance, I am wondering whether there is a way we can reduce the memory overhead here/implement this while retaining the original storage representation. For instance, would it be feasible to instead call the existing `GetMessages(channelID, limit)` procedure multiple times for each channel in the guild (with limit scaled down appropriately?)

KTStephano:
@jo3-l Right now the only way I can think to make it work with the existing `GetMessages` would be to iterate over all channels in the guild (with scaled down limit) and merge them into a single list sorted by timestamp. Then it could walk the resulting list and see if the user posted more than X duplicates. Would bot perf be impacted a lot for bigger guilds?

I do think the memory footprint could be reduced though. After thinking more about it I could revert to the previous message representation without the map of maps, but add a single new list per guild keeping track of some small snapshot of most recent messages. Then `GetMessage` with channel id 0 could still return sorted messages from the whole guild. For the memory footprint it would be as if the bot was tracking 1 extra channel per guild with id 0.

jo3-l:
> After thinking more about it I could revert to the previous message representation without the map of maps, but add a single new list per guild keeping track of some small snapshot of most recent messages.

This implementation sounds good (and preferable to the current one in this PR) to me. That said I would wait for the developer @ashishjh-bst to chime in before making any changes in case he has an alternate suggestion or preference.

ashishjh-bst:
> @jo3-l Right now the only way I can think to make it work with the existing `GetMessages` would be to iterate over all channels in the guild (with scaled down limit) and for each one walk the message list and count up the number of duplicates. Would bot perf be impacted for bigger guilds with lots of activity and lots of channels?
> 
> I do think the memory footprint could be reduced though. After thinking more about it I could revert to the previous message representation without the map of maps, but add a single new list per guild keeping track of some small snapshot of most recent messages. Then `GetMessage` with channel id 0 could still return sorted messages from the whole guild. For the memory footprint it would be as if the bot was tracking 1 extra channel per guild with id 0.

This would most certainly cause issues in yagpdb.xyz, and will most likely get ratelimited too.

KTStephano:
@ashishjh-bst That makes sense. I went with option 2 since the memory footprint increase is very small per guild.

ashishjh-bst (approved):

PR #1656, opened 2024-05-29
feat: Add CodeMirror editor to control panel

sum117:
# Code Mirror Editor for Control Panel
image

This PR adds CodeMirror to YAGPDB bot, by implementing the unused vendor package `CodeMirror`. There are style inconsistencies, but an admin told me it was alright, as users wouldn't care. I included a CSS file of a theme that I like, downloaded from their website directly. It should function as reference if someone ever wants to write the CSS to match the style of the highlight.js that is already included in this page [Permalink](https://github.com/botlabs-gg/yagpdb/blob/447a52732f5b901314397fe9079290e59d9f8d45/customcommands/assets/customcommands.html#L193).

## Why did I alter `spongebob.js`

I had to alter this file because I used a go template inside my code to access the keywords, and since a new textarea is created every time a button is clicked, I removed the jquery from `spongebob.js` and rewrote it in modern syntax, implementing the code mirror instance every click.

## About themes

I think it'll take a while for the team to decide which theme should be used, or maybe not, idk. In case you guys didn't like the theme I chose, just pick one from [here](https://codemirror.net/5/theme/) and i'll update the code.

My Discord is currently `shar_alter`.

PR #1676, opened 2024-06-20
cc: ensure additional responses are saved properly

jo3-l:
Currently, if one uses the `+` button to create a new response for a custom command, then save, only the first response is persisted: that is, additional responses are lost and not saved.

This bug is a regression from the previous version. I bisected it to the CodeMirror change, #1656. The issue is that we are not saving changes to the underlying textarea for newly added CodeMirror instances, so the `POST`ed form always contains empty responses (save the first one.) This PR therefore patches `setupCodeMirrorInstance` to not merely initialize the CodeMirror instance but also attach a change event listener.

@ashishjh-bst perhaps we could consider a cherry-pick to the current release branch, given that this bug does result in lost data if one attempts to add an additional custom command response.


ashishjh-bst:
Closing this as I have cherry picked the commit into another PR.

PR #1677, opened 2024-06-20
templates: block nested template executors

SoggySaussages:
This PR blocks looping `exec` or `execAdmin` on commands which parse a template (such as moderation actions' Punish
DM or the opening message in a ticket).

Upon executing one of these commands, a new context flag marks it as executed from a command template. If within this 
new template one of these commands is used again, its context gets a flag marking it as executed from a *nested*
command template. Any of the functions executed within a nested command template will fail.

This PR also blocks using `execCC` within any of these commands' templates. This will likely soon be mollified by the addition of an 'autoScheduleExecCC' function suggested by @jo3-l which will allow users to once again execute CCs
within these commands' templates in a manageable and rate limited fashion.

### DIFF
--- a/automod/effects.go
+++ b/automod/effects.go
@@ -269,7 +269,7 @@ func (kick *KickUserEffect) Apply(ctxData *TriggeredRuleData, settings interface
 		reason += ctxData.ConstructReason(true)
 	}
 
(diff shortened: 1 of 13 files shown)


jo3-l:
@SoggySaussages, you may want to rebase this onto dev.

SoggySaussages:
> @SoggySaussages, you may want to rebase this onto dev.

Thx, donezo

PR #1681, opened 2024-06-22
Disable custom command groups (Suggestion #2224)

Borbot33:
# What's changed?

- Small warning if the command group is disabled
- A switch for disabling the custom commang group

[Suggestion #2224](https://discord.com/channels/166207328570441728/356486960417734666/1246836065872908298)

# Screenshots
image
screenshot 2

### DIFF
--- a/customcommands/assets/customcommands.html
+++ b/customcommands/assets/customcommands.html
@@ -108,7 +108,7 @@
                                 </div>
                                 <div class="form-group">
                                     <label>Group disabled</label><br>
-                                    {{checkbox "disabled" "disabled" "Enables or Disables the group" .CurrentCommandGroup.Disabled }}
+                                    {{checkbox "disabled" "disabled" "Enables or Disables the group" .CurrentCommandGroup.Disabled}}
                                 </div>
                                 <div class="form-group">
                                     <button type="submit"
--- a/customcommands/assets/customcommands.html
+++ b/customcommands/assets/customcommands.html
@@ -47,7 +47,7 @@
                     class="nav-item {{if $dot.CurrentCommandGroup}}{{if eq $dot.CurrentCommandGroup.ID .ID}}active{{end}}{{end}}">
                     <a data-partial-load="true"
                         class="nav-link show {{if $dot.CurrentCommandGroup}}{{if eq $dot.CurrentCommandGroup.ID .ID}}active{{end}}{{end}}"
-                        href="/manage/{{$dot.ActiveGuild.ID}}/customcommands/groups/{{.ID}}">{{.Name}}</a>
+                        href="/manage/{{$dot.ActiveGuild.ID}}/customcommands/groups/{{.ID}}">{{.Name}}{{if .Disabled}} <code>Disabled</code>{{end}}</a>
                 </li>
                 {{end}}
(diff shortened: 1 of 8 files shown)


customcommands/assets/customcommands.html
```diff
@@ -106,6 +106,10 @@ <h2>Custom commands</h2>
                                         {{textChannelOptionsMulti .ActiveGuild.Channels .CurrentCommandGroup.IgnoreChannels }}
                                     </select>
                                 </div>
+                                <div class="form-group">
+                                    <label>Group disabled</label><br>
+                                    {{checkbox "disabled" "disabled" "Enables or Disables the group" .CurrentCommandGroup.Disabled }}
```
jo3-l, line 111: nit
```suggestion
                                    {{checkbox "disabled" "disabled" "Enables or Disables the group" .CurrentCommandGroup.Disabled}}
```

jo3-l (changes requested):
You should load the related group then use `cmd.R.Group.Disabled` instead of adding `getDisabledGroups`, which fetches all disabled groups even if we only need information about one.

Borbot33:
> You should load the related group then use `cmd.R.Group.Disabled` instead of adding `getDisabledGroups`, which fetches all disabled groups even if we only need information about one.

done :)

customcommands/web.go
```diff
@@ -643,11 +644,16 @@ func handleRunCommandNow(w http.ResponseWriter, r *http.Request) (web.TemplateDa
 	}
 
 	// only interval commands can be ran from the dashboard currently
-	cmd, err := models.CustomCommands(qm.Where("guild_id = ? AND local_id = ? AND trigger_type = 5", activeGuild.ID, cmdID)).OneG(context.Background())
+	cmd, err := models.CustomCommands(qm.Where("guild_id = ? AND local_id = ? AND trigger_type = 5", activeGuild.ID, cmdID), qm.Load("Group")).OneG(context.Background())
 	if err != nil {
 		return templateData, err
 	}
 
+	if cmd.GroupID.Int64 != 0 && cmd.R.Group.Disabled {
```
jo3-l, line 652: (untested)
```suggestion
	if cmd.R.Group != nil && cmd.R.Group.Disabled {
```
is probably less brittle and ditto everywhere else.

jo3-l (commented):
Nice, LGTM other than one minor nit

customcommands/tmplextensions.go
```diff
@@ -180,11 +180,20 @@ func tmplRunCC(ctx *templates.Context) interface{} {
 			return "", templates.ErrTooManyCalls
 		}
 
-		cmd, err := models.FindCustomCommandG(context.Background(), ctx.GS.ID, int64(ccID))
+		// cmd, err := models.FindCustomCommandG(context.Background(), ctx.GS.ID, int64(ccID))
+		// if err != nil {
+		// 	return "", errors.New("Couldn't find custom command")
+		// }
+
```
jo3-l, line 187: You can remove this commented-out code completely, and likewise below. Otherwise everything looks OK to me.

Borbot33: yup :)

PR #1682, opened 2024-06-22
Move the base64 & sha256 functions to general.go and change name of sha256 function to hash

Borbot33:


jo3-l:
From my UI, your last commit does not contribute to the overall diff: that is, this PR does not seem to delete the context functions you added. Is that a visual issue on my side, or is it the case for you as well (check the "Files changed" tab?) If the latter, try rebasing on current dev; I suspect your base branch is outdated.

common/templates/general.go
```diff
@@ -1687,3 +1689,24 @@ func tmplHumanizeDurationSeconds(in interface{}) string {
...
+
+func tmplEncodeBase64(str string) string {
+	return base64.StdEncoding.EncodeToString([]byte(str))
+}
+
+func tmplSha256(str string) string {
+    hash := sha256.New()
+    hash.Write([]byte(str))
+
+    sha256 := base64.URLEncoding.EncodeToString(hash.Sum(nil))
```
jo3-l, line 1709: You may also want to run this change by gofmt: these lines are indented with spaces instead of tabs like the rest of the code.

PR #1696, opened 2024-07-04
moderation: explicitly separate config functions that access bot process cache

jo3-l:
This PR is one of several aiming to mitigate the impact of bugs similar to those introduced in the previous release.

Another issue introduced (again, by me) in the previous release was that the web handlers for both the moderation and notifications packages called `GetCachedConfig` instead of `GetConfig`. `GetCachedConfig` is only meant to called in the bot process; using it in the web process caused the following cache invalidation issue:
1. `GetCachedConfig` in web process creates a config cache on the web process, separate from that maintained on the bot process.
2. When changes to the config are saved via the control panel, a pubsub event indicating that the config cache should be invalidated is fired.
3. However, only the bot process is configured to handle pubsub events, so only the cache on the bot process is invalidated. The cache inadvertently created on the web process, on the other hand, becomes stale and so changes do not reflect on the control panel.

To try to prevent this issue in future, we make the following (automated) renames:
- `GetCachedConfigOrDefault` becomes `BotCachedGetConfig`, following the convention in the `streaming` plugin. The prefix `BotCached` is more easily noticed and its presence in web code should raise a bigger alert during review.
- `GetConfig` becomes `FetchConfig` (again, trying to make it as different from the bot cached variant as possible) and is moved out of plugin_bot.go to config.go.

## Testing strategy

This change is (almost) completely done with automated renaming tools so there is little risk of a copy-paste mistake: the only manual input was moving `GetConfig` and `SaveConfig` from plugin_bot.go to config.go.

I also tested this change locally with the bot and web processes ran separately and verified that the moderation configuration saves properly on the control panel and that the changes reflect on the bot process (for instance, by disabling the `timeout` command on the control panel and then attempting to run the command in Discord.)

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -27,7 +27,7 @@ import (
 )
 
 func MBaseCmd(cmdData *dcmd.Data, targetID int64) (config *Config, targetUser *discordgo.User, err error) {
-	config, err = GetConfigOrDefault(cmdData.GuildData.GS.ID)
+	config, err = FetchConfig(cmdData.GuildData.GS.ID)
 	if err != nil {
 		return nil, nil, errors.WithMessage(err, "GetConfig")
 	}
--- a/automod/automod_web.go
+++ b/automod/automod_web.go
@@ -506,7 +506,7 @@ func (p *Plugin) handlePostAutomodUpdateRule(w http.ResponseWriter, r *http.Requ
 		}
 	}
 	if anyMute {
-		conf, err := moderation.GetConfigOrDefault(g.ID)
+		conf, err := moderation.FetchConfig(g.ID)
 		if err != nil || conf.MuteRole == 0 {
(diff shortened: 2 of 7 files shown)


moderation/commands.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func MBaseCmd(cmdData *dcmd.Data, targetID int64) (config *Config, targetUser *discordgo.User, err error) {
-	config, err = GetConfigOrDefault(cmdData.GuildData.GS.ID)
+	config, err = FetchConfig(cmdData.GuildData.GS.ID)
```
jo3-l, line 30: Commands run on the bot process and so I think it is safe (and a performance optimization) to call `BotCachedGetConfig` here. I have not made that change in this PR in the interest of keeping its scope to automated, obviously safe changes only.

PR #1700, opened 2024-07-15
dstate: fully populate member struct

l-zeuch:
Please note that this patch is ~~entirely~~ partially untested on my part, 
because I lack a member boosting the server ~~and one with flags set.~~ I have
reasonably high confidence that this will work, but I cannot guarantee
it.

Fully populate the discordgo.Member struct in ms.DgoMember(); prior to
this change, this function did not set PremiumSince nor Flags, thereby
rendering these fields effectively useless in custom commands.

They are present everywhere else, and are documented in our custom
command documentation as available, so it appears as if these fields
have been left out by accident, rather than on purpose.

### DIFF
--- a/lib/dstate/interface.go
+++ b/lib/dstate/interface.go
@@ -381,6 +381,8 @@ func (ms *MemberState) DgoMember() *discordgo.Member {
 		Roles:                      ms.Member.Roles,
 		User:                       &ms.User,
 		Pending:                    ms.Member.Pending,
+		PremiumSince:               ms.Member.PremiumSince,
+		Flags:                      ms.Member.Flags,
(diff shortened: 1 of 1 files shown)


l-zeuch:
Quick addendum: I have modified one of my members to carry the
`BYPASSES_VERIFICATION` member flag (`1<<2`); this now works as expected
in custom commands. Other flags are not modifiable, as documented.

To test this, I ran the following request:
```
curl -fsSL -X PATCH https://discord.com/api/v9/guilds/{guild_id}/members/{user_id} -H "Authorization: $YAGPDB_BOTTOKEN" --data '{"flags": 4}' -H "content-type: application/json"
```

PR #1705, opened 2024-07-17
templates: implement Del method for slices

l-zeuch:
Implement a Del method for custom command slices; to be consistent with
the other methods, this also returns a new slice. The usage is as
follows:

```
{{ (cslice 1 2 3 4 5).Del 2 }}
{{/* returns [1 2 4 5] */}}
```

See the following suggestion on the support server:
https://canary.discord.com/channels/166207328570441728/356486960417734666/764864357917392906

----
This is my first time working with the `reflect` package—I figured this
is the way to go, given how the other methods also use `reflect`. Please
let me know if I did something wrong or if there is a better way to do
this.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -111,9 +111,9 @@ var (
 		"roleAbove":   roleIsAbove,
 		"seq":         sequence,
 
-		"shuffle": shuffle,
-		"verb":    common.RandomVerb,
-		"hash":    tmplSha256,
+		"shuffle":      shuffle,
+		"verb":         common.RandomVerb,
+		"hash":         tmplSha256,
 		"decodeBase64": tmplDecodeBase64,
 		"encodeBase64": tmplEncodeBase64,
 
@@ -955,6 +955,16 @@ func (s Slice) Append(item interface{}) (interface{}, error) {
 		result := reflect.Append(reflect.ValueOf(&s).Elem(), reflect.ValueOf(v))
 		return result.Interface(), nil
 	}
+}
+
+func (s Slice) Del(index int) (interface{}, error) {
+	if index >= len(s) {
+		return nil, errors.New("Index out of bounds.")
+	}
+
+	result := reflect.ValueOf(&s).Elem()
+	result = reflect.AppendSlice(result.Slice(0, index), result.Slice(index+1, result.Len()))
+	return result.Interface(), nil
 }
 
 func (s Slice) Set(index int, item interface{}) (string, error) {


common/templates/context.go
```diff
@@ -957,6 +957,16 @@ func (s Slice) Append(item interface{}) (interface{}, error) {
 	}
 }
 
+func (s Slice) Del(index int) (interface{}, error) {
+	if index >= len(s) {
```
jo3-l, line 961: `index < 0` needs to be checked as well; the fact that Set leaves this out is a mistake (and results in a runtime panic, though one that is recovered.)

common/templates/context.go
```diff
@@ -957,6 +957,16 @@ func (s Slice) Append(item interface{}) (interface{}, error) {
...
+		return nil, errors.New("Index out of bounds.")
```
jo3-l, line 962:
```suggestion
		return nil, errors.New("index out of bounds")
```

common/templates/context.go
```diff
@@ -957,6 +957,16 @@ func (s Slice) Append(item interface{}) (interface{}, error) {
...
+	}
+
+	result := reflect.ValueOf(&s).Elem()
+	result = reflect.AppendSlice(result.Slice(0, index), result.Slice(index+1, result.Len()))
+	return result.Interface(), nil
```
jo3-l, line 967: Reflection is not needed (and `Append` and `AppendSlice` use it gratuitously.)
```suggestion
	// Clip the capacity of the left slice to force a copy on append, for consistency with Append and AppendSlice
	// (which return copies.)
	return append(s[:index:index], s[index+1:]...)
```
should be sufficient.

jo3-l (commented):
Though I understand this suggestion is popular, I have never understood the use-case for this exact function as written. Deleting one and only one item from a list *by index* is not an operation I have reached for often in any code[^1], especially given that it is inefficient by design—there are often better data structures that are easier to understand. Note that until 1.21, Go did not provide a shortcut for this operation either (one had to write `append(s[:i], s[i+1:]...)`.)

On the other hand, I empathize more with the need to delete an item by value (i.e., `$xs.Del 2` removes the item with value 2.) If anything, I wonder if this API is the one we should provide instead. I am also worried about potential confusion between "delete by value" and "delete by index" (i.e., would it be clear to users whether `$xs.Del 2` deletes the value `2` or the item at index `2`?), and wonder whether a name other than `Del` would be clearer. 

For reference, as a short survey of what other programming languages call this operation:
- Go 1.21 has `slices.Delete(i, j int)` which accepts *two* indices (specifying the subslice to remove); the single-index version was, IIRC, explicitly rejected for fear of potential misuse;
- C# has `List.RemoveAt(int index)`, with the `At` suffix (I like this) and also `List<T>.Remove(T value)` for deletion by value;
- Python has special syntax for deletion by index, `del list[index]` (which is very clear), and also has a `list.remove` function that deletes by value;
- JavaScript, always the odd one out, has `Array.splice(index, deleteCount)`;
- Rust has `Vec.remove(index: usize)`;
- Java has `List.remove(int index)`.

Out of these, I like `RemoveAt` (or potentially `DelAt`) the most. In any case, the variety of names and existing APIs strongly indicates to me that we need to put more thought into what deletion operations we should add here, if any. To this end, I would like to see some concrete use-cases in real code before going forward with extending a fundamental data structure.

[^1]: On the other hand, I have often wanted to *filter* a list by some criteria (i.e., delete elements that do not match some predicate), but such an API is difficult to model within the limitations of the template system.

l-zeuch:
Thanks for the comment and breakdown of various similar APIs.

Personally, I haven't reached for deleting by index either, nor can I
think of a situation where I would—as you said, there are data
structures that are better suited for that. I only ever had to delete a
certain value from the slice., never said "gee, let me delete that
element at position X without knowing what it is".

In whole, I agree that the name `Del` is suboptimal. Your proposed
`DelAt` or `RemoveAt` are much clearer about what they do, but as we
kind of established, they don't make much sense.

In a similar fashion, I propose we rename this method to `DelValue` or
similar, to clearly state "this deletes by a specific value" and
consequently implement it as such.

As for your review, ironically my initial solution did not use
reflection at all, but I felt this wasn't correct, given how Append and
AppendSlice use it (unnecessarily so, as you pointed out).

I'll keep this PR open for discussion purposes and the subsequent rework
of this proposed API.

l-zeuch:
Internal discussion has concluded that this isn't something that makes sense, no matter how we twist it.

There are already idomatic ways to achieve this that are also present in a handful of mainstream languages, specifically the "iterate and then delete" approach.

PR #1707, opened 2024-07-20
make tmplGetChannel return threads too

ashishjh-bst:
QOL improvement, otherwise you need to do a stupid check to see if the the channel was a thread or not, this can get difficult if things need to be done in a non-triggering channel or the channel-id is coming from db.


ashishjh-bst:
nvm, getChannelOrThread already exists.

PR #1709, opened 2024-07-21
cc: fix tmpl exec on ephemeral context messages

jo3-l:
Currently, `exec` and `execAdmin` do not work as expected (often resulting in runtime panics) when triggered from an interaction on an ephemeral message.

## Observing the bug

<details>

<summary>Click to expand reproduction steps</summary>

(For convenience, the necessary custom commands can be imported as https://yagpdb.xyz/cc/olWroLOPY0 and https://yagpdb.xyz/cc/ecfP_ypYhY.)

1. Create a custom command `-bug` that replies with a button:
    ```
    {{ sendMessage nil (complexMessage "buttons" (cbutton "label" "Click me!" "custom_id" "clickme")) }}
    ```
2. Create an interaction custom command `clickme(_ephemeral)?` that replies with an ephemeral message containing another button when the previous button is clicked; and triggers `exec "whois" .User` when this ephemeral button is clicked:
    ```
    {{ if in .CustomID "ephemeral" }}
        {{ sendResponse nil (exec "whois" .User) }}
    {{ else }}
        {{ sendResponse nil (complexMessage
            "buttons" (cbutton "label" "Click me (ephemeral)!" "custom_id" "clickme_ephemeral")
            "ephemeral" true
        ) }}
    {{ end }}
    ```
3. Run `-bug`, click the first button, click the second button, and observe a runtime panic.

</details>

repro illustrating runtime panic

## Analysis

The issue stems from the fact that Discord sends all ephemeral messages with no `guild_id` field, even if they are sent in guild channels (see  https://github.com/discord/discord-api-docs/issues/4557). Hence the guild ID field of the context message is `0` for interactions triggered on ephemeral messages. However, `exec` and `execAdmin` infer guild data based on the guild ID of the context message:
https://github.com/botlabs-gg/yagpdb/blob/b9dabf9eba7d45c1189b695c55076f3f9669c022/commands/tmplexec.go#L186
and in `FillDataLegacyMessage`,
https://github.com/botlabs-gg/yagpdb/blob/b9dabf9eba7d45c1189b695c55076f3f9669c022/lib/dcmd/system.go#L221-L239

So the data passed to dcmd has nil `GuildData`, causing various issues downstream: in the example reproduction code above, `whois` tries to parse a member argument, but in doing so it assumes that `data.GuildData` is set. This ultimately causes the nil pointer dereference on `data.GuildData.GS.ID`:
https://github.com/botlabs-gg/yagpdb/blob/b9dabf9eba7d45c1189b695c55076f3f9669c022/commands/util.go#L235
(Note that all commands that access guild data in any way suffer from the same issue; it is not unique to `whois`.)

The fix is straightforward: ~~simply monkeypatch the input message to have a valid GuildID before `execCmd` sees it~~ patch the context message in the interaction create handler before executing custom commands.

### DIFF
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -399,7 +399,7 @@ func handleDelayedRunCC(evt *schEventsModels.ScheduledEvent, data interface{}) (
 	if err != nil {
 		return false, errors.WrapIf(err, "find_command")
 	}
-	
+
 	if cmd.R.Group != nil && cmd.R.Group.Disabled {
 		return false, errors.New("custom command group is disabled")
 	}
@@ -685,6 +685,10 @@ func handleInteractionCreate(evt *eventsystem.EventData) {
 		return
 	}
 
+	// Ephemeral messages always have guild_id = 0 even if created in a guild channel; see
+	// https://github.com/discord/discord-api-docs/issues/4557. But exec/execAdmin rely
+	// on the guild ID of the message to fill guild data, so patch it here.
+	interaction.Message.GuildID = evt.GS.ID
 	interaction.Member.GuildID = evt.GS.ID
 
 	switch interaction.Type {
@@ -1122,7 +1126,7 @@ func ExecuteCustomCommand(cmd *models.CustomCommand, tmplCtx *templates.Context)
 	// deal with the results
 	if err != nil {
 		logger.WithField("guild", tmplCtx.GS.ID).WithError(err).Error("Error executing custom command")
-		
+
 		errChannel := tmplCtx.CurrentFrame.CS.ID
 		if cmd.RedirectErrorsChannel != 0 {
 			errChannel = cmd.RedirectErrorsChannel


SoggySaussages:
It seems to be Yag convention to make these types of patches in the trigger handlers, see [reaction handler](https://github.com/botlabs-gg/yagpdb/blob/master/customcommands/bot.go#L640). There has already been an attempt to fix this issue in the interaction handler, but some other less experienced programmer without their head screwed on straight (me) seems to have implemented it incorrectly. Please see https://github.com/botlabs-gg/yagpdb/blob/master/customcommands/bot.go#L691.

My proposal for the fix is to add a line immediately proceeding [line 688](https://github.com/botlabs-gg/yagpdb/blob/master/customcommands/bot.go#L688) setting the interaction message guild ID. This way it would happen regardless of if the trigger was is a component or a modal, and regardless of Yag’s ability to retrieve the message in question. 

Since this monkeypatch more closely matches existing convention and it would exist only in the interaction-related contexts, it would be more immediately evident why it was there and be able to exist without a large comment.

jo3-l:
I thought about putting it in the handler but decided against it originally since the problem is really with `exec`/`execAdmin`; the missing guild_id seems to be by design from Discord. But your point about convention (indeed, there are patches for `exec`/`execAdmin`) is convincing.

jo3-l:
Switched to patching the message in the interaction create handler in https://github.com/botlabs-gg/yagpdb/pull/1709/commits/b753abd46a5df49fc65c00e600a6caa58345e672. I left the comment in: I think it's important to have a clear justification for patching any data Discord sends us.

PR #1723, opened 2024-09-13
cc/interactions: basic component functions

SoggySaussages:
Not sure if this would be useful, since adding more functions may just cause more confusion, but I have created simplified versions of button menu and modal creation. Offering significantly less customization they are very quick and much more intuitive to use. They do break convention a little and rather than functioning like `cembed` they fall back on the simple pre-defined order of arguments to create an object. However, it reduces arguments required to 2-3 and may offer a friendlier front to jump into interactions.

| **Function**                          | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                        |
|---------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `bbuton` label customID (buttonStyle) | Creates a simple button to be sent in a message or response. `label` is what text the button will display. The string you place in `customID` allows you to determine which Component Trigger CCs will be triggered by the button being pressed. Alternatively, provide this field with a URL if you are using the `"link"` buttonStyle. `buttonStyle` is optional. Options are "blue", "green", "red", "grey", "link", or "disabled". |
| `bmenu` customID options              | Creates a simple menu to be sent in a message or response. The string you place in `customID` allows you to determine which Component Trigger CCs will be triggered by the button being pressed. `options` must be a slice of strings with the options you would like the user to pick from.                                                                                                                                           |
| `bmodal` title customID fields        | Creates a simple menu to be sent in a message or response. `title` is the text displayed at the top of the modal. The string you place in `customID` allows you to determine which Component Trigger CCs will be triggered by the button being pressed. `fields` must be a slice of strings with the fields you would like the user to fill out to submit the modal.  


ashishjh-bst:
I think this is an overkill and only adds bloat.

l-zeuch:
I agree with Des here. The current functions are all quite intuitive
already, and they mirror an already existing function (`cembed`).

To be frank, I don't think having positional arguments, as you propose
in these basic functions, over keyword arguments as we have them in
`cembed` and the already existing component builders is "more
intuitive". If anything, I tend to find it difficult to remember which
argument goes where; `execCC` may serve as a rather extreme example of
this issue I am describing.

Secondly, I fear that this change will (more or less significantly)
increase cognitive overhead—now there are multiple ways to achieve a
similar result (creating a message component).

We deliberately point out that custom interactions are an advanced topic
and require a certain level of understanding custom commands as well as
familiarity with them. We introduce how to build embeds quite early in
the learning course, it is by no means a far stretch to have users apply
this knowledge to the current way of building message components.

PR #1727, opened 2024-09-21
Add Guide to host with an external Nginx Proxy Manager in Docker

egelhaus:
Hello,
I wanted to contribute to YAGPDB.
My Contributation is, that I added an Guide, on how to install YAGPDB on an Docker Server, with an external NGM on an different Server.


SoggySaussages:
This seems to me like a very specific use case which would be best kept in its own community-hosted guide. It doesn't make a lot of sense to me that we would include and maintain this extensive documentation primarily for use of a completely separate interface. You might consider raising this PR against https://hostyagpdb.xyz.

egelhaus:
@SoggySaussages I will try to do that.

PR #1748, opened 2024-10-30
Replaced the terms blacklist/whitelist with more appropriate terms

ashishjh-bst:
The terms though don't have racist origins, I can see how they can be inappropriate and get lost in context.


Happypig375:
Might as well rename `master` to `main` while we're there

PR #1749, opened 2024-11-01
Make integer divide by zero a catchable error

Happypig375:
Runtime errors bypass {{catch}} and are generally unexpected.


lemm-e:
Open PRs to `dev` instead of `master`.

Happypig375:
Done, no idea why the test is failing with action-at-a-distance though.

PR #1753, opened 2024-11-10
module: fix go version

SoggySaussages:
This solves an issue with go build attempting to automatically update go to a version that doesn’t technically exist.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/botlabs-gg/yagpdb/v2
 
-go 1.23
+go 1.23.0
 
 require (
 	emperror.dev/errors v0.8.1
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/botlabs-gg/yagpdb/v2
 
-go 1.23.0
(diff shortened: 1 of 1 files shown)


ashishjh-bst:
I have never had this issue, can you add steps for me to reproduce this issues please?

SoggySaussages:
I encountered on Ubuntu Server 20.04 whilst running standalone, if I try to build without the `.0` appended it doesn’t work. Even after running with the `.0` appended and then subsequently removing it again, same issue.

```
***@***.***:~/sgpdb/cmd/sgpdb$ sh build.sh
Building version SGPDBv1.6.0-44-g6ecc4c5b
go: downloading go1.23 (linux/amd64)
go: download go1.23 for linux/amd64: toolchain not available
```

galen8183:
Apparently excluding the minor version indicates a development version only: https://github.com/golang/go/issues/62278#issuecomment-1693538776

ashishjh-bst:
in that case it should point to the latest version, which at this point is 1.23.3

SoggySaussages:
Donezo

PR #1755, opened 2024-11-11
AMv2 "Send Alert" effect/functionality

KTStephano:
The purpose of this is to add an AMv2 effect type that is similar to the "Send alert" option that Discord's builtin AutoMod has. It allows for mod teams to log AMv2 events (with an optional custom message) that may require additional review before an action is taken rather than relying on automatic actions alone.

The reason a custom message was included is to allow mod teams to include extra context or role pings to go along with the alert.

## Examples

Here is an example of what it looks like in the UI:

Screenshot 2024-11-10 at 8 58 37 PM
Screenshot 2024-11-10 at 8 58 50 PM

Here is an example of what it looks like when a rule is triggered (custom message left blank):

Screenshot 2024-11-10 at 9 09 14 PM

### DIFF
--- a/automod/effects.go
+++ b/automod/effects.go
@@ -2,6 +2,7 @@ package automod
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


galen8183:
This is not much different from the existing Send Message effect, why not add a checkbox like "Include trigger message info" to it rather than a whole new effect?

ashishjh-bst:
> This is not much different from the existing Send Message effect, why not add a checkbox like "Include trigger message info" to it rather than a whole new effect?

The key difference to me looks to be that this is more for the mods and the sendMessage effect is more for the user who triggered the automod. IIRC you cannot have two of the same effect in a rule.

SoggySaussages:
> > This is not much different from the existing Send Message effect, why not add a checkbox like "Include trigger message info" to it rather than a whole new effect?
> 
> The key difference to me looks to be that this is more for the mods and the sendMessage effect is more for the user who triggered the automod. IIRC you cannot have two of the same effect in a rule.

I agree, and I feel like if included in the regular send message effect rather than separating it to an entirely separate rule, people would start asking for more customization. Keeping it as a separate rule enforces that it is a specific purpose built effect for mod review.

That said,

galen8183:
> IIRC you cannot have two of the same effect in a rule.
You can, they are deliberately de-duplicated currently: https://github.com/botlabs-gg/yagpdb/blob/85ca471658f78ccbebd3364611b8e2a39cd943cb/automod/effects.go#L873-L875

galen8183:
> people would start asking for more customization

I'm not sure I agree, if it's only a checkbox whether to include log info. Just my opinion though :)

> the sendMessage effect is more for the user who triggered the automod

Fair, though I'm not sure when you'd want a message to the user sent to another channel. That's mainly what makes these seem redundant to me.

KTStephano:
> The key difference to me looks to be that this is more for the mods

Yeah this is how I imagined it would be used for most cases. I wasn't sure what to do with the autodelete and ping user checkboxes in `sendMessage` so I added it as a separate one.

I'm also still not sure what the default behavior should be for when channel == nil if we keep it separate. Should leaving channel nil effectively disable the mod alert?

SoggySaussages:
Meh just let it send to current channel, could make it useful for testing your triggers and conditions

PR #1763, opened 2024-11-15
templates: reorganize context func declarations

SoggySaussages:
A completely visual reorganization of the context func declaration and removal of exact duplicates which is imo much 
needed, but may cause some merge conflicts downstream, hence separating into a different PR instead of lumping into 
something else.


ashishjh-bst:
I am not sure why is this needed at all.

SoggySaussages:
Purely because there is evidently some attempt at organization here that isn't standardized and leads to confusion when
trying to read or decide where to fit in new functions. It's totally just something that bothers me that I'd love to see fixed
for easier readability down the line, I'm not married to it. If you want to close I can just remove the exact duplicates along 
with a different PR, just thought that if I was removing duplicates I might as well do the whole thing if it was wanted by more
than just me.

PR #1764, opened 2024-11-15
templates: add message forwarding

SoggySaussages:
Adds message forwarding capability to complexMessage. Cannot be done in channels YAGPDB doesn't have read message
perms in, or a channel outside the guild. Effectively no different from the message link preview CC but utilizes Discord's UI,
additionally removes the need to parse a message link and get the message, nor build your own embed or other means of
providing a message timestamp, link, and indication that it was a forwarded message. It also looks neat.


ashishjh-bst:
Can you add example template too please?

common/templates/general.go
```diff
@@ -431,8 +431,41 @@ func CreateMessageSend(values ...interface{}) (*discordgo.MessageSend, error) {
...
+				case "message":
+					msg.Reference.MessageID = ToInt64(v)
+					if msg.Reference.MessageID <= 0 {
+						return nil, errors.New(fmt.Sprintf("invalid message id '%s' provided to forward.", ToString(val)))
+					}
+				}
+			}
 		default:
-			return nil, errors.New(`invalid key "` + key + `" passed to send message builder`)
+			return nil, errors.New(`invalid key "` + key + `" passed to forward`)
```
galen8183, line 468: This error is for `complexMessage`, 'send message builder' is correct.

galen8183:
> Can you add example template too please?

Here's an example:
```go
{{ sendMessage nil (complexMessage
    "forward" (sdict
        "channel" .Channel.ID
        "message" <messageID>
    )
) }}
```
image

common/templates/general.go
SoggySaussages: Thanks for the catch, fixed.

SoggySaussages:
> > Can you add example template too please?
> 
> Here's an example:
> 
> ```go
> {{ sendMessage nil (complexMessage
>     "forward" (sdict
>         "channel" .Channel.ID
>         "message" <messageID>
>     )
> ) }}
> ```
> 
> image

I'm SoggySaussages and I approve this message.

PR #1765, opened 2024-11-15
cmddocsgen: note un/nick hist cmds disabled

SoggySaussages:
When generating command docs with username logging config disabled, instead of omitting the relevant commands as is the current behaviour, this PR adds a note mentioning the relevant  commands are disabled on the public instance. This 
allows for unaffected command docs to be generated including the commands in question on an instance running nickname 
logging. For instances with username logging disabled, the docs generate for the benefit of downstream forks using the 
upstream's docs while informing it's own users it won't be accessible.

### DIFF
--- a/logs/plugin_bot.go
+++ b/logs/plugin_bot.go
@@ -9,6 +9,7 @@ import (
 	"emperror.dev/errors"
 	"github.com/botlabs-gg/yagpdb/v2/bot/paginatedmessages"
 	"github.com/botlabs-gg/yagpdb/v2/common/config"
+	"github.com/botlabs-gg/yagpdb/v2/common/run"
 
 	"github.com/botlabs-gg/yagpdb/v2/bot"
 	"github.com/botlabs-gg/yagpdb/v2/bot/eventsystem"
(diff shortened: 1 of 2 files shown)


logs/plugin_bot.go
```diff
@@ -30,6 +31,12 @@ func (p *Plugin) AddCommands() {
 		commands.AddRootCommands(p, cmdLogs, cmdWhois, cmdNicknames, cmdUsernames, cmdClearNames)
 	} else {
 		commands.AddRootCommands(p, cmdLogs, cmdWhois)
+		if run.FlagGenCmdDocs {
+			for _, cmd := range []*commands.YAGCommand{cmdNicknames, cmdUsernames, cmdClearNames} {
+				cmd.Description = fmt.Sprint(cmd.Description, " Disabled on the public instance.")
```
l-zeuch, line 36: Perhaps "disabled on this instance" instead? Though I'm not sure if the average self-hoster bothers to host their own docs, let alone generate command docs, so maybe not.

SoggySaussages: That's my thought as well, if you're using gencmddocs you can always mod the source code.

PR #1773, opened 2024-11-23
Adding Keys to Dict and SDict

ColinLDev:
Des said wan be added (and I want that)
"maybe that can be implemented on dicts and sdicts too"


ashishjh-bst:
This isn't what I said, I said, maybe .Keys can be added but it also returns an iterable which still needs ranging over it. 
The simpler thing to do is just range over it in the CC itself instead of doing this junky behaviour.

PR #1776, opened 2024-11-27
util: add strict msg type checking

SoggySaussages:
This PR changes the IsNormalUserMessage func to ignore messages which are not Default, Reply, or Thread Opening types 
by default. It adds an arg allowing the function to ignore this new check, which I have set only to be used by CCs and the
eventsystem.

This means this currently impacts both versions of automod, commands, and reputation thanks detection.

### DIFF
--- a/bot/util.go
+++ b/bot/util.go
@@ -447,11 +447,18 @@ func CheckDiscordErrRetry(err error) bool {
 	return true
 }
 
-func IsNormalUserMessage(msg *discordgo.Message) bool {
+var allowedMsgTypes = []int{int(discordgo.MessageTypeDefault), int(discordgo.MessageTypeReply), int(discordgo.MessageTypeThreadStarterMessage)}
+
+// set strictType to true to only permit default, reply, and thread starter message types
+func IsNormalUserMessage(msg *discordgo.Message, ignoreMsgType bool) bool {
 	if msg.Author == nil || msg.Author.ID == common.BotUser.ID || msg.WebhookID != 0 || msg.Author.Discriminator == "0000" || (msg.Member == nil && msg.GuildID != 0) {
 		// message edits can have a nil author, those are embed edits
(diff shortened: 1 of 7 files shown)


bot/util.go
```diff
@@ -447,13 +447,20 @@ func CheckDiscordErrRetry(err error) bool {
 	return true
 }
 
-func IsNormalUserMessage(msg *discordgo.Message) bool {
+var allowedMsgTypes = []int{int(discordgo.MessageTypeDefault), int(discordgo.MessageTypeReply), int(discordgo.MessageTypeThreadStarterMessage)}
+
+// set strictType to true to only permit default, reply, and thread starter message types
+func IsNormalUserMessage(msg *discordgo.Message, ignoreMsgType bool) bool {
```
jo3-l, line 453: What do you think about separating into two functions, `IsUserMessage` for the current check and `IsNormalUserMessage` for the behavior with `ignoreMsgType` set to `true`? The two are meaningfully different and combining both into one function with a flag is a little confusing. Distinct names is also clearer from the call site (instead of opaque true/false values.)

bot/util.go
```diff
@@ -447,13 +447,20 @@ func CheckDiscordErrRetry(err error) bool {
 	return true
 }
 
-func IsNormalUserMessage(msg *discordgo.Message) bool {
+var allowedMsgTypes = []int{int(discordgo.MessageTypeDefault), int(discordgo.MessageTypeReply), int(discordgo.MessageTypeThreadStarterMessage)}
```
jo3-l, line 450: What do you think about inlining the check and instead writing:
```go
switch msg.Type {
case discordgo.MessageTypeDefault, discordgo.MessageTypeReply, discordgo.MessageTypeThreadStarterMessage:
	return true
default:
	return false
}
```
This avoids the cast, is a little shorter, and is more performant (though that isn't the main reasoning for this suggestion, to be clear.) We also use this pattern in a couple other places.

SoggySaussages: Oo I like it, didn’t cross my mind.

PR #1780, opened 2024-12-01
Support attached images when issuing warnings to user

Sasiko:
The purpose is to allow moderators to add an attached image when issuing a warning to a user. Usecase would be a screenshot of the offense like DM advertisement

/warnings will show attached image as part of the warning "reason".

bild

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -886,13 +886,23 @@ var ModerationCommands = []*commands.YAGCommand{
 			if parsed.TraditionalTriggerData != nil {
 				msg = parsed.TraditionalTriggerData.Message
 			}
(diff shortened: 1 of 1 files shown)


ashishjh-bst:
First, you must read contributing guidelines, and know how to raise a PR. 
Second, the attachment link would become invalid as soon as the original moderation message is deleted, so this just extends the functionality to store dud urls.

ashishjh-bst:
Also, this doesn't compile, please build your code locally before sending in PRs. 
```
# github.com/botlabs-gg/yagpdb/v2/moderation
../../moderation/commands.go:900:114: not enough arguments in call to WarnUser
        have (*Config, int64, *dstate.ChannelState, *discordgo.Message, *discordgo.User, *discordgo.User, string)
        want (*Config, int64, *dstate.ChannelState, *discordgo.Message, *discordgo.User, *discordgo.User, string, bool)
```

PR #1788, opened 2024-12-08
tickets: add optional buttons to interface

SoggySaussages:
Adds optional buttons which server admins can use on their server for opening and closing tickets.

- Add a new command to create a ticket menu
    - By default, sends a standard message with a button allowing users to fill out a modal for their reason. Then
opens the ticket with that reason.
    - With the `-message` switch, can instead append the buttons to another message sent by the bot.
    - Can preset up to 9 reasons in separate buttons with switches. These buttons open with the set reason
instead of prompting the user via modal.
    - If preset reasons are used, can set the `-disable-custom` switch to disable the custom reason button, only
allowing users to open with the preset reasons.

- Add new ticket config to append "Close" and/or "Close with Reason" buttons to a ticket opener message.
    - Close with reason sends a modal to prompt the user to specify the reason
    - Scalable to allow adding more in-ticket buttons in the future (add/remove user, join/leave ticket, etc.)

Screenshot 2024-12-08 at 00 45 18

Screenshot 2024-12-08 at 00 38 57

Screenshot 2024-12-08 at 00 39 22

Screenshot 2024-12-08 at 00 45 47

Screenshot 2024-12-08 at 00 46 28

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -495,7 +495,13 @@ func (c *Context) SendResponse(content string) (m *discordgo.Message, err error)
 	embeds = append(embeds, c.CurrentFrame.EmbedsToSend...)
 	msgSend.Embeds = embeds
 	msgSend.Content = content
-	if (len(msgSend.Embeds) == 0 && strings.TrimSpace(content) == "") || (c.CurrentFrame.DelResponse && c.CurrentFrame.DelResponseDelay < 1) {
+	if len(c.CurrentFrame.ComponentsToSend) > 0 {
+		msgSend.Components = append(msgSend.Components, c.CurrentFrame.ComponentsToSend...)
+		if len(msgSend.Components) > 5 {
+			msgSend.Components = msgSend.Components[:5]
+		}
+	}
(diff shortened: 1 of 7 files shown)


l-zeuch:
This seems already possible with custom commands and the ticket message
template, or am I missing something? Perhaps you could describle what
this solves that you cannot achieve with custom commands.

At the time, this just seems like a few more knobs and dials to turn on
an already bloated control panel. Your proposed mechanism of creating a
menu seems... not less complicated than just writing a custom command,
if I'm honest.

l-zeuch:
> We want more users to use features natively instead of pushing them
>through the trial of custom commands. This would be a good quality of
> impovement as a built-in feature.

Ah, that makes sense. Thanks for the clarification.

SoggySaussages:
> At the time, this just seems like a few more knobs and dials to turn on an already bloated control panel. Your proposed mechanism of creating a menu seems... not less complicated than just writing a custom command, if I'm honest.

This method is similar to the poll command, i’d argue having one 
command you don’t need to enable or configure on the dashboard
at all (presuming you already have tickets setup) is much easier
than even rolemenus.

PR #1789, opened 2024-12-08
refact(reddit): Clean up subreddit name processing

Wolveric:
Users prefixing their input with `r/` when adding Reddit feeds,
has been a persistent issue, enough to the point it had a dedicated note
on the config. page in the control panel.

This PR initially intended to address exclusively this issue,
by adding behaviour to strip leading instances of `/r/`
from inputs provided when adding subreddit feeds,
but also discovered the functions handling modifying
and removing these feeds, seemed to be performing redundant work
processing the subreddit names, when this work was already performed
when creating the feed.

Consequently, this PR also removes these redundant function calls,
cleaning up the handling of subreddit names in general.


Wolveric:
Damn, too many commits... closing...

PR #1790, opened 2024-12-08
refact(reddit): Clean up subreddit name processing

Wolveric:
Users prefixing their input with r/ when adding Reddit feeds,
has been a persistent issue, enough to the point it had a dedicated note
on the config. page in the control panel.

This PR initially intended to address exclusively this issue,
by adding behaviour to strip leading instances of /r/
from inputs provided when adding subreddit feeds,
but also discovered the functions handling modifying
and removing these feeds, seemed to be performing redundant work
processing the subreddit names, when this work was already performed
when creating the feed.

Consequently, this PR also removes these redundant function calls,
cleaning up the handling of subreddit names in general.

### DIFF
--- a/reddit/plugin_web.go
+++ b/reddit/plugin_web.go
@@ -261,7 +261,7 @@ func HandleRemove(w http.ResponseWriter, r *http.Request) interface{} {
 
 	go cplogs.RetryAddEntry(web.NewLogEntryFromContext(r.Context(), panelLogKeyRemovedFeed, &cplogs.Param{Type: cplogs.ParamTypeString, Value: item.Subreddit}))
 	go pubsub.Publish("reddit_clear_subreddit_cache", -1, PubSubSubredditEventData{
-		Subreddit: strings.ToLower(strings.TrimSpace(item.Subreddit)),
+		Subreddit: item.Subreddit,
 		Slow:      item.Slow,
 	})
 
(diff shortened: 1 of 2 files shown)


reddit/plugin_web.go
```diff
@@ -136,10 +136,18 @@ func HandleNew(w http.ResponseWriter, r *http.Request) interface{} {
 		return templateData.AddAlerts(web.ErrorAlert(fmt.Sprintf("Max %d feeds allowed (or %d for premium servers)", GuildMaxFeedsNormal, GuildMaxFeedsPremium)))
 	}
 
+	subreddit := strings.TrimSpace(newElem.Subreddit)
+	subreddit = strings.ToLower(subreddit)
+	subreddit = strings.TrimPrefix(subreddit, "/")
+
+	for ; strings.HasPrefix(subreddit, "r/"); {
```
SoggySaussages, line 143: How about splitting at `/` and only taking the last string?

jo3-l: To be honest, I think just `subreddit = strings.TrimPrefix(subreddit, "r/")` is enough--no loop. I don't see a real reason to support `r/r/foo`, which is clearly a typo, whereas accepting `r/foo` is a clear improvement.

ashishjh-bst: > To be honest, I think just `subreddit = strings.TrimPrefix(subreddit, "r/")` is enough--no loop. I don't see a real reason to support `r/r/foo`, which is clearly a typo, whereas accepting `r/foo` is a clear improvement.

I agree with this.

Wolveric: Yeah, I didn't love the idea either, so I'll drop the `for` loop from the behaviour.

PR #1792, opened 2024-12-12
Add `suppressEmbeds` and `unsuppressEmbeds` functions for CC

Aresiel:
These functions enable the suppressing of embeds on messages not sent by YAGPDB via the [SUPPRESS_EMBEDS](https://discord.com/developers/docs/resources/message#message-object-message-flags) flag.

The implementation should be functional if not for the bug discussed in https://github.com/bwmarrin/discordgo/issues/1484 which is patched in https://github.com/bwmarrin/discordgo/pull/1483/. Discussion in the Discord was split on whether it is best to implement that solution in YAGPDB's discordgo library or to create an entirely new restapi function `ChannelMessageEditFlags` that handles specifically editing only flags. Look around [here](https://discord.com/channels/166207328570441728/384011387132706816/1316820207095906426) if context is wanted.

There has also been discussion of whether this should be part of `complexMessageEdit`, but little of a consensus, so it so far not.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -657,6 +657,8 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("publishMessage", c.tmplPublishMessage)
 	c.addContextFunc("publishResponse", c.tmplPublishResponse)
 	c.addContextFunc("unpinMessage", c.tmplPinMessage(true))
+	c.addContextFunc("suppressEmbeds", c.tmplSuppressEmbeds(true))
+	c.addContextFunc("unsuppressEmbeds", c.tmplSuppressEmbeds(false))
 
 	// Message send functions
 	c.addContextFunc("sendDM", c.tmplSendDM)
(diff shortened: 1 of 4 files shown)


common/templates/context.go
SoggySaussages, line 1: These should be ordered alphabetically.

common/templates/context_funcs.go
```diff
@@ -518,6 +518,43 @@ func (c *Context) tmplPinMessage(unpin bool) func(channel, msgID interface{}) (s
...
+		mID := ToInt64(msgID)
+
+		message, _ := common.BotSession.ChannelMessage(cID, mID)
+		if message == nil {
+			return "", errors.New("unknown message")
+		}
+
+		newFlags := message.Flags | discordgo.MessageFlagsSuppressEmbeds
+		if !suppress {
+			newFlags = newFlags ^ discordgo.MessageFlagsSuppressEmbeds
```
SoggySaussages, line 541: Why XOR here instead of OR?

Aresiel: The intent is for L539 to ensure the presence of the flag. While L541 executes if the function is set to unsuppress to then remove the flag.

PR #1795, opened 2024-12-13
Add unwarn modlog (with option on dashbord)

ColinLeDev:
Starting from https://discord.com/channels/166207328570441728/1316169654502293514
- Added two options (db+dashboard)
		UnwarnSendToModlog: In the name
		UnwarnIncludeWarnReason: Include the dbStored (not the modlog message) original warn reason
- Added a ModlogAction (**You should choose color & emote**)
	MAUnwarned       = ModlogAction{Prefix: "Warn removed", Emoji: "🧽", Color: 0xfca253}


moderation/commands.go
```diff
@@ -1033,6 +1047,24 @@ var ModerationCommands = []*commands.YAGCommand{
...
+			if config.UnwarnSendToModlog && config.ActionChannel != 0 {
+				userid, err := strconv.ParseInt(warning.UserID, 10, 64)
+				if err != nil {
+					return "Failed parsing user ID", err
+				}
+				user := bot.GetUsers(parsed.GuildData.GS.ID, userid)[0]
+
+				message := parsed.Args[1].Str()
+				if config.UnwarnIncludeWarnReason {
+					message = fmt.Sprintf("%s\n~~%s~~", message, warning.Message)
```
SoggySaussages, line 1059: I wonder if this might be more clear if the previous, strikenthrouwgh reason was first, followed by the new one. What are your thoughts?

moderation/commands.go
```diff
@@ -1033,6 +1047,24 @@ var ModerationCommands = []*commands.YAGCommand{
...
 			}
 
+			if config.UnwarnSendToModlog && config.ActionChannel != 0 {
+				userid, err := strconv.ParseInt(warning.UserID, 10, 64)
+				if err != nil {
+					return "Failed parsing user ID", err
+				}
+				user := bot.GetUsers(parsed.GuildData.GS.ID, userid)[0]
+
+				message := parsed.Args[1].Str()
```
SoggySaussages, line 1057: Since we are now actually using the reason, should it be parsed initially with MBaseCmdSecond on line 1018? This way we’d get the trademark (no reason specified) when not set. 

While we’re at it, should there be a config to make delete warning reason optional?

moderation/commands.go
```diff
@@ -1004,7 +1004,7 @@ var ModerationCommands = []*commands.YAGCommand{
 		RequiredArgs:  1,
 		Arguments: []*dcmd.ArgDef{
 			{Name: "Id", Type: dcmd.Int},
-			{Name: "Reason", Type: dcmd.String},
+			{Name: "Reason", Type: dcmd.String, Default: ""},
```
SoggySaussages, line 1007: I’m curious what this accomplishes.

moderation/assets/moderation.html
```diff
@@ -472,6 +472,8 @@ <h4>Delete ALL server warnings?</h4>
 
         {{checkbox "WarnIncludeChannelLogs" "WarnIncludeChannelLogs" "Create message logs in the channel that the command was run in when a user is warned" .ModConfig.WarnIncludeChannelLogs}}
         {{checkbox "WarnSendToModlog" "WarnSendToModlog" "Send warnings to the modlog" .ModConfig.WarnSendToModlog}}
+        {{checkbox "UnwarnSendToModlog" "UnwarnSendToModlog" "Send warning removal to the modlog" .ModConfig.UnwarnSendToModlog}}
+        {{checkbox "UnwarnIncludeWarnReason" "UnwarnIncludeWarnReason" "Include warning reason in the modlog" .ModConfig.UnwarnIncludeWarnReason}}
```
SoggySaussages, line 476: I’m not sure it’s immediately evident this is still referring to a warn removal.

moderation/assets/moderation.html
```diff
@@ -472,6 +472,8 @@ <h4>Delete ALL server warnings?</h4>
 
         {{checkbox "WarnIncludeChannelLogs" "WarnIncludeChannelLogs" "Create message logs in the channel that the command was run in when a user is warned" .ModConfig.WarnIncludeChannelLogs}}
         {{checkbox "WarnSendToModlog" "WarnSendToModlog" "Send warnings to the modlog" .ModConfig.WarnSendToModlog}}
+        {{checkbox "UnwarnSendToModlog" "UnwarnSendToModlog" "Send warning removal to the modlog" .ModConfig.UnwarnSendToModlog}}
```
SoggySaussages, line 475: Have you considered a horizontal rule here?

moderation/schema.go
```diff
@@ -51,6 +51,8 @@ CREATE TABLE IF NOT EXISTS moderation_configs (
 	warn_include_channel_logs BOOLEAN,
 	warn_send_to_modlog BOOLEAN,
 	warn_message TEXT,
+	unwarn_send_to_modlog BOOLEAN,
```
SoggySaussages, line 54: From what I understand, you will instead need to add these as alterations to the table after defining the table
in the schema. If you add them in here, on existing hosts, it will simply note that the `moderation_configs` table already exists and move on. If you instead insert before line 81 
```txt
`ALTER TABLE moderation_configs ADD COLUMN IF NOT EXISTS unwarn_send_to_modlog BOOLEAN NOT NULL DEFAULT true;`
```
then it will be able to set the new column into the existing table while also assigning a default value to every existing row (every 
guild’s existing configuration). This is what I was referring to [here](https://discord.com/channels/166207328570441728/578976767058903057/1316992372084834449).

moderation/commands.go
ColinLeDev: I tested that way, but having the warn comment first is weird, because the action is the unwarn

moderation/assets/moderation.html
ColinLeDev: Mhh, the command being called deletewarning, it doesn't make sense at all

I'll change that

moderation/assets/moderation.html
ColinLeDev: Nop, but that's a good idea

ColinLeDev: After reviewing the page, I don't think that's a good idea, there is already two columns on the page

moderation/assets/moderation.html
ColinLeDev: Any idea on the text to show ?

moderation/assets/moderation.html
SoggySaussages: A horizontal rule would just separate vertically, you can see there’s one just a few lines below this. It’s just a line with a little bit of spacing that goes across the column, not creating a new column.

ColinLeDev: I misread, yeah I'll add that

moderation/assets/moderation.html
ColinLeDev: image
for now

ColinLeDev: And in back, all renamed to Delwarnxxxxx to match the command name

SoggySaussages: “Append original warn reason to removal” :maby:

moderation/config.go
```diff
@@ -218,6 +222,8 @@ func configFromModel(model *models.ModerationConfig) *Config {
 		WarnCmdRoles:           model.WarnCmdRoles,
 		WarnIncludeChannelLogs: model.WarnIncludeChannelLogs.Bool,
 		WarnSendToModlog:       model.WarnSendToModlog.Bool,
+		DelwarnSendToModlog:     model.DelwarnSendToModlog.Bool,
+		DelwarnIncludeWarnReason: model.DelwarnIncludeWarnReason.Bool,
```
jo3-l, line 226: Did you rerun sqlboiler after changing the schema per Soggy's suggestion? Now that these columns are `NOT NULL`, you should not need the `.Bool` anymore; sqlboiler should just generate normal booleans.

moderation/commands.go
```diff
@@ -1015,12 +1015,27 @@ var ModerationCommands = []*commands.YAGCommand{
...
+					// don't get warning from other servers, even if ID is correct
+					models.ModerationWarningWhere.GuildID.EQ(parsed.GuildData.GS.ID),
+				).OneG(parsed.Context())
+				if err != nil {
+					return fmt.Sprintf("Could not find warning with ID `%d`", warningID), nil
+				}
+			}
+
 			numDeleted, err := models.ModerationWarnings(
 				models.ModerationWarningWhere.ID.EQ(warningID),
```
jo3-l, line 1040: I think fetching `warning` unconditionally and calling `warning.Delete` is more straightforward here. You can remove the separate deletion query below if you do this.

moderation/schema.go
```diff
@@ -75,6 +75,9 @@ ALTER TABLE moderation_configs ALTER COLUMN created_at SET NOT NULL;
 `, `
 ALTER TABLE moderation_configs ALTER COLUMN updated_at SET NOT NULL;
 `, `
+ALTER TABLE moderation_configs ADD COLUMN IF NOT EXISTS delwarn_send_to_modlog BOOLEAN NOT NULL DEFAULT false;
+ALTER TABLE moderation_configs ADD COLUMN IF NOT EXISTS delwarn_include_warn_reason BOOLEAN NOT NULL DEFAULT false;
```
SoggySaussages, line 79: Missing a
```txt
`, `
```

moderation/commands.go
SoggySaussages: What were your thoughts about adding config to make warning reason optional?

moderation/assets/moderation.html
ColinLeDev: I don't like replacing "removing" by "deletion" but it matches the function....

SoggySaussages: yep that makes sense to me

.gitignore
SoggySaussages, line 1: Was this intentional?

ColinLeDev: Not at all

ashishjh-bst:
It is isn't still clear to me what this PR actually does and what is the purpose of it. Please showcase demo usage along with screenshots.

ColinLeDev:
> It is isn't still clear to me what this PR actually does and what is the purpose of it. Please showcase demo usage along with screenshots.

You can see here https://sh.col1n.fr/YAG-PR1795-demo.html
(modlog is the channel)

lemm-e:
I don't understand the purpose of displaying warn reason in the deletewarning entry. We don't do this with mutes or bans, so why do it here?

ColinLeDev:
At first, I wasn't going to add the original warn reason, but I was already getting the warn entry, and it can be useful for admins to quickly have the original reason without looking in the modlogs (which don't display the warnID)

So if you delete a warning, there is no way for the admin to have the original reason
Mutes, ban, kick are discord Audit-logged if I remember correctly

lemm-e:
If you delete a warning, you do so because you don't need it anymore. So why would you wanna have the original reason? Imo consistency amongst all commands should be favored over the little bit of effort it would take to look at the mod-logs to find the original reason if it is ever needed at all

ColinLeDev:
Fair point, I can remove that part (but modlog the delwarn is consistent)

SoggySaussages:
I am against removal. If you’re unmuted, you know there’s only one mute they can be talking about. If you’re unbanned, there could only be one ban you’re talking about. If you delete a warning, there could be multiple warnings and you are only removing one. If the only identifier is a warning ID, it will be literally impossible to find which warning it is referring to without manually counting every warning ever issued on your server.

PR #1797, opened 2024-12-17
templates: be stricter with accepted role inputs

jo3-l:
In https://github.com/botlabs-gg/yagpdb/pull/883, we consolidated the role functions so that they all call `FindRole` under the hood, which accepts role IDs, role mentions, role names, and role objects. However, this change introduced a minor regression--it eagerly resolves inputs in the order above--so a string such as `"1"` will always be interpreted as a role ID instead of a role name (and thus never resolve to the role named `1`, if it exists.) This scenario is admittedly degenerate, but we have gotten the occasional complaint about broken code from servers using numerical role names with `addRoleName` and similar.

In this PR, we try to fix the above issue by allowing to restrict the types of input `FindRole` accepts and adjusting the role functions accordingly. So, for example, `addRole` still accepts all types of input--IDs, mentions, role names, role objects--but `addRoleID` only accepts role IDs and `addRoleName` only accepts role names.

This following custom command illustrates the difference on a range of inputs.
<details>
<summary>Show code</summary>

```
{{$ro := .uninitialized}}
{{range .Guild.Roles}}
    {{- if eq .Name "myrole"}}
        {{- $ro = .}}
    {{- end -}}
{{end}}

{{$inputs := cslice
    (sdict
             "Label" "string ID"
             "Value" "1318709349304111145"
    )
    (sdict
        "Label" "numerical role name"
        "Value" "1"
    )
    (sdict
        "Label" "role mention"
        "Value" "1318709349304111145"
    )
    (sdict
        "Label" "string @everyone"
        "Value" "@everyone"
    )
    (sdict
        "Label" "role object"
        "Value" $ro
    )
    (sdict
        "Label" "integer ID"
        "Value" 1318709349304111145
    )
}}

{{range $inputs}}
- {{.Label}}: `{{printf "%T(%#[1]v)" .Value}}`
  - getRole: {{if $v := getRole .Value}}resolves to {{$v.Mention}}{{else}}not found{{end}}
  - getRoleID: {{if $v := getRoleID .Value}}resolves to {{$v.Mention}}{{else}}not found{{end}}
  - getRoleName: {{if ne (kindOf .Value) "string"}}unsupported{{else if $v := getRoleName .Value}}resolves to {{$v.Mention}}{{else}}not found{{end }}
{{ end }}
```
</details>

<details>
<summary>Show before/after comparison</summary>

Before:
before

After:
after

</details>

It's not clear whether we want to make this change for such an edge case. In particular, the role functions always silently fail when they cannot resolve the role, and it's possible this change will introduce issues for users who have gotten used to code such as `addRoleName <roleID>` working. (Before this change, `addRoleName` behaved exactly like `addRole`, so accepted role IDs, role objects, and so on as well.) But I figured I should put the change up to see if there is any interest regardless.

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -2118,70 +2118,93 @@ func comparatorOf(v reflect.Value) (comparator, error) {
 	}
 }
 
-// c.FindRole accepts all possible role inputs (names, IDs and mentions)
-// and tries to find them on the current context
-func (c *Context) FindRole(role interface{}) *discordgo.Role {
+type roleInputType int
+
+const (
+	acceptRoleID roleInputType = 1 << iota
+	acceptRoleMention
+	acceptRoleName
+	acceptRoleObject
+	acceptAllRoleInput = acceptRoleID | acceptRoleMention | acceptRoleName | acceptRoleObject
+)
+
+// FindRole tries to resolve the argument to a role. `accept` specifies the set of allowed input
+// types, tested in the following order: ID, role mention, role name, role object.
+func (c *Context) FindRole(role interface{}, accept roleInputType) *discordgo.Role {
 	switch t := role.(type) {
 	case string:
-		parsed, err := strconv.ParseInt(t, 10, 64)
-		if err == nil {
-			return c.GS.GetRole(parsed)
-		}
-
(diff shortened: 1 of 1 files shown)


jo3-l (commented):
The diff here for `FindRole` looks worse than it actually is. Reviewing in split mode with ignore whitespace enabled will be more helpful:
GitHub review options

PR #1798, opened 2024-12-18
moderation: revert removal of report ephemeral

SoggySaussages:
Report should always be ephemeral for privacy of the reporter. 

### DIFF
--- a/moderation/commands.go
+++ b/moderation/commands.go
@@ -583,7 +583,7 @@ var ModerationCommands = []*commands.YAGCommand{
 		},
 		SlashCommandEnabled: true,
 		DefaultEnabled:      false,
-		IsResponseEphemeral: false,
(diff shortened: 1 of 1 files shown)


Wolveric:
Ok, I wouldn't say I disagree with this stance, but I disabled it by default in the first place because there are some people who don't, and getting messages about it bugged me... it also assumes servers keep their reports channels private.

I guess this would be a valid case for saying, *"Self-host to edit that"* though...

SoggySaussages:
The report command is the only mod command with a configurable channel to send reports that you can
set to a different channel than the mod log. Even if it wasn't, almost every server I've been on doesn't keep
a public mod log. The report command is designed by nature to have privacy, I don't think the people who
would be upset that its not configurable outweigh the people who would be upset that they just accidentally
outed themselves while trying to report someone.

PR #1802, opened 2024-12-24
reputation: implement basic rep roles system

jo3-l:
We implement a basic reputation role system, similar to the `reproles` snippet on the support server but built-in. Normal servers can create up to 5 rep roles (25 for premium.)

web ui


reputation/reputation.go
```diff
@@ -201,16 +210,114 @@ func ModifyRep(ctx context.Context, conf *models.ReputationConfig, guildID int64
...
+)
+
+func GuildMaxRepRoles(guildID int64) int {
+	if isPrem, _ := premium.IsGuildPremium(guildID); isPrem {
+		return MaxRepRolesPremium
+	}
+	return MaxRepRoles
+}
+
+func UpdateRepRoles(gs *dstate.GuildSet, ms *dstate.MemberState, rep int64) error {
```
jo3-l, line 225: This code is a bit nastier than one might imagine because it needs to handle the case where the target member is higher in the role hierarchy than the bot, in which case we cannot bulk edit the member's roles. I don't think there's a way to avoid this case; such a setup is not uncommon: we would, for instance, run into this case if we tried to set up rep roles on the support server.

reputation/reputation.go
```diff
@@ -201,16 +210,114 @@ func ModifyRep(ctx context.Context, conf *models.ReputationConfig, guildID int64
...
+	for _, r := range add {
+		if err = common.BotSession.GuildMemberRoleAdd(gs.ID, ms.User.ID, r); err != nil {
+			return err
+		}
+	}
+	for _, r := range remove {
+		if err := common.BotSession.GuildMemberRoleRemove(gs.ID, ms.User.ID, r); err != nil {
+			return err
+		}
+	}
```
jo3-l, line 262: I don't really like the loop here, but as above I don't know if there's a way to avoid it. There is some potential for abuse here, I suppose, but I think it is balanced by the fact there's only one role to add/remove in the common case. (The worst case happens when a premium server has set up the full 25 rep roles, a member above the bot has all the rep roles, and someone resets their reputation in which case we will issue 25 API requests.) One can already write dubious loops like this using custom commands as well.

jo3-l:
The last commit makes the unsaved change popup work for form elements in a table, which incidentally fixes the same issue for a couple other pages on the control panel. The following screen recording illustrates the issue before this commit on the YouTube page, which is structured similarly to the new rep roles configuration section:

https://github.com/user-attachments/assets/01ed66be-f921-4078-930a-6d70f371b59e

Observe that the unsaved change popup triggers when one edits the main settings, but does not display for changes made to configuration items for individual feeds in the table. The issue was that the `form` element is separate from the configuration items in the table (which specified the form they were linked to using the `form=` attribute), but the frontend JS did not listen for changes to elements specifying a `form=` attribute.

frontend/static/js/spongebob.js
```diff
@@ -614,6 +614,10 @@ function trackForm(form) {
 		console.log("Form changed!");
 		checkForUnsavedChanges();
 	})
+	$(`[form="${form.id}"]`).change(function() {
```
ashishjh-bst, line 617: This should be in `yagInitUnsavedForms` in spongebob.js, all form tracking starts from there.

ashishjh-bst (changes requested):

frontend/static/js/spongebob.js
jo3-l: I'm happy to move it there if you feel it's more appropriate, but I placed this code in `trackForm` (which is called by `yagInitUnsavedForms` anyway) since the main listener is attached there as well. The full context is
```js
// in trackForm...

	$(form).change(function () {
		console.log("Form changed!");
		checkForUnsavedChanges();
	}) // code before this change
	$(`[form="${form.id}"]`).change(function() {
		console.log("Form changed!");
		checkForUnsavedChanges()
	}) // ADDED
```

ashishjh-bst: I mean you could just append the element with that selector to the list of forms in `yagInitUnsavedForms` and not have redundant change handler for it again here.

jo3-l: I am not sure at a glance how we can make that work easily. `trackForm` needs to know what the main form element is, so we cannot treat `form`s and elements with `[form="id"]` the same way.

ashishjh-bst: I was thinking something like this, additionally your current logic doesn't  ignores tracking changes on the element with [form="id"] if it doesn't have the `no-unsaved-popup` class. 
```diff
 function yagInitUnsavedForms(selectorPrefix) {
        let unsavedForms = $(selectorPrefix + "form")
        unsavedForms.each(function (i, rawElem) {
-              if (rawElem.classList.contains("no-unsaved-popup")) return;
               trackForm(rawElem);
+             $(`[form="${rawElem.id}"]`).each((i,e) => trackForm(e) )
        });
 }
 
 function trackForm(form) {
+  if (form.classList.contains("no-unsaved-popup")) return;
        let savedVersion = serializeForm($(form));
 
        let hasUnsavedChanges = false
```

ashishjh-bst: oh wait, the "[form="${rawElem.id}"] is not necessarily a form, yeah ignore this.

PR #1804, opened 2024-12-24
draft: premium: provide 3d grace period after slot removed

jo3-l:
Here is a possible implementation of a grace period before which premium features--additional custom commands, feeds, and so on--are disabled after a guild loses its premium slot.

It is intended really only as a possible starting point and has not been tested very thoroughly (though it works in the simple cases I have tried): I'm happy to see parts of this code cherry-picked and included in the final implementation and the original PR abandoned.

One important issue to note with this and similar approaches is that, as is, users can assign a premium slot to a server, then immediately unlink it to get 3 free days of premium while keeping their premium slot available for use on other servers. In other words, this code does not distinguish between users deliberately unlinking their premium slot from a server and a premium slot expiring naturally--it probably should.

### DIFF
--- a/customcommands/bot.go
+++ b/customcommands/bot.go
@@ -683,7 +683,9 @@ func CCActionExecLimit(guildID int64) int {
 	return CCActionExecLimitNormal
 }
 
-func (p *Plugin) OnRemovedPremiumGuild(GuildID int64) error {
+var _ premium.PremiumGuildGracePeriodEndedListener = (*Plugin)(nil)
+
+func (p *Plugin) OnPremiumGuildGracePeriodEnded(GuildID int64) error {
 	commands, err := models.CustomCommands(qm.Where("guild_id = ?", GuildID), qm.Offset(MaxCommands)).AllG(context.Background())
 	if err != nil {
 		return errors.WrapIf(err, "failed getting custom commands")
--- a/customcommands/web.go
+++ b/customcommands/web.go
@@ -103,6 +103,7 @@ func (p *Plugin) InitWeb() {
 	getDBHandler := web.ControllerHandler(handleGetDatabase, "cp_custom_commands_database")
 
 	subMux := goji.SubMux()
+	subMux.Use(premium.GracePeriodEndingMiddleware("custom commands exceeding standard limits will be disabled"))
 	web.CPMux.Handle(pat.New("/customcommands"), subMux)
 	web.CPMux.Handle(pat.New("/customcommands/*"), subMux)
 	web.CPMux.Handle(pat.New("/customcommands/database"), subMux)
--- a/premium/plugin_web.go
+++ b/premium/plugin_web.go
@@ -99,6 +99,22 @@ func PremiumGuildMW(inner http.Handler) http.Handler {
 	return http.HandlerFunc(mw)
 }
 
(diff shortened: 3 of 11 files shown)


SoggySaussages:
https://github.com/botlabs-gg/yagpdb/blob/master/premium/patreonpremiumsource/patreonpremiumsource.go#L137
why not just make this a scheduled event? Every time a new slot is 
added, check what events are scheduled, choose the one that’s 
closest to running and run the delete event immediately, and it will 
automatically migrate empty slots (like the new one) to fill the gap. 
Otherwise after three days it deletes and if there’s available slots they 
fill in.

jo3-l:
Sorry, I don't fully understand your suggested approach. It seems (?) like it might be an addon to the existing patch rather than a replacement--is that correct?

ashishjh-bst:
In poller.go I can simply check if a payment is pending or declined and next charge date was 3 days from now, and send an alert, no need for a complicated event system or a schema change. 
```

			if len(attributes.LastChargeStatus) > 0 && attributes.LastChargeStatus != patreonapi.ChargeStatusPaid && attributes.LastChargeStatus != patreonapi.ChargeStatusPending {
				continue
			}

			// Skip if the next charge date is in the past
			if attributes.NextChargeDate != nil && attributes.NextChargeDate.Before(time.Now()) {
				continue
			}
```

ashishjh-bst:
Or rather this is sufficient, no need for alerting as patreon will alert the user. 
```golang
@@ -188,16 +188,12 @@ func (p *Poller) Poll() {
                                continue
                        }
 
-                       if attributes.PatronStatus != "active_patron" {
+                       if len(attributes.LastChargeStatus) > 0 && attributes.LastChargeStatus != patreonapi.ChargeStatusPaid && attributes.LastChargeStatus != patreonapi.
ChargeStatusPending && attributes.LastChargeStatus != patreonapi.ChargeStatusDeclined {
                                continue
                        }
 
-                       if len(attributes.LastChargeStatus) > 0 && attributes.LastChargeStatus != patreonapi.ChargeStatusPaid && attributes.LastChargeStatus != patreonapi.
ChargeStatusPending {
-                               continue
-                       }
-
-                       // Skip if the next charge date is in the past
-                       if attributes.NextChargeDate != nil && attributes.NextChargeDate.Before(time.Now()) {
+                       // Skip if the next charge date is passed by 3 days
+                       if attributes.NextChargeDate != nil && attributes.NextChargeDate.Before(time.Now().Add(-1*3*24*time.Hour)) {
                                continue
                        }
```

jo3-l:
Sounds reasonable. I will close this draft then.

ashishjh-bst:
The above plan has an issue, the entitled cents amount is 0, and hence the user isn't assigned a slot, and there doesn't seem too be a way to figure out how much the user paid in the last charge. 

I could possible get the users current set of slots and calculate it myself, or mark the slots to be removed after 3 days, and if a current entitled users slot is marked for removal after three days undo it.

jo3-l:
The latter idea is somewhat like what I implemented here (and is, I think, the easiest way to go), so feel free to reuse code from this PR as needed or reopen and modify directly, whichever is easiest. Unfortunately, I will not have time to work on this more for the time being.

PR #1807, opened 2024-12-26
(Bug) Corrected emoji existence check in tickets menucreate

ColinLeDev:


tickets/tickets_commands.go
```diff
@@ -333,7 +333,7 @@ func (p *Plugin) AddCommands() {
 					CustomID: "tickets-open-" + reason,
 				}
 
-				if emoji != nil {
+				if emoji.ID != 0 || emoji.Name != "" {
```
User9684, line 336: Won't this break non-custom emojis, due to their ID being "0" by default?

jo3-l: The check is an OR, not an AND, and I think the name field should be set for default emojis.

User9684: haha i n oticed before you

PR #1808, opened 2024-12-30
tickets: Make ticket closing prompt more granular

Wolveric:
A server manager recently expressed their concerns, with YAGPDB
unconditionally suggesting it was downloading attachments
sent in tickets. They left this disabled for the privacy of members,
and felt the prompt may scare members, as it contradicted
their stated behaviour.

This PR updates the ticket closing function,
to use a more granular prompt, based on whether
it is creating a transcript of the ticket and/or downloading attachments.

### DIFF
--- a/tickets/tickets_bot.go
+++ b/tickets/tickets_bot.go
@@ -216,6 +216,10 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
 		var closingMsgBuilder strings.Builder
 		closingMsgBuilder.WriteString(closingMsg)
 
+		if conf.TicketsUseTXTTranscripts {
+			closingMsgBuilder.WriteString("\nCreating logs.")
+		}
+
 		closingMsgBuilder.WriteString("\nThis may take a while, if the ticket is long.")
 		closingMsg = closingMsgBuilder.String()
 	}
--- a/tickets/tickets_bot.go
+++ b/tickets/tickets_bot.go
@@ -220,6 +220,10 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
 			closingMsgBuilder.WriteString("\nCreating logs.")
 		}
 
+		if conf.DownloadAttachments {
+			closingMsgBuilder.WriteString("\nDownloading attachments.")
+		}
+
 		closingMsgBuilder.WriteString("\nThis may take a while, if the ticket is long.")
 		closingMsg = closingMsgBuilder.String()
 	}
--- a/tickets/tickets_bot.go
+++ b/tickets/tickets_bot.go
@@ -210,6 +210,16 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
 
 	closingMsg := "Closing ticket."
 
+	// We only need to build up a more detailed closing msg.
+	// if we're creating logs.
+	if conf.TicketsUseTXTTranscripts || conf.DownloadAttachments {
+		var closingMsgBuilder strings.Builder
+		closingMsgBuilder.WriteString(closingMsg)
+
+		closingMsgBuilder.WriteString("\nThis may take a while, if the ticket is long.")
+		closingMsg = closingMsgBuilder.String()
+	}
+
 	// send a heads up that this can take a while
 	common.BotSession.ChannelMessageSend(currentTicket.Ticket.ChannelID, closingMsg)
 
--- a/tickets/tickets_bot.go
+++ b/tickets/tickets_bot.go
@@ -193,8 +193,6 @@ func openTicket(ctx context.Context, gs *dstate.GuildSet, ms *dstate.MemberState
 var closingTickets = make(map[int64]bool)
 var closingTicketsLock sync.Mutex
 
-const closingTicketMsg = "Closing ticket, creating logs, downloading attachments and so on.\nThis may take a while if the ticket is big."
-
 func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.ChannelState, conf *models.TicketConfig, member *discordgo.User, reason string, ctx context.Context) (string, error) {
 	// protect again'st calling close multiple times at the sime time
 	closingTicketsLock.Lock()
@@ -210,8 +208,10 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
 		closingTicketsLock.Unlock()
 	}()
 
+	closingMsg := "Closing ticket."
+
 	// send a heads up that this can take a while
-	common.BotSession.ChannelMessageSend(currentTicket.Ticket.ChannelID, closingTicketMsg)
+	common.BotSession.ChannelMessageSend(currentTicket.Ticket.ChannelID, closingMsg)
 
 	currentTicket.Ticket.ClosedAt.Time = time.Now()
 	currentTicket.Ticket.ClosedAt.Valid = true


Wolveric:
Ok, presumably that constant was used elsewhere... will submit a fix promptly...

Wolveric:
Gonna convert this to a draft, because the interactions with slash commands may make it more complicated.

SoggySaussages:
With commit [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29), would you please test this with the 
“Close with reason” button and let me know what you find?

tickets/tickets_bot.go
```diff
@@ -210,8 +208,28 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
 		closingTicketsLock.Unlock()
 	}()
 
+	closingMsg := "Closing ticket."
+
+	// We only need to build up a more detailed closing msg.
+	// if we're creating logs.
```
jo3-l, line 214:
```suggestion
	// We only need to generate a more detailed closing message if we're creating logs.
```
the full stop is misplaced as is

tickets/tickets_bot.go
```diff
@@ -210,8 +208,28 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
...
+
+		if conf.TicketsUseTXTTranscripts {
+			closingMsgBuilder.WriteString("\nCreating logs.")
+		}
+
+		if conf.DownloadAttachments {
+			closingMsgBuilder.WriteString("\nDownloading attachments.")
+		}
+
+		closingMsgBuilder.WriteString("\nThis may take a while, if the ticket is long.")
```
jo3-l, line 227:
```suggestion
		closingMsgBuilder.WriteString("\nThis may take a while if the ticket is long.")
```
The original message did not include a comma here, and it reads better without.

tickets/tickets_bot.go
```diff
@@ -210,8 +208,28 @@ func closeTicket(gs *dstate.GuildSet, currentTicket *Ticket, ticketCS *dstate.Ch
...
+		var closingMsgBuilder strings.Builder
+		closingMsgBuilder.WriteString(closingMsg)
+
+		if conf.TicketsUseTXTTranscripts {
+			closingMsgBuilder.WriteString("\nCreating logs.")
+		}
+
+		if conf.DownloadAttachments {
+			closingMsgBuilder.WriteString("\nDownloading attachments.")
+		}
```
jo3-l, line 225: The syntax of the generated message is a bit awkward:
```
Closing ticket.
Creating logs.
Downloading attachments.
This may take a while if the ticket is long.
```
I wonder if we can arrange for the message to be more similar to the original. Thoughts?

Wolveric: I felt the *"and so on"* read weirdly, as those are the only two actions it's doing, and the comma version reads weirdly without it.

I could have tried to do a version conditionally inserting *"and"*, but this seemed like the least complex approach...

jo3-l: Fair enough. For completeness, an alternative that is not too complicated is a switch enumerating all the cases, which is not too difficult here:
```go
transcripts, attachments := conf.TicketsUseTXTTranscripts, conf.DownloadAttachments
var closingMsg string
switch {
case transcripts && attachments:
	closingMsg = "Closing ticket, creating logs, and downloading attachments. [...]"
case transcripts:
	...
...
}
```
but it is arguable whether this is a real improvement.

Wolveric: From the standpoint of minimising allocations, it's almost certainly an improvement. I was thinking of making a change of growing the allocation before building the string, but choosing a single option is almost certainly better,

jo3-l: It is definitely more performant, but this code is not on the hot path and we do not make much effort to minimize allocations in general so that concern may be misplaced. The main questions, I think, are quality of the output message, code readability, and extensibility, and it's a bit less clear whether the above suggestion is a meaningful improvement on all three fronts to be worthwhile--I don't have a strong opinion.

SoggySaussages:
[333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.

tickets/tickets_bot.go
Wolveric: Maybe more notably in that case: The switch statement can make for formatting more intuitive...

I previously noted how it would be awkward to insert "and"s into the resulting message, but the switch statement would make directly using those responses easier.

Wolveric:
> [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.

I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...

tickets/tickets_bot.go
Wolveric: Alternatively: The initial line could be added as three statements;
`Closing ticket. Creating logs. Downloading attachments.`

Having looked at my solution, using a `switch` statement, and string builder, it doesn't look better to me...

SoggySaussages:
> > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> 
> I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...

What do you think about taking the builder you have currently and 
separating into a function that can be called in all three places?

Wolveric:
> > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > 
> > 
> > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> 
> What do you think about taking the builder you have currently and separating into a function that can be called in all three places?

That would result in the work being done up to twice, including when it's not necessary.

`tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.

SoggySaussages:
> > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > 
> > > 
> > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > 
> > 
> > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> 
> That would result in the work being done up to twice, including when it's not necessary.
> 
> `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.

I’m not saying don’t construct it conditionally, I’m saying put those
conditions into a function and execute it where needed.

Wolveric:
> > > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > > 
> > > > 
> > > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > > 
> > > 
> > > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> > 
> > 
> > That would result in the work being done up to twice, including when it's not necessary.
> > `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.
> 
> I’m not saying don’t construct it conditionally, I’m saying put those conditions into a function and execute it where needed.

The issue is, that message is only needed in the condition the ticket can be successfully closed. 
Those ephemeral responses aren't based on that condition.

SoggySaussages:
> > > > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > > > 
> > > > > 
> > > > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > > > 
> > > > 
> > > > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> > > 
> > > 
> > > That would result in the work being done up to twice, including when it's not necessary.
> > > `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.
> > 
> > 
> > I’m not saying don’t construct it conditionally, I’m saying put those conditions into a function and execute it where needed.
> 
> The issue is, that message is only needed in the condition the ticket can be successfully closed. Those ephemeral responses aren't based on that condition.

it is still possible for that ticket to fail to close after sending that 
message, therefore that message is never a guarantee that the ticket
is about to successfully close.

SoggySaussages:
> > > > > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > > > > 
> > > > > > 
> > > > > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > > > > 
> > > > > 
> > > > > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> > > > 
> > > > 
> > > > That would result in the work being done up to twice, including when it's not necessary.
> > > > `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.
> > > 
> > > 
> > > I’m not saying don’t construct it conditionally, I’m saying put those conditions into a function and execute it where needed.
> > 
> > 
> > The issue is, that message is only needed in the condition the ticket can be successfully closed. Those ephemeral responses aren't based on that condition.
> 
> it is still possible for that ticket to fail to close after sending that message, therefore that message is never a guarantee that the ticket is about to successfully close.

It’s arguably significantly more likely that the ticket will fail to create
and/or send message logs or fail to delete the channel than someone
will try to spam closing a ticket. Also, in the case that the ticket is in 
fact already being closed the ephemeral response would not be
making a false claim.

Wolveric:
> > > > > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > > > > 
> > > > > > 
> > > > > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > > > > 
> > > > > 
> > > > > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> > > > 
> > > > 
> > > > That would result in the work being done up to twice, including when it's not necessary.
> > > > `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.
> > > 
> > > 
> > > I’m not saying don’t construct it conditionally, I’m saying put those conditions into a function and execute it where needed.
> > 
> > The issue is, that message is only needed in the condition the ticket can be successfully closed. Those ephemeral responses aren't based on that condition.
> 
> it is still possible for that ticket to fail to close after sending that 
> message, therefore that message is never a guarantee that the ticket
> is about to successfully close. 

Ok. 

Regardless, it doesn't make sense to construct it before testing those conditions, then again if those conditions do succeed. It makes more sense to do it in `tickets.closeTicket`, and just have a generic response acknowledging the interaction worked.

SoggySaussages:
> > > > > > > > [333d879](https://github.com/botlabs-gg/yagpdb/pull/1808/commits/333d8794badbdd58e80a9a72c3873dee0903ac29) will need to be reverted to maintain current functionality.
> > > > > > > 
> > > > > > > 
> > > > > > > I don't think I will revert that directly, as it still has the issue of depending on the existing constant. An edited version with a more generic response should be reasonable though...
> > > > > > 
> > > > > > 
> > > > > > What do you think about taking the builder you have currently and separating into a function that can be called in all three places?
> > > > > 
> > > > > 
> > > > > That would result in the work being done up to twice, including when it's not necessary.
> > > > > `tickets.closeTicket` can exit before attempting to log & delete the ticket, so it only makes sense to construct that prompt conditionally.
> > > > 
> > > > 
> > > > I’m not saying don’t construct it conditionally, I’m saying put those conditions into a function and execute it where needed.
> > > 
> > > 
> > > The issue is, that message is only needed in the condition the ticket can be successfully closed. Those ephemeral responses aren't based on that condition.
> > 
> > 
> > it is still possible for that ticket to fail to close after sending that
> > message, therefore that message is never a guarantee that the ticket
> > is about to successfully close.
> 
> Ok.
> 
> Regardless, it doesn't make sense to construct it before testing those conditions, then again if those conditions do succeed. It makes more sense to do it in `tickets.closeTicket`, and just have a generic response acknowledging the interaction worked.

The “conditions” you’re referring to are just <ticket is already being 
closed>, right? In either case, the ephemeral message would not be 
making a false claim in this case. 

I still maintain that in most instances people are going to be seeing
these two messages pop up for about half a second before the 
channel is deleted. If you give two different messages, you’re just 
going to increase the amount of text that needs to be read in that
time.

tickets/tickets_bot.go
SoggySaussages: What about a slice of actions joined by commas and switched to title case
at the end?

Wolveric:
We should outline the potential approaches here, from the user and technical perspective:

1. Remove he ephemeral response entirely (current change):
    - Technical points:
        - Removes any extra work outright
        - Discord will get in a huff for improperly handling the interaction
    - User points:
        - If user sees the interaction complete before the channel is deleted, they will be shown an error, which is confusing if YAGPDB handles closing the ticket successfully
2. Add a method for constructing the closing prompt:
    - Technical points:
        - Discord is happy, because we responded to the interaction
        - We're building the response in both the button handlers, and `tickets.closeTicket`
    -  User points:
        - If the user sees the interaction complete, they won't see any errors
        - If the user sees the interaction complete, they will see the closing message parroted soon afterwards, which is now redundant info to them, and may be confusing
3. Use a generic response for button interactions:
    - Technical points:
        - Discord is happy, because we responded to the interaction
        - We're spending very little time on the response, if we use a constant string
    - User points:
        - If the user sees the interaction complete, they won't see any errors
        - If they see the generic response, they will know YAGPDB is trying to close the ticket
        - If they see the follow-up message from `tickets.closeTicket`, they can see what actions YAGPDB is trying to do.

tickets/tickets_bot.go
Wolveric: I don't particularly see how that helps. It would create a single source for changing any separators, yes, but would largely result in checking the same conditions anyway...

SoggySaussages: Just joins it all into one sentence instead of three separate ones.
maybe I’m not understanding what the current issue is.

Wolveric: The initial quibble, was about the fact I changed the response to be separated onto different lines, and the message reading somewhat strangely as a result. I kind of agree with that in hindsight, and I think the three statements could be on the same line.

So basically, just swapping `\n` for ` `

SoggySaussages:
Your first user point for option two is inaccurate, and I’m not sure 
what your user point for option one is. I don’t see how seeing two of the.

Wolveric:
Confusion might be overstating the response to the message being doubled up, but they may question, *"Why did the bot post that message twice...??"*

If YAGPDB was already attempting to close the ticket though, a message stating YAGPDB is attempting to close the ticket, then another saying it's already closing the ticket, almost certainly would be.

Alternatively, a generic response of, *"Attempting to close ticket."* makes a lot more sense followed up with, *"Already working on closing this ticket, please wait..."*

SoggySaussages:
I think there is a difference between questioning something and 
being confused. I don’t see how a message stating that the bot is
attempting to close the ticket followed up by a message that the
ticket is already being closed could be confusing. Both messages 
would be 100% accurate. Notably, the ticket closing message does 
not mention _ initializing_ the closing of the ticket, only that it is in 
present tense occurring.

Wolveric:
I think using `"Attempting to close ticket"` would be the preferable response for the ephemeral message. It could fit both with the response for YAGPDB already closing the ticket, and when its starting a new closing process.

SoggySaussages:
Imo, given the choice between having a user question something, 
and having a user be frustrated that they don’t have any time to 
read two separate messages before the entire channel disappears, 
we should choose the former. My guess is that most users will 
understand that the second message is for the benefit of the entire 
channel, seeing that the first message has the only you can see 
this message note beneath it and the second message doesn’t.

Wolveric:
I think something this PR has highlighted, is that there's too much functionality wrapped up in `tickets.closeTicket`, so I'm going to convert it back to a draft for the time being, and look at redoing it as a rework of the ticket system in future.

PR #1813, opened 2025-01-03
bot: declare guild emojis intent

jo3-l:
This lets us receive guild emoji updates and update the cached Emojis fields accordingly.


jo3-l:
The second commit updates the intent names and associated comments to match https://discord.com/developers/docs/events/gateway#gateway-intents. (Some intents have been renamed and many have corresponding events updated.)

I have opted to not add the new `GUILD_MESSAGE_POLLS` and `DIRECT_MESSAGE_POLLS` intents here, since we do not use those intents or related events, nor do we support polls at the moment. A future PR implementing the relevant poll endpoints and gateway events should add those intents instead.

PR #1814, opened 2025-01-08
logs: fix login link in pnotify alert

galen8183:
Add a missing quotation mark to properly display the login hyperlink
prompt when trying to view logs restricted to server members.
Reword the alert to be more accessible (include "login" in the hyperlink
text).

Signed-off-by: Galen CC <galen8183@gmail.com>

### DIFF
--- a/logs/web.go
+++ b/logs/web.go
@@ -239,8 +239,8 @@ func CheckCanAccessLogs(w http.ResponseWriter, r *http.Request, config *models.G
 	member := web.ContextMember(ctx)
 	if member == nil {
 		goTo := url.QueryEscape(r.RequestURI)
-		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s>here</a>`, web.BaseURL(), goTo)
-		alertMsg := fmt.Sprintf("This server has restricted log access to members only.\nIf you are a member, click %s to login.", alertLink)
+		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s">login here</a>`, web.BaseURL(), goTo)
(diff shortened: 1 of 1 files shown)


logs/web.go
```diff
@@ -239,8 +239,8 @@ func CheckCanAccessLogs(w http.ResponseWriter, r *http.Request, config *models.G
 	member := web.ContextMember(ctx)
 	if member == nil {
 		goTo := url.QueryEscape(r.RequestURI)
-		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s>here</a>`, web.BaseURL(), goTo)
-		alertMsg := fmt.Sprintf("This server has restricted log access to members only.\nIf you are a member, click %s to login.", alertLink)
+		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s">login here</a>`, web.BaseURL(), goTo)
```
l-zeuch, line 242:
```suggestion
		alertLink := fmt.Sprintf(`<a href="%s/login?goto=%s">please login via Discord</a>`, web.BaseURL(), goTo)
```
It is my understanding that any combinations with "here" are discouraged. Instead, web authors are asked to reword hyperlinks such that they clearly outline which action is going to happen when a user clicks that link.

(Of note, with this suggestion the hyperlink may be too long, I just decided to keep the suggest diff low-noise.)

galen8183: "If you are a member, please [login with Discord] to view this log." reads better to me.
I wonder if we should also omit "If you are a member" since it's a redundant qualifier, WDYT?

l-zeuch: Sounds good to me!

galen8183: Thanks, done!

PR #1820, opened 2025-01-12
Custom Commands: New error output for regex cache limit

H1nr1:
When a custom command uses more than 10 different regular expressions, the CC would error stating "too many discord api calls function." As CC regex functions do not use the Discord API, this is misleading and causes confusion around the issue.

Discord_4B0cghSwf1

This PR adds a new error, ErrRegexCacheLimit, which instead outputs "too many different regular expressions."
This PR also rewords and reformats the ErrTooManyAPICalls error to read "too many Discord API calls."

### DIFF
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -24,7 +24,7 @@ import (
 var (
 	ErrTooManyCalls    = errors.New("too many calls to this function")
 	ErrTooManyAPICalls = errors.New("too many potential Discord API calls")
-	ErrRegexCacheLimit = errors.New("too many different regular expressions")
+	ErrRegexCacheLimit = errors.New("too many unique RegExes")
 )
 
 func (c *Context) tmplSendDM(s ...interface{}) string {
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -24,7 +24,7 @@ import (
 var (
 	ErrTooManyCalls    = errors.New("too many calls to this function")
 	ErrTooManyAPICalls = errors.New("too many potential Discord API calls")
-	ErrRegexCacheLimit = errors.New("too many unique RegExes")
+	ErrRegexCacheLimit = errors.New("too many unique regular expressions (regex)")
 )
 
 func (c *Context) tmplSendDM(s ...interface{}) string {
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -23,7 +23,8 @@ import (
 
 var (
 	ErrTooManyCalls    = errors.New("too many calls to this function")
-	ErrTooManyAPICalls = errors.New("too many potential discord api calls function")
+	ErrTooManyAPICalls = errors.New("too many potential Discord API calls")
+	ErrRegexCacheLimit = errors.New("too many different regular expressions")
 )
 
 func (c *Context) tmplSendDM(s ...interface{}) string {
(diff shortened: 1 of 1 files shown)


common/templates/context_funcs.go
```diff
@@ -23,7 +23,8 @@ import (
 
 var (
 	ErrTooManyCalls    = errors.New("too many calls to this function")
-	ErrTooManyAPICalls = errors.New("too many potential discord api calls function")
+	ErrTooManyAPICalls = errors.New("too many potential Discord API calls")
+	ErrRegexCacheLimit = errors.New("too many different regular expressions")
```
SoggySaussages, line 27: Perhaps “unique” instead of “different” here? I also feel like we should
keep with the name “RegEx” for consistency, despite the unfortunate
RAS syndrome that would create.

H1nr1: I was looking for a word other than "different;" I didn't like it. "Unique" is much better, will do!
I didn't shorthand regular expression for, honestly, I didn't like the plural version of it: "regexes." However, "too many unique RegExes" reads pretty easily.

SoggySaussages: What are your thoughts on using RegEx Expressions and
ignoring the resulting RAS Syndrome in favour of potentially
reducing the cognitive barrier?

H1nr1: I think it should be spelled out entirely at that point; supporting the no RAS syndrome there. My thought is "regular expression" would bring exactly what the user is looking for if not familiar and needed to Google it, even though "regex" would too.

H1nr1: While "unique RegExes" read well, I'm not liking at all how it looks, and couldn't imagine how a user unfamiliar to the full phrase would react.

lemm-e: A user not familiar with a regex wouldn't be using a regex function

jo3-l: To contribute a new color to this bikeshed, where else do we use the phrase "RegEx" (with that exact capitalization?) I am not really a fan and prefer just "regex" or "regular expression", as in the original commit.

H1nr1: > A user not familiar with a regex wouldn't be using a regex function

Maybe that idea doesn't hold entirely. But for ease of being able to find help, for whatever the reason, I believe "regular expressions" fits best here.

lemm-e: I don't quite understand what ease the full terminology provides given the fact that "regex" is how we refer to them in the custom command trigger type itself, in the documentation and everywhere else. It isn't an abbreviation we've come up with and is something recognised and used globally.

common/templates/context_funcs.go
```diff
@@ -23,7 +23,8 @@ import (
 
 var (
 	ErrTooManyCalls    = errors.New("too many calls to this function")
-	ErrTooManyAPICalls = errors.New("too many potential discord api calls function")
+	ErrTooManyAPICalls = errors.New("too many potential Discord API calls")
+	ErrRegexCacheLimit = errors.New("too many unique RegExes")
```
ashishjh-bst, line 27: call it "too many regular expressions ( regex)  used" and end all the unnecessary debate.

H1nr1: I don't mean to continue it but I believe "unique" is a key work here and should be kept. Else it could be taken as too many regex function calls, and not just unique regular expressions.

ashishjh-bst: this is pondering over nothing, just give it a meaningful description and end it.

ashishjh-bst: if people complain, raise another pr to improve it.

PR #1832, opened 2025-01-23
add -bots switch to the clean command

Borbot33:
I like sometimes to delete bot message only.

Just filters if messages are sent by a bot.


moderation/commands.go
```diff
@@ -1396,6 +1401,12 @@ type MessagesWithAttachmentsFilter struct{}
 func (*MessagesWithAttachmentsFilter) Matches(msg *dstate.MessageState) (delete bool) {
 	return len(msg.GetMessageAttachments()) > 0
 }
+// Only delete messages with attachments.
+type BotMessagesOnly struct{}
```
jo3-l, line 1405:
```suggestion

// Only delete bot messages.
type BotMessagesFilter struct{}
```
Fix the comment and make the struct name consistent with the other filters.

Borbot33: Okay, changed. Thank you :3

PR #1834, opened 2025-01-26
moderation: improve external mod action logging

l-zeuch:
This PR aims to overall improve the performance of reporting external
moderation actions (i.e. actions not made through YAGPDB), by utilising
the Guild Audit Log Entry Create event fired by the Gateway whenever a
new audit log entry was created.

Due to the now more centralised nature of handling external mod actions,
we can remove some now obsolete event handlers as well, with an
additional (though probably subjective) benefit to readability of the
relevant code.

Testing reveals a rather significant speedup in posting to the mod-log
compared to the previous implementation, simply because we no longer
wait three seconds to fetch the audit log due to a potential for race
conditions. I confirmed, to the best of my ability, that the new
implementation is functionally equivalent to the old one.

This is a combination of three commits:

* discordgo: add guild_id extra field to AuditLogEntry

  When receiving Guild Audit Log Entry Create events, Discord sends an
  Audit Log entry object with an extra guild_id key as inner payload. Add
  that field such that we can use it later.

* moderation: use guild audit log entry create event

  Use the Guild Audit Log Entry Create event instead of subscribing to
  individual events and having to (slowly) parse through the guild's audit
  log.
  That way, we can significantly improve the speed of reporting external
  moderation actions not made with YAGPDB.

* moderation: remove now obsolete event handlers

  601d0a9876 (moderation: use guild audit log entry create event)
  introduced a new way to handle external moderation action logging,
  making some event handlers now obsolete. Remove them.

### DIFF
--- a/moderation/plugin_bot.go
+++ b/moderation/plugin_bot.go
@@ -50,11 +50,8 @@ func (p *Plugin) BotInit() {
 	scheduledevents2.RegisterLegacyMigrater("unmute", handleMigrateScheduledUnmute)
 	scheduledevents2.RegisterLegacyMigrater("mod_unban", handleMigrateScheduledUnban)
 
-	eventsystem.AddHandlerAsyncLastLegacy(p, bot.ConcurrentEventHandler(HandleGuildBanAddRemove), eventsystem.EventGuildBanAdd, eventsystem.EventGuildBanRemove)
-	eventsystem.AddHandlerAsyncLast(p, HandleGuildMemberRemove, eventsystem.EventGuildMemberRemove)
 	eventsystem.AddHandlerAsyncLast(p, LockMemberMuteMW(HandleMemberJoin), eventsystem.EventGuildMemberAdd)
 	eventsystem.AddHandlerAsyncLast(p, LockMemberMuteMW(HandleGuildMemberUpdate), eventsystem.EventGuildMemberUpdate)
-	eventsystem.AddHandlerAsyncLast(p, HandleGuildMemberTimeoutChange, eventsystem.EventGuildMemberUpdate)
 
 	eventsystem.AddHandlerAsyncLast(p, HandleGuildAuditLogEntryCreate, eventsystem.EventGuildAuditLogEntryCreate)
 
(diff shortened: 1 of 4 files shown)


l-zeuch:
Sidenote: it may be useful to review this commit-by-commit--especially b659f8b (moderation: remove now obsolete event handlers) should receive additional scrutiny, lest that we accidentally remove a handler that we do not account for with audit log entry crate.

moderation/plugin_bot.go
```diff
@@ -302,194 +301,75 @@ func RefreshMuteOverrideForChannel(config *Config, channel dstate.ChannelState)
...
+			logger.WithError(err).WithField("guild", data.GuildID).Error("Failed sending kick log message")
+		}
 	}
 
-	// The bot only unbans people in the case of timed bans
-	if botPerformed {
-		author = common.BotUser
-		reason = "Timed ban expired"
+	if config.LogBans && *data.ActionType == discordgo.AuditLogActionMemberBanAdd {
+		err = CreateModlogEmbed(config, &author.User, MABanned, target, data.Reason, "")
```
galen8183, line 359: Previously we logged scheduled unban events here, since this combines all the modlog events into one, we erroneously no longer accept *any* bot user authored events.

This scheduled unban modlog could either be reimplemented just before checking the audit log entry author, or in the scheduled unban handler. I think I prefer the former, WDYT?

moderation/plugin_bot.go
```diff
@@ -302,194 +301,75 @@ func RefreshMuteOverrideForChannel(config *Config, channel dstate.ChannelState)
...
 	}
 
-	err = CreateModlogEmbed(config, author, action, user, reason, "")
-	if err != nil {
-		logger.WithError(err).WithField("guild", guildID).Error("Failed sending " + action.Prefix + " log message")
+	if config.LogUnbans && *data.ActionType == discordgo.AuditLogActionMemberBanRemove {
+		err = CreateModlogEmbed(config, &author.User, MAUnbanned, target, data.Reason, "")
 	}
```
galen8183, line 364: With the above changes, this event checking can quite nicely use a switch statement instead:
```go
	// setup done, now we get to the actions.
	switch {
	case config.LogTimeouts && *data.ActionType == discordgo.AuditLogActionMemberUpdate:
		err = CreateModlogEmbed(config, &author.User, MATimeoutAdded, target, data.Reason, "")
	case config.LogKicks && *data.ActionType == discordgo.AuditLogActionMemberKick:
		err = CreateModlogEmbed(config, &author.User, MAKick, target, data.Reason, "")
	case config.LogBans && *data.ActionType == discordgo.AuditLogActionMemberBanAdd:
		err = CreateModlogEmbed(config, &author.User, MABanned, target, data.Reason, "")
	case config.LogUnbans && *data.ActionType == discordgo.AuditLogActionMemberBanRemove:
		err = CreateModlogEmbed(config, &author.User, MAUnbanned, target, data.Reason, "")
	}
```

l-zeuch:
Thanks @galen8183 for the review and providing a `git am`-ble patch!
I committed and pushed your changes, with your changes this should be good to go now.

moderation/plugin_bot.go
galen8183: This is still an outstanding issue, scheduled unbans by yag are no longer logged.

PR #1851, opened 2025-02-23
Tickets: DM ticket owner with transcript

Ranger-4297:
This PR adds a DM with the transcript attached as a file to the owner of the ticket
If the DM fails, i.e the user has the bot blocked then nothing happens.

I am open to suggestions or ideas on changing/improving this like mentioned in the support server
Given this is my first contribution I doubt it is perfect.
image


ashishjh-bst:
This will not be added to Prime yag, I will be closing this PR as this can lead to DM spam, and DMs have quite a bit of ratelimits, and you cannot just send random attachments to people.

SoggySaussages:
What do you think about locking this particular feature 
behind an env var and disabling on botlabs yag?

ashishjh-bst:
> What do you think about locking this particular feature behind an env var and disabling on botlabs yag?

I don't want to add things I don't want to support to a repository that I am responsible for maintaining

PR #1862, opened 2025-03-19
templates: fix whitespace trim

SoggySaussages:
Adjusts the whitespace trim for templates limitedWriter to ensure trimming leading whitespace does not inadvertently trim
trailing whitespace from the initial call of `Write()` to the writer. This was resulting in the first bytes written to a limitedWriter
having trailing whitespace trimmed, even if it didn't have any leading whitespace, which is undesirable.

Before vs. After 902ced5
Screenshot 2025-03-19 at 05 06 34

Before vs. After 902ced5 with a leading whitespace challenge
Screenshot 2025-03-19 at 05 16 02


SoggySaussages:
Before and after with a database challenge. Each run of the `-fish` command fishes 25 times, failing either when the 
database write encounters the inadvertent trailing whitespace trim on a previous iteration, or when the max DB calls is 
reached. Below is the equivalent of 50 writes and 50 reads of different values before and after. This demonstrates with
significant confidence that this fix remedies the database write issue identified as a symptom of this bug.

Screenshot 2025-03-19 at 05 21 42
Screenshot 2025-03-19 at 05 22 00

PR #1872, opened 2025-04-16
cc/interactions: add response deferral

SoggySaussages:
Add the ability to configure a deferral to interaction CCs. These deferrals are calculated and then executed before
sequential CC execution. They are also executed in the concurrent event handler, so they should be resilient to many
members triggering the same interaction commands in a short timeframe.


customcommands/models/boil_view_names.go
ashishjh-bst, line 1: This file is not needed

PR #1875, opened 2025-04-19
Add support for editing target user's nickname in templates

LoayGhreeb:
We have a case where we want to change a member's nickname as admins.
The only available way is through [`editNickname`](https://help.yagpdb.xyz/docs/reference/templates/functions/#editnickname), but it only works for the member who triggered the command.
This PR adds `editTargetNickname` to allow changing someone else's nickname by mention or ID.
Should I submit a separate PR to update the docs as well?

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -739,6 +739,7 @@ func baseContextFuncs(c *Context) {
 	c.addContextFunc("getMember", c.tmplGetMember)
 	c.addContextFunc("getMemberVoiceState", c.tmplGetMemberVoiceState)
 	c.addContextFunc("editNickname", c.tmplEditNickname)
+	c.addContextFunc("editTargetNickname", c.tmplEditTargetNickname)
 
 	// Thread functions
 	c.addContextFunc("addThreadMember", c.tmplThreadMemberAdd)
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -1897,7 +1897,7 @@ func (c *Context) tmplOnlineCountBots() (int, error) {
 	return 0, nil
 }
 
-func (c *Context) tmplEditNickname(Nickname string) (string, error) {
+func (c *Context) tmplEditTargetNickname(target interface{}, Nickname string) (string, error) {
 	if c.IncreaseCheckCallCounter("edit_nick", 2) {
 		return "", ErrTooManyCalls
 	}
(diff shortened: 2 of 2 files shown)


ashishjh-bst (commented):
This should have a permission check on the triggering user to see if they have Manage Nicknames permission. Otherwise you're just allowing anyone with access to this custom command to modify nicknames in your server. 

I will have to look into the abuse aspect of it before considering this to be merged.

LoayGhreeb:
> Otherwise you're just allowing anyone with access to this custom command to modify nicknames in your server.

Not really. It depends on how the custom command is set up. When creating a command that uses this function, we can set restrictions on which users or roles are allowed to execute it. At the end of the day, it's the responsibility of the server owner or moderators to manage those permissions properly.

ashishjh-bst:
> > Otherwise you're just allowing anyone with access to this custom command to modify nicknames in your server.
> 
> Not really. It depends on how the custom command is set up. When creating a command that uses this function, we can set restrictions on which users or roles are allowed to execute it. At the end of the day, it's the responsibility of the server owner or moderators to manage those permissions properly.

That isn't true. An average discord user is not very technical, and will end up doing something stupid. You cannot give a gun to an average human and assume that they will ve responsible with it. Anything with moderation or member updates will have to have a mandatory check for permissions, a responsible discord mod will make sure people who use the command have this perm, or make alternate arrangements. 

Keeping the yet to he looked into abuse aspect of this aside. This is non-negotiable.

RhykerWells:
Does that mean that in the future templates that give roles and others will have these checks incorporated into the context function?
If not, why different for this `editTargetNickname` function?

SoggySaussages:
> Does that mean that in the future templates that give roles and such will have these checks incorporated into the context function? If not, why different for this `editTargetNickname` function?

Role assignments are different than direct visible member updates, which would need to be more heavily restricted. You’re directly altering someone’s identity on the server as opposed to roles which are hidden to the member context, unless hoisted.

LoayGhreeb:
> Anything with moderation or member updates will have to have a mandatory check for permissions, a responsible discord mod will make sure people who use the command have this perm, or make alternate arrangements.

Got it, that makes sense. So to move forward with the PR, should I add a permission check to ensure that the user calling this function has the right to manage nicknames? Or would the PR still be unlikely to be accepted due to potential for abuse?

Also, if I do add the permission check, should it only apply to `editTargetNickname`, and not `editNickname`?

ashishjh-bst:
Hi, this is unliked to added after quite a detailed discussion internally. Please feel free to selfhost.

PR #1889, opened 2025-05-12
Update swearwords.go

yzkxi3:
can u please accept this for me


ashishjh-bst:
- Read contributing guidelines. 
- Explain the reasoning behind a PR. 

Random deletions with the comment "Can you merge this for me" is not good enough.

savage4618:
I see all you did was remove some pretty common swear words. Instead of changing the master swear words list in the global bot, you probably want to implement a custom list of words for your server.

yzkxi3:
Yes

yzkxi3:
Can yall accept the one i made for my server

ashishjh-bst:
This would remove these swearwords from all servers the bot is on.

yzkxi3:
Yh ik i cleared some swear word i dont wana be filtered like fuck and those

Wolveric:
Why do you think those should be removed for everyone...? That's pretty specifically your opinion, even if the list is already opinionated...

PR #1896, opened 2025-06-16
logs/whois: show roles of target member

l-zeuch:
Show roles of the target member in the response of `whois` as well.
This one was often requested, I'm just sending this patch to make these
requests stop (there were quite a lot).

---

| Before | After |
| ------ | ----- |
| before | after |

### DIFF
--- a/logs/plugin_bot.go
+++ b/logs/plugin_bot.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 	"time"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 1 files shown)


l-zeuch:
I have received expert knowledge that some bozos have a truckload of roles on themselves/their members, possibly exceeding 1024 chars.
Will add some logic that goes around that, either by truncating or by extending. (or using the description instead :thinking:)

l-zeuch:
I have resolved this silly problem by just showing the first 25 roles and then defaulting to `and more` at the end of the string.

PR #1897, opened 2025-06-18
Replace problematic work in roasts with alternative

mkwarman:
People in my server love the `-roast` command but don't love the use of the word "Fucktard". Obviously roasts aren't meant to be nice, but in many communities people are trying not to use "retard" as in insult nowadays since it is offensive to people with mental handicaps, so I figured a change from "Fucktard" to "Fuckface" would be a nice alternative.


H1nr1:
Semantically, the change in word use does not make sense. If the use of "fuck" is acceptable, I instead suggest using "dumbass" rather than "fuckface" here. The intention of the original roast was to call someone dumb, stupid, etc. The changes to use "fuckface" suggest the roast to call the user ugly, but this does not fit with the rest of the sentence. "Dumbass" keeps the same idea as the original roast while removing the potentially problematic part.

mkwarman:
@H1nr1 updated to "dumbass" here: https://github.com/botlabs-gg/yagpdb/pull/1898

PR #1902, opened 2025-06-24
Add shardorchestrator usage guide to cmd/shardorchestrator/README.md

egelhaus:


l-zeuch (changes requested):
Notwithstanding that I find it questionable if any selfhosted instance ever reaches a scale where sharding is necessary, I have left a review with my thoughts, simplifying and omitting unnecessary detail.

egelhaus:
I have fixed everything @l-zeuch

PR #1909, opened 2025-07-14
Update adjectives.go

savage4618:
Remove `airline` from adjectives as it is not an adjective, but a noun. (I checked multiple sources)


ashishjh-bst:
https://www.dictionary.com/browse/airline

PR #1911, opened 2025-07-16
fix(web): respect X-Forwarded-Proto to prevent unnecessary HTTPS redirects behind proxy

MiguVT:
When running yagpdb behind a reverse proxy (such as nginx) that handles HTTPS, the Go web server could not correctly detect when the original request was HTTPS, causing infinite or unnecessary HTTPS redirects.   Now, before redirecting to HTTPS, the code also checks the X-Forwarded-Proto header.   This ensures no redirect happens when the original connection was secure, even if the proxy communicates with yagpdb via HTTP.

PR help users self-hosting for reverse-proxy, no doc changes needed. ty for this amazing project <3

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -231,6 +231,21 @@ func runServers(mainMuxer *goji.Mux) {
 		if err != nil {
 			logger.Error("Failed http ListenAndServe:", err)
 		}
+	} else if exthttps {
+		// External HTTPS is handled by nginx/reverse proxy
+		// Only serve HTTP but application logic assumes HTTPS URLs
+		logger.Info("Starting yagpdb web server http:", ListenAddressHTTP, " (external HTTPS)")
+
+		server := &http.Server{
+			Addr:        ListenAddressHTTP,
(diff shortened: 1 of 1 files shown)


MiguVT:
Oh mb, didnt saw that branch, thinked it was master because last commit was yesterday

ashishjh-bst:
> Oh mb, didnt saw that branch, thinked it was master because last commit was yesterday

No issues, I'll review and merge whenever you have the PR read and not raised as a Draft.

MiguVT:
Im making a fix and refactoring the http system

ashishjh-bst:
have you tried running it with the https flag as false and -extHttps flag as true?  I think that will fix your problem and none of the changes would be needed.  In case you want to test, there is a docker-compose file for that in yagpdb_docker.

MiguVT:
I found a bug, will fix it now

MiguVT:
Closed cause added into https://github.com/botlabs-gg/yagpdb/pull/1913

PR #1915, opened 2025-07-17
Fix External HTTPS Support for Reverse Proxy

MiguVT:
**Changes:**
- Modified `runServers()` to properly handle `exthttps` mode - now only starts HTTP server when behind reverse proxy
- Added `IsHTTPS()` function to detect HTTPS from nginx headers (`X-Forwarded-Proto`, `X-Scheme`)
- Updated middleware to conditionally set HSTS headers based on actual protocol detection

**How it works:**
When `exthttps=true`, YAGPDB listens only on HTTP but generates HTTPS URLs and detects the real protocol from proxy headers for security features.

*Note: This implementation was done carefully and thoughtfully, unlike my previous hasty attempt. The solution now properly integrates with existing codebase patterns and handles edge cases correctly.*

### DIFF
--- a/web/middleware.go
+++ b/web/middleware.go
@@ -57,8 +57,10 @@ func MiscMiddleware(inner http.Handler) http.Handler {
 			"url": r.URL.Path,
 		})
 		ctx = context.WithValue(ctx, common.ContextKeyLogger, entry)
-		// force https for a year
-		w.Header().Set("Strict-Transport-Security", "max-age=31536000")
+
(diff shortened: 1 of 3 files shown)


MiguVT:
Understood. Maybe updating the documentation would help, as it should mention something like:
"If you're using a reverse proxy for HTTPS, you need to run the bot with the `--https` flag set to `false` and the `--extHttps` flag set to `true`."

I would make a simple PR for that

PR #1920, opened 2025-07-28
README: clarify and simplify HTTPS setup instructions in README

MiguVT:
- Added clear explanations for both built-in and reverse proxy HTTPS configurations.
- Explained what `--https=false` and `--extHttps=true` do in simple terms. (For more long terms the user should investigate more)

Checked [Documentation](https://github.com/botlabs-gg/yagpdb-docs-v2) and I didnt find anything releated to this (Self-hosting setup)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,32 @@ In addition, you will need to add the following urls to the bot's "REDIRECT URI(
 * <https://YourHostNameHere/confirm_login>
 * <https://YourHostNameHere/manage>
 
+### HTTPS Options
+
+YAGPDB supports two main HTTPS setups:
+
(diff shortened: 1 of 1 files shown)


MiguVT:
httpsKey and httpsCert doesn't exist, tell me if i should implement it or just make it more simple and keep it built-in self-signed cert

MiguVT:
(Dont merge it, I added --httpsCert=/path/to/cert.pem --httpsKey=/path/to/key.pem because i think it can be a good feature but want to know suggestion from contributors and if it isnt needed or should add it, if we should add it i will implement it here)
image

PR #1921, opened 2025-07-30
Added a fallback to get thread info from discord

ashishjh-bst:
This is needed in scenarios where a thread is archived and user wants to do some operation on that, since archived threads are not stored in state, this often causes issues.


l-zeuch (commented):
Thanks for trying to make thread management a little more useful in CC; I tested out your changes, unfortunately they don't work? :thinking:

Started a quick debugging session, this is what I found:

At the moment, we're never even getting past this line if the thread is no longer in state, because `c.ChannelArg` via `c.baseChannelArg` attempts to grab the channel/thread from `c.GS`, i.e. our state.

PR #1972, opened 2025-10-27
templates: consider `uint`s for comparison in `in`

l-zeuch:
For unknown reasons, `in` did not consider `uint`s for comparable types,
forcing users to a) be extremely confused why their code doesn't work
when it should and b) figuring a way around that (either by type
conversion or some other arguably unnecessary things).

Adjust the definition of `in` to consider `uint`s.
Future considerations: can we somehow rewrite this using generics,
elimiating the need to exhaustively list all possible comparable types?

### DIFF
--- a/common/templates/general.go
+++ b/common/templates/general.go
@@ -767,46 +767,31 @@ func in(l interface{}, v interface{}) bool {
 	lv, _ := indirect(reflect.ValueOf(l))
 	vv := reflect.ValueOf(v)
 
-	if !reflect.ValueOf(vv).IsZero() {
-		switch lv.Kind() {
-		case reflect.Array, reflect.Slice:
-			for i := 0; i < lv.Len(); i++ {
-				lvv := lv.Index(i)
-				lvv, isNil := indirect(lvv)
-				if isNil {
-					continue
-				}
-				switch lvv.Kind() {
-				case reflect.String:
-					if vv.Type() == lvv.Type() && vv.String() == lvv.String() {
-						return true
-					}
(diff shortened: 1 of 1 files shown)


common/templates/general.go
```diff
@@ -788,6 +788,13 @@ func in(l interface{}, v interface{}) bool {
 							return true
 						}
 					}
+				case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
```
jo3-l: It might be easier to use `lvv.CanUint()` (resp. `lvv.CanFloat()`, `lvv.CanInt()` here instead. (Note that this will require rewriting the switch as `switch { /* ... */ }` instead of a switch on `lvv.Kind()`.)

l-zeuch: Pushed the requested changes; the switch statement now uses the respective type's `Can*`.

I also went ahead and changed some prior checks around, decreasing indentation. While I was at it I satisfied `gopls/modernize` by modernising the for loop. Seems to still work as intended.

common/templates/general.go
```diff
@@ -767,39 +767,31 @@ func in(l interface{}, v interface{}) bool {
...
+			if isNil {
+				continue
 			}
-		case reflect.String:
-			if vv.Type() == lv.Type() && strings.Contains(lv.String(), vv.String()) {
-				return true
+			switch {
+			case lvv.Kind() == reflect.String:
+				return vv.Type() == lvv.Type() && vv.String() == lvv.String()
+			case lvv.CanInt() && vv.CanInt():
```
jo3-l: Nice, like the symmetry here!

PR #2049, opened 2026-04-06
Make ComplexMessageEdit redundant, because from a CC point of view it doesn't make sense to have a separate builder for editing a message. 

ashishjh-bst:


galen8183:
Perhaps the conversion from MessageSend to MessageEdit should be handled in a dedicated function rather than duplicating it thrice

ashishjh-bst:
> Perhaps the conversion from MessageSend to MessageEdit should be handled in a dedicated function rather than duplicating it thrice

makes sense, will do!

PR #2054, opened 2026-04-24
 customcommands: new hasAnyPermissions functions

l-zeuch:
As it is, the hasPermissions function checks whether *all* bits are
set. To test for any bit, users currently need to write their own logic,
which is quite unergonomic.

Add two new function, targetHasAnyPermissions and hasAnyPermissions that
checks whether at least one permission bit is set, allowing for more
ergonomic permission checking when only one of many permissions is
necessary to the user.

### DIFF
--- a/common/templates/context.go
+++ b/common/templates/context.go
@@ -850,7 +850,9 @@ func baseContextFuncs(c *Context) {
 
 	// Permission functions
 	c.addContextFunc("hasPermissions", c.tmplHasPermissions)
+	c.addContextFunc("hasAnyPermissions", c.tmplHasAnyPermissions)
 	c.addContextFunc("targetHasPermissions", c.tmplTargetHasPermissions)
+	c.addContextFunc("targetHasAnyPermissions", c.tmplTargetHasAnyPermissions)
 	c.addContextFunc("getTargetPermissionsIn", c.tmplGetTargetPermissionsIn)
 
 	// Channel functions
--- a/common/templates/context_funcs.go
+++ b/common/templates/context_funcs.go
@@ -773,22 +773,18 @@ func (c *Context) tmplHasPermissions(needed int64) (bool, error) {
 		return false, ErrTooManyAPICalls
 	}
 
-	if c.MS == nil {
-		return false, nil
-	}
-
-	if needed < 0 {
-		return false, nil
-	}
-
-	if needed == 0 {
-		return true, nil
-	}
-
-	return c.hasPerms(c.MS, c.CurrentFrame.CS.ID, needed)
+	return c.hasPerms(c.MS, c.CurrentFrame.CS.ID, needed, false)
 }
 
-func (c *Context) tmplTargetHasPermissions(target interface{}, needed int64) (bool, error) {
+func (c *Context) tmplHasAnyPermissions(needed int64) (bool, error) {
+	if c.IncreaseCheckGenericAPICall() {
+		return false, ErrTooManyAPICalls
+	}
+
+	return c.hasPerms(c.MS, c.CurrentFrame.CS.ID, needed, true)
+}
+
+func (c *Context) tmplTargetHasPermissions(target any, needed int64) (bool, error) {
 	if c.IncreaseCheckGenericAPICall() {
 		return false, ErrTooManyAPICalls
 	}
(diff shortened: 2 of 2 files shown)


jo3-l (commented):
I think that a boolean flag is confusing in this context; it is not clear what `hasPermissions ... true` does at a glance. If this functionality is desirable, what about adding a separate function `hasAnyPermissions` instead?

l-zeuch:
> I think that a boolean flag is confusing in this context; it is not clear what `hasPermissions ... true` does at a glance. If this functionality is desirable, what about adding a separate function `hasAnyPermissions` instead?

Yeah, I was thinking that as well but then thought "this is just another function to add to the context". On second thought adding these variations is actually the more sensible solution.

ashishjh-bst:
> > I think that a boolean flag is confusing in this context; it is not clear what `hasPermissions ... true` does at a glance. If this functionality is desirable, what about adding a separate function `hasAnyPermissions` instead?
> 
> Yeah, I was thinking that as well but then thought "this is just another function to add to the context". On second thought adding these variations is actually the more sensible solution.

I don't really understand the utility for this, what scenario would need it to return true if any of the perms are present?

l-zeuch:
> I don't really understand the utility for this, what scenario would need it to return true if any of the perms are present?

For similar use-cases as the inbuilt commands? Sometimes you want to check for a few permissions available but just one is sufficient. We just solve that with internal code inside the runFuncs, this is no different than that.

ashishjh-bst:
> > I don't really understand the utility for this, what scenario would need it to return true if any of the perms are present?
> 
> For similar use-cases as the inbuilt commands? Sometimes you want to check for a few permissions available but just one is sufficient. We just solve that with internal code inside the runFuncs, this is no different than that.

can you share the example of an internal command doing that?

l-zeuch:
https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L32

https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L64

https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L159

To name a few. See also via `grep -E "RequireDiscordPerms:\s*\[\]int64\{([^}]*discordgo\.Permission\w+){2,}[^}]*\}" -R . -C 2 -n`.

This is in combination with https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/commands/yagcommmand.go#L119 and

https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/commands/yagcommmand.go#L602-L607

which clearly states that *any* of the listed permissions are required. Stricter enforcement has always been handled by the `runFunc`.

ashishjh-bst:
> https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L32
> 
> https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L64
> 
> https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/automod/commands.go#L159
> 
> To name a few. See also via `grep -E "RequireDiscordPerms:\s*\[\]int64\{([^}]*discordgo\.Permission\w+){2,}[^}]*\}" -R . -C 2 -n`.
> 
> This is in combination with https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/commands/yagcommmand.go#L119 and
> 
> https://github.com/botlabs-gg/yagpdb/blob/97da3f6dffa841412e5d1604a1c08ee3a95155af/commands/yagcommmand.go#L602-L607
> 
> which clearly states that *any* of the listed permissions are required. Stricter enforcement has always been handled by the `runFunc`.

Thanks, make sense for non-inclusive perms like manage server and moderate members, in most cases the Admin Perms check above is redundant.

PR #2065, opened 2026-06-01
Add sin, cos and tan trigonometric functions

lemm-e:
This PR adds `sin`, `cos` and `tan` trigonometric functions from the math package.


common/templates/general.go
```diff
@@ -909,6 +909,33 @@ func tmplPow(argX, argY interface{}) float64 {
 	return math.Pow(xySlice[0], xySlice[1])
 }
 
+func tmplSin(arg interface{}) float64 {
+	switch arg.(type) {
+	case int, int16, int32, int64, uint8, uint16, uint32, uint64, float32, float64:
+		return math.Sin(ToFloat64(arg))
+	default:
```
l-zeuch, line 916: This currently does not account for complex numbers, but they're also valid in custom commands (do not ask how I know...). The sine of a complex number can be defined as

$$
\sin (a+bi) = \sin a \cdot \cosh b + i \cos a \cdot \sinh b
$$

I'd suggest we
1. add complex numbers as a valid input to these function as well using the definition above, or
2. also add the hyperbolic functions.

Personally I'm leaning towards both; the hyperbolic functions may have some interesting properties as well and given that we're already here, why not.

common/templates/general.go
```diff
@@ -909,6 +909,33 @@ func tmplPow(argX, argY interface{}) float64 {
...
+		return math.NaN()
+	}
+}
+
+func tmplCos(arg interface{}) float64 {
```
l-zeuch, line 921: The complex cosine can be defined as

$$
\cos (a+bi) = \cos a \cdot \cosh b - i \sin a \cdot \sinh b
$$

common/templates/general.go
```diff
@@ -909,6 +909,33 @@ func tmplPow(argX, argY interface{}) float64 {
...
+	switch arg.(type) {
+	case int, int16, int32, int64, uint8, uint16, uint32, uint64, float32, float64:
+		return math.Cos(ToFloat64(arg))
+	default:
+		return math.NaN()
+	}
+}
+
+func tmplTan(arg interface{}) float64 {
```
l-zeuch, line 930: Similarly,

$$
\tan (a + bi) = \frac{\sin (a+bi)}{\cos (a+bi)}
= \frac{\sin a \cdot \cosh b + i \cos a \cdot \sinh b}{\cos a \cdot \cosh b - i \sin a \cdot \sinh b}
$$

l-zeuch (changes requested):

PR #2072, opened 2026-06-13
New parseNaturalTime function for custom commands: accept natural language inputs

mothka:
This introduces a new function, `parseNaturalTime`. It is similar to [parseTime](https://help.yagpdb.xyz/docs/reference/templates/functions/#parsetime) but takes a natural language input.

The events system and timezone companion already use [when](https://github.com/olebedev/when), but this is not exposed to custom commands. `parseNaturalTime` uses `when` with some improvements so it can read GMT and UTC offsets.    
It will fall back to UTC if no timezone is found.

Usage in a custom command:
```
{{$t := parseNaturalTime "tomorrow at 9pm EDT"}}
{{$t := parseNaturalTime "friday 3pm UTC+1"}}
{{$t := parseNaturalTime "next monday 10am GMT-5:30"}}
{{$t.Format "2006-01-02T15:04:05Z"}}
```
It returns a `time.Time` in UTC.

I was able to get this working on my self-hosted YAGPDB.

Disclaimer: this code was written almost entirely by claude.ai.

Thank you for your consideration. ฅ^•ﻌ•^ฅ


naturaltime/tmplextensions.go
l-zeuch, line 1: This may be better suited as an extension to `when` instead of its own subpackage at the root of the repository. Then, add `tmplParseNaturalTime` to common/templates/whatever-file-is-good and expose it there via either the standardFuncs map or a similar mechanism.

naturaltime/timezones.json
l-zeuch, line 1: How will these be regenerated when/if timezones change or new ones are introduced? It does happen every now and then.

naturaltime/tmplextensions.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// Counts against the template's generic API-call budget (100 calls per execution).
+//
+// Usage in a custom command:
+//
+//	{{$t := parseNaturalTime "tomorrow at 9pm EDT"}}
+//	{{$t := parseNaturalTime "friday 3pm UTC+1"}}
+//	{{$t := parseNaturalTime "next monday 10am GMT-5:30"}}
+//	{{$t.Format "2006-01-02T15:04:05Z"}}
+func tmplParseNaturalTime(ctx *templates.Context) interface{} {
```
l-zeuch:
```suggestion
func tmplParseNaturalTime(ctx *templates.Context) any {
```
Newly written code should ideally use `any` over `interface{}` for clarity, they're equivalent in any and all ways anyway.

naturaltime/tmplextensions.go
```diff
@@ -0,0 +1,144 @@
...
+	return func(input string) (time.Time, error) {
+		if ctx.IncreaseCheckGenericAPICall() {
+			return time.Time{}, templates.ErrTooManyAPICalls
+		}
+
+		loc, cleaned := parseTimezone(input)
+
+		base := time.Now().In(loc)
+		r, err := when.EN.Parse(cleaned, base)
```
l-zeuch, line 134: This will consider English as the sole language for natural time parsing, which is also why I am not exactly on-board with this as a custom command function. `when` provides parsing rules for (Brazilian) Portuguese, English, and Russian. But our userbase is certainly more than just those three languages (I myself am German), so the speakers of the remaining languages will still have to write their own time parsers for natural language, which leaves a bad taste for me.

There is, to my knowledge, no CC function that only considers English for valid inputs. I think we should keep custom commands language-agnostic.

naturaltime/timezones.json
mothka: I think in most cases a "timezone change" is actually a  region deciding to use a seasonal timezone permanently. E.g. EDT for the whole year instead of part EDT and part EST. In that case we would not have to change anything.

If an actual timezone change occurs or is added, of course we would have to do that manually here. I don't see any way around that, unless you want to depend on `when` adding it instead.

Personally I feel that situation would be so rare to not be worried about.

naturaltime/tmplextensions.go
mothka: In that case, there does not seem to be any chance of this feature ever being included, and I will close the pull request. Thank you for your review anyway.

PR #2073, opened 2026-06-15
Add nick and globalname to LeaderboardEntry

jone-cmd:
This PR adds - as the name says `Nick` (the server nickname) and `Globalname` (the global displayname) of the user/guild member to `LeaderboardEntry`, which displays it on the Control Panel and in the public API response.

This PR doesn't raise any docs changes.


jone-cmd:
I accidentally selected the wrong base branch.
I created a new PR with the correct base branch: #2074.

PR #2074, opened 2026-06-15
Add nick and globalname to LeaderboardEntry

jone-cmd:
This PR adds - as the name says `Nick` (the server nickname) and `Globalname` (the global displayname) of the user/guild member to `LeaderboardEntry`, which displays it on the Control Panel and in the public API response.

This PR doesn't raise any docs changes.

Old PR with wrong base branch: #2073

### DIFF
--- a/reputation/reputation.go
+++ b/reputation/reputation.go
@@ -446,9 +446,11 @@ func GetConfig(ctx context.Context, guildID int64) (*models.ReputationConfig, er
 
 type LeaderboardEntry struct {
 	*RankEntry
-	Username string `json:"username"`
-	Bot      bool   `json:"bot"`
-	Avatar   string `json:"avatar"`
+	Username   string `json:"username"`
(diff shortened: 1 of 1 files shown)


l-zeuch:
This won't actually show the nickname on the control panel in its current form, as you're never telling the webpage to render it. Take a look at
https://github.com/botlabs-gg/yagpdb/blob/df0bdd8ddaf65a0847b105d7729f2bfc48aff59a/reputation/assets/reputation_leaderboard.html#L35-L48
and
https://github.com/botlabs-gg/yagpdb/blob/df0bdd8ddaf65a0847b105d7729f2bfc48aff59a/reputation/assets/reputation_leaderboard.html#L70-L83

It also may be pertinent to update the output of the `toprep` command and friends to match this change.

jone-cmd:
Closing because of privacy concerns

PR #2076, opened 2026-06-21
Fix: Components V2 message builder not rendering fields

Mortimer-RR:
The components V2  message builder was failing to render the configuration components in the web UI when the user clicked to add each component. The root cause was a type/missed rename of `rebuild` -> `rebuildEditor`. This appears to be isolated to just these two lines of code and so the fix is quite simple.


ashishjh-bst:
This is already fixed in dev branch and is live on yagpdb already.

Refer to commit here https://github.com/botlabs-gg/yagpdb/commit/e285e7b6d793e4cb8048d56f8b3803d9cb597089

PR #2089, opened 2026-08-01
discordgo: add message attachment flags

l-zeuch:
Add the newly introduced message attachment flags[0]. Without this
change, users will only be able to detect a spoilered file when it is
using the legacy `SPOILER_` prefix, which not every spoilered file will
have. In custom command terms, it would look something like this:

```yag
{{ $ma := index .Message.Attachments 0 }}
{{/* a spoilered file has the third bit set: 1 << 3 */}}
{{ $mask := bitwiseLeftShift 1 3 }} 
{{/* The flag is set when: ($ma.Flags & $mask) != 0 */}}
{{ $is_spoiler := ne (bitwiseAnd $a.Flags $mask) 0 }}
```

[0]: https://docs.discord.com/developers/resources/message#attachment-object-attachment-flags

### DIFF
--- a/lib/discordgo/message.go
+++ b/lib/discordgo/message.go
@@ -414,14 +414,25 @@ func (m *MessageEdit) SetEmbeds(embeds []*MessageEmbed) *MessageEdit {
 
 // A MessageAttachment stores data for message attachments.
 type MessageAttachment struct {
-	ID       string `json:"id"`
-	URL      string `json:"url"`
-	ProxyURL string `json:"proxy_url"`
-	Filename string `json:"filename"`
-	Width    int    `json:"width"`
-	Height   int    `json:"height"`
-	Size     int    `json:"size"`
+	ID       string                 `json:"id"`
+	URL      string                 `json:"url"`
+	ProxyURL string                 `json:"proxy_url"`
+	Filename string                 `json:"filename"`
+	Width    int                    `json:"width"`
+	Height   int                    `json:"height"`
+	Size     int                    `json:"size"`
+	Flags    MessageAttachmentFlags `json:"flags"`
 }
+
+type MessageAttachmentFlags int
+
+const (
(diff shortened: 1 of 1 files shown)


l-zeuch:
On second thought I'm not sure if we should/want to keep `ma.IsSpoiler()` around. I fear the camel's nose with this one -- add this, and suddenly users want to have this for every conceivable flag, when there's *already* a perfectly viable way, which are bitwise operations.

What do you think?

ashishjh-bst:
> On second thought I'm not sure if we should/want to keep `ma.IsSpoiler()` around. I fear the camel's nose with this one -- add this, and suddenly users want to have this for every conceivable flag, when there's _already_ a perfectly viable way, which are bitwise operations.
> 
> What do you think?

IMO we can expose named flags, and expose a hasFlag function, which does something hasFlag(attachment.flags, attachmentsFlags.Spoiler), this would make it generic instead of having a dedicated helper for each flag.

l-zeuch:
> IMO we can expose named flags, and expose a hasFlag function, which does something hasFlag(attachment.flags, attachmentsFlags.Spoiler), this would make it generic instead of having a dedicated helper for each flag.

Sounds good, though we should be careful with this one so we don't pollute the context with too much data. I'll edit this PR to not have `IsSpoiler`. We can discuss how to implement your idea -- I'll look around to gain some inspiration for a decent user experience.

l-zeuch:
As discussed, I removed the `IsSpoiler` convenience function.
A check for the spoiler flag would look something like this:

```gotmpl
{{ $ma := index .Message.Attachments 0 }}
{{/* a spoilered file has the third bit set: 1 << 3 */}}
{{ $mask := bitwiseLeftShift 1 3 }} 
{{/* The flag is set when: ($ma.Flags & $mask) != 0 */}}
{{ $is_spoiler := ne (bitwiseAnd $ma.Flags $mask) 0 }}
```

PR #2094, opened 2026-08-05
Added 'Colors' as both a struct and an element of 'Role'

gandalf-potter00:
As per Discord's documentation, the 'Color' element that is provided by the Role struct is now considered deprecated. The new way to access the primary color of a role is through the element 'Colors'. Colors is itself a struct that contains the primary, secondary and tertiary colors of the role. Even if a role does not have all colors, all elements will be present in the struct. If a role doesn't have a particular color, it will simply return 'null'.

Additionally, 'Color' has been properly labeled as deprecated in the comment directly above the element.

### DIFF
--- a/lib/discordgo/structs.go
+++ b/lib/discordgo/structs.go
@@ -694,6 +694,13 @@ type GuildParams struct {
 	Splash                      string             `json:"splash,omitempty"`
 }
 
+// A Colors holds information about a Discord guild role's colors
+type Colors struct {
+  PrimaryColor int32 `json:"primary_color"`
+  SecondaryColor int32 `json:"secondary_color"`
+  TertiaryColor int32 `json:"tertiary_color"`
+}
+
 // A Role stores information about Discord guild member roles.
 type Role struct {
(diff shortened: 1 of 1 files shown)


lib/discordgo/structs.go
```diff
@@ -712,9 +719,12 @@ type Role struct {
 	// Whether this role is hoisted (shows up separately in member list).
 	Hoist bool `json:"hoist"`
 
-	// The hex color of this role.
+	// The hex color of this role (deprecated).
```
l-zeuch:
```suggestion
	// Deprecated: The hex color of this role.
```
In Go, the magic happens when a docstring paragraph starts with `Deprecated:`: https://go.dev/wiki/Deprecated

l-zeuch (commented):
Nice work! Just one nit

lib/discordgo/structs.go
gandalf-potter00: Ah alright. I'll change that then. Thanks

PR #2106, opened 2026-09-09
rolemenus: use edit member endpoint in "single" mode

l-zeuch:
Use the Edit Member endpoint in "Single" mode. This saves us one API
call and also prevents some weird state where the bot may have already
removed the old role, but due to unforeseen circumstances did not (yet)
manage to add the new role.

Inspired by
https://discord.com/channels/166207328570441728/356486960417734666/767608317224550430,
though for different reasons (saving an API call).

### DIFF
--- a/rolecommands/commonrole.go
+++ b/rolecommands/commonrole.go
@@ -280,7 +280,7 @@ func (c *CommonRoleSettings) ToggleRole(ms *dstate.MemberState) (gaveRole bool, 
 }
 
 func (c *CommonRoleSettings) GroupToggleRole(ctx context.Context, ms *dstate.MemberState) (gaveRole bool, err error) {
-	// Default behaviour of groups is no more restrictions than reuiqred and ignore roles
+	// Groups other than "single" impose no additional restrictions.
 	if c.ParentGroupMode != GroupModeSingle {
 		// We already passed all checks
 		gaveRole, err = c.ToggleRole(ms)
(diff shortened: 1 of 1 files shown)


rolecommands/commonrole.go
```diff
@@ -290,26 +290,37 @@ func (c *CommonRoleSettings) GroupToggleRole(ctx context.Context, ms *dstate.Mem
...
-	// If user already has role it's attempting to give itself
+	// User already has this role, so toggle it off.
 	if slices.Contains(ms.Member.Roles, c.RoleId) {
 		err = common.BotSession.GuildMemberRoleRemove(ms.GuildID, ms.User.ID, c.RoleId)
 		return false, err
 	}
 
-	// Check if the user has any other role commands in this group
 	commands := c.AllGroupRoles(ctx)
+	newRoles := slices.Clone(ms.Member.Roles) // copy so we can safely modify
```
l-zeuch, line 300: I'm not too sure whether we need to actually clone here, but I figured that a slice of ints is not *that* expensive.

rolecommands/commonrole.go
```diff
@@ -290,26 +290,37 @@ func (c *CommonRoleSettings) GroupToggleRole(ctx context.Context, ms *dstate.Mem
...
 
-	// Finally give the role
-	err = common.BotSession.GuildMemberRoleAdd(ms.GuildID, ms.User.ID, c.RoleId)
+	newRoles = append(newRoles, c.RoleId)
+	rs := make([]string, len(newRoles))
+	for i, id := range newRoles {
+		rs[i] = discordgo.StrID(id)
+	}
+
+	err = common.BotSession.GuildMemberEdit(ms.GuildID, ms.User.ID, rs)
```
l-zeuch, line 323: We could consider bumping go.mod to Go 1.27, that would allow us to change the signature of `session.GuildMemberEdit` to accept both `[]int64` and `[]string`, thereby hiding this implementation detail and having the method coerce the type, instead of remembering on every callsite to first convert `[]int64` to `[]string`. Let me know if you want me to implement that change.

PR #2108, opened 2026-09-18
replace all prefixed command usage examples with slash

ashishjh-bst:


l-zeuch:
Discord supports mention-style slash command markup in the form of `</name:id>`. When users click that, their input is prefilled with that slash command. Might be something worth looking into? Unsure how much extra work that'd require, though.

If I recall correctly we already hold some sort of index of globally registered slash commands we can tap into?

ashishjh-bst:
> Discord supports mention-style slash command markup in the form of `</name:id>`. When users click that, their input is prefilled with that slash command. Might be something worth looking into? Unsure how much extra work that'd require, though.
> 
> If I recall correctly we already hold some sort of index of globally registered slash commands we can tap into?

I was already working on that, right now the ids were only stored when something changes with the commands, but now every start would fetch and store them in memory, the latest commit has those changes.

l-zeuch:
> I was already working on that

Heh, great minds think alike :joy:

PR #2111, opened 2026-09-21
dictionary: increase max query length (64 -> 128)

l-zeuch:
This is going to be a fun one. Be prepared for some *really long*
compound words.

Potentially the most known "longest word" is
*Rindfleischetikettierungsüberwachungsaufgabenübertragungsgesetz*, the
short title for a German law (yes, you read that right. there's a long
title too). This one clocks in at 63 (!) letters, so fits within the old
limit of 64.

However, another *actually used* word is
*Grundstücksverkehrsgenehmigungszuständigkeitsübertragungsverordnung*[^1]
(yes, it's another real regulation, see the cites source), though that
has been rescinded in 2007 or so. Yet, it's on Wiktionary[^2], which the
API uses to look up words. This one clocks in at 67 (!!) letters,
surpassing the old limit of 64.

Another, although disputed, contestant is
*Donaudampfschifffahrtselektrizitätenhauptbetriebswerkbauunterbeamtengesellschaft*(-skapitän...)
clocking in with *at least* 80 characters(!!!). Ask a German to continue
chaining this one until they are tired. I bet I could even get over the
new 128 letter limit. However, this word is disputed as "officially
recognised". It's only mentioned on German Wikipedia[^3] in a passing
paragraph, without any actual Wiktionary entry.

I'm not going to try and find even longer words that have an entry, but
I think 128 is a good "upper bound".

[^1]: https://www.buzer.de/gesetz/5188/index.htm
[^2]: https://en.wiktionary.org/wiki/Grundst%C3%BCcksverkehrsgenehmigungszust%C3%A4ndigkeits%C3%BCbertragungsverordnung
[^3]: https://de.wikipedia.org/wiki/Donaudampfschiffahrtsgesellschaftskapit%C3%A4n#Wortl%C3%A4ngen-Rekorde


l-zeuch:
Open to discuss the changes introduced in 613cccd -- the API is case-sensitive (just like language is) so now we will end up having to cache entries like `kuh` and `Kuh`, as they may yield different responses.

